PR #12, opened 2017-11-14
Update FifoBuffer.cpp

jplatipus:
I get compilation "error: use of undeclared identifier 'memcpy'"
Adding the memory.h include fixes this.


dturner:
Many thanks for this

Issue #1, opened 2017-10-23
Add code to handle disconnects


philburk:
The stream can become disconnected when, for example, plugging or unplugging headphones.
By default, always stop and close the stream.
Let developer override method that will reopen a new stream or whatever the app needs.

PR #16, opened 2017-11-23
oboe: automatically stop and close a stream when disconnected

philburk:
WARNING: API change
Modifies OboeStreamCallback onError methods.

Adds protection against stopping or closing a stream at the same time
that the error callback is being processed.

Fix issue #1

### DIFF
--- a/include/oboe/OboeStreamBuilder.h
+++ b/include/oboe/OboeStreamBuilder.h
@@ -190,6 +190,15 @@ public:
         return this;
     }
 
+    /**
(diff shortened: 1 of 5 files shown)


dturner:
Here's the work I needed to do to update the hello-oboe sample:

1) For any class that implements `OboeStreamCallback::onError`, change the method name to `OboeStreamCallback::onErrorAfterClose`
2) In this method remove any code which creates a new thread. Now that Oboe calls this method from a separate thread, we can safely use this thread to do any stream restart work. 

Tested on Pixel XL running 8.1.

PR #18, opened 2017-11-27
New namespace, constexprs and scoped enums

dturner:
!!!BREAKING CHANGE!!! 

## Summary 
1) New `oboe::` namespace
2) Macro constants replaced with `constexpr` and scoped enums
3) `Oboe_convert*Text` methods replaced with single `convertToText` method 
4) Oboe depends on the C++14 standard library

### 1) `oboe` namespace used for all classes and functions

**Why?** 
Having a namespace provides better encapsulation, avoids polluting the global namespace and also avoids having to prefix all Oboe classes and methods with the word "Oboe". 

**What code changes do I need to make?**
If you're using any classes which start with "Oboe" you'll need to change them to use `oboe::`. For example: 

    OboeStreamBuilder builder;

becomes 

    oboe::AudioStreamBuilder builder;

or you can use the oboe namespace without qualification: 

    using namespace oboe;
    AudioStreamBuilder builder;

### 2) Constants defined by macros are replaced by `constexpr` and scoped enums 

**Why?** 
C++ Core Guidelines ES.31: Don’t use macros for constants or “functions” http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-macros2. 
Effective Modern C++ (Scott Meyer) Item 15 - Use constexpr wherever possible. 
Scoped enums allows better type checking at compile time and eliminate the need for long #defines in the global namespace. 

**What code changes do I need to make?** 
Anywhere you were using a value starting with OBOE_ you'll need to change it either to the corresponding value starting with oboe::k. For example `OBOE_NANOS_PER_MICROSECOND` becomes `oboe::kNanosPerMicrosecond`. Or you'll need to change it to the corresponding scoped enum:

`OBOE_AUDIO_FORMAT_PCM_FLOAT` becomes `oboe::AudioFormat::Float`

### 3) `Oboe_convert*Text` methods replaced with single `convertToText` method. **

**Why?** 
A single method is easier to remember and to use.

**What code changes do I need to make?** 
If you are using a `Oboe_convert*Text` method then you'll need to change it to use `convertToText`. For example: 

    Oboe_convertResultToText(result)

becomes

    oboe::convertToText(result)

### 4) Oboe depends on the C++14 standard library

***Why?***
Internally Oboe now uses `std::make_unique` to create a `std::unique_ptr`. Using smart pointers allows for more robust memory management.  

**What code changes do I need to make?** 
When compiling Oboe you must include `-std=c++14` in your compiler flags. You should update your `CMakeLists.txt` to include the following line: 

    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14")


GettingStarted.md
```diff
@@ -48,24 +48,24 @@ Include the Oboe header:
 
     #include <oboe/Oboe.h>
 
-Streams are built using an `OboeStreamBuilder`. Create one like this:
+Streams are built using an `StreamBuilder`. Create one like this:
```
philburk: c/an/a/

dturner: an -> a

GettingStarted.md
```diff
@@ -48,24 +48,24 @@ Include the Oboe header:
...
 
-    builder.setDirection(OBOE_DIRECTION_OUTPUT);
-    builder.setPerformanceMode(OBOE_PERFORMANCE_MODE_LOW_LATENCY);
-    builder.setSharingMode(OBOE_SHARING_MODE_EXCLUSIVE);
+    builder.setDirection(oboe::Direction::Output);
+    builder.setPerformanceMode(oboe::PerformanceMode::LowLatency);
+    builder.setSharingMode(oboe::SharingMode::Exclusive);
 
-Define an `OboeStreamCallback` class to receive callbacks whenever the stream requires new data.
+Define an `StreamCallback` class to receive callbacks whenever the stream requires new data.
```
dturner: an -> a

GettingStarted.md
```diff
@@ -76,21 +76,21 @@ Supply this callback class to the builder:
 
 Open the stream:
 
-    OboeStream *stream;
-    oboe_result_t result = builder.openStream(&stream);
+    oboe::Stream *stream;
+    oboe::Result result = builder.openStream(&stream);
 
 Check the result to make sure the stream was opened successfully. Oboe has many convenience methods for converting its types into human-readable strings, they all start with `Oboe_convert`:
```
dturner: all start with `oboe::convert`

GettingStarted.md
philburk: Are the class names Stream and StreamBuilder too generic? Is there a danger of colliding if developers are using multiple namespaces?

GettingStarted.md
philburk: they now start with "oboe::convert"

include/oboe/Stream.h
```diff
@@ -177,16 +175,16 @@ class OboeStream : public OboeStreamBase {
      * @param timeoutNanoseconds Maximum number of nanoseconds to wait for completion.
      * @return The number of frames actually written or a negative error.
      */
-    virtual oboe_result_t write(const void *buffer,
+    virtual int32_t write(const void *buffer,
```
dturner: IMO this method would be better off returning a `std::tuple` of `Result` and `int32_t`. This avoids the return type representing both the frames written (if positive) or an error (if negative).

include/oboe/Stream.h
```diff
@@ -14,31 +14,31 @@
...
+namespace oboe {
 
-#define DEFAULT_TIMEOUT_NANOS    (2000 * OBOE_NANOS_PER_MILLISECOND)
+constexpr int64_t kDefaultTimeoutNanos = (2000 * kNanosPerMillisecond);
 
 /**
  * Base class for Oboe C++ audio stream.
  */
-class OboeStream : public OboeStreamBase {
+class Stream : public StreamBase {
```
dturner: We might want to consider calling this `AudioStream` since `Stream` is quite a generic term. At some point in the future we might also have a `MidiStream`

src/common/LatencyTuner.cpp
```diff
@@ -36,49 +38,50 @@ oboe_result_t OboeLatencyTuner::tune() {
     }
 
     switch (mState) {
-        case STATE_IDLE:
+        case State::Idle:
             if (--mIdleCountDown <= 0) {
-                mState = STATE_ACTIVE;
+                mState = State::Active;
             }
             mPreviousXRuns = mStream.getXRunCount();
```
dturner: `getXRunCount` might also be better returning a `std::tuple`

include/oboe/StreamBuilder.h
```diff
@@ -171,7 +174,7 @@ class OboeStreamBuilder : public OboeStreamBase {
      * @param performanceMode for example, OBOE_PERFORMANCE_MODE_LOW_LATENCY
```
philburk: old

include/oboe/StreamBuilder.h
```diff
@@ -185,7 +188,7 @@ class OboeStreamBuilder : public OboeStreamBase {
      * @param deviceId device identifier or OBOE_DEVICE_UNSPECIFIED
```
philburk: old

src/common/Utilities.cpp
```diff
@@ -0,0 +1,122 @@
...
+        case AudioFormat::Float:
+            size = sizeof(float);
+            break;
+        default:
+            break;
+    }
+    return size;
+}
+
+const char *convertResultToText(Result returnCode) {
```
philburk: If running with a future version AAudio, there may be new results or new formats. So we should add a default "Unrecognized" return value.

include/oboe/Definitions.h
```diff
@@ -0,0 +1,114 @@
...
+#define OBOE_DEFINITIONS_H
+
+#include <cstdint>
+#include <aaudio/AAudio.h>
+
+namespace oboe {
+
+    constexpr int32_t kUnspecified = 0;
+
+    constexpr int64_t kNanosPerMicrosecond =    1000;
```
mnaganov: Please add a TODO to consider using std::chrono, as it provides a lot of time unit conversion utilities.

include/oboe/Definitions.h
```diff
@@ -0,0 +1,114 @@
...
+        I16 = AAUDIO_FORMAT_PCM_I16,
+        Float = AAUDIO_FORMAT_PCM_FLOAT,
+    };
+
+    enum class DataCallbackResult : aaudio_data_callback_result_t {
+        Continue = AAUDIO_CALLBACK_RESULT_CONTINUE,
+        Stop = AAUDIO_CALLBACK_RESULT_STOP,
+    };
+
+    enum class Result : aaudio_result_t {
```
mnaganov: Consider renaming this class to "Error", which would allow dropping the "Error" prefix from the enum items.

include/oboe/Definitions.h
```diff
@@ -0,0 +1,114 @@
...
+        OK,
+        ErrorBase = AAUDIO_ERROR_BASE,
```
mnaganov: ErrorBase shouldn't be used as a source of an assignment anywhere. To prevent this, I would suggest introducing a helper function 'isError' which compares the value against AAUDIO_ERROR_BASE.

src/common/Stream.cpp
```diff
@@ -0,0 +1,125 @@
...
+    int scheduler = sched_getscheduler(0) & ~SCHED_RESET_ON_FORK; // for current thread
+    if (scheduler != mPreviousScheduler) {
+        LOGD("Stream::fireCallback() scheduler = %s",
+             ((scheduler == SCHED_FIFO) ? "SCHED_FIFO" :
+             ((scheduler == SCHED_OTHER) ? "SCHED_OTHER" :
+             ((scheduler == SCHED_RR) ? "SCHED_RR" : "UNKNOWN")))
+        );
+        mPreviousScheduler = scheduler;
+    }
+    if (mStreamCallback == NULL) {
```
mnaganov: nullptr

include/oboe/StreamBuilder.h
```diff
@@ -199,7 +202,7 @@ class OboeStreamBuilder : public OboeStreamBase {
      * @param streamCallback
      * @return
      */
-    OboeStreamBuilder *setCallback(OboeStreamCallback *streamCallback) {
+    StreamBuilder *setCallback(StreamCallback *streamCallback) {
```
mnaganov: Please consider specifying StreamCallback ownership via use of std::unique_ptr or std::shared_ptr. With a raw pointer, it's easy to run into a trouble of the client code deletes the callback instance.

src/common/Utilities.cpp
```diff
@@ -0,0 +1,122 @@
...
+        if (sample < 0) sample = 0;
+        else if (sample > 0x0FFFF) sample = 0x0FFFF;
+        sample -= 32768; // center at zero
+        destination[i] = (int16_t) sample;
+    }
+}
+
+void convertPcm16ToFloat(const int16_t *source, float *destination, int32_t numSamples) {
+    for (int i = 0; i < numSamples; i++) {
+        destination[i] = source[i] * (1.0f / 32768.0f);
```
mnaganov: Curious: why not just "source[i] / 32768.0f"?

GettingStarted.md
dturner: Done

include/oboe/Definitions.h
dturner: Is it normal to have an `Error::OK` value? We're going to see a lot of: 

`if (error != Error::OK){ `

GettingStarted.md
dturner: Deferring doing the an/a changes until we have made a decision on `Stream` vs `AudioStream` naming

dturner: There would only be a danger if they're using non-qualified namespaces e.g. 

`using namespace oboe; 
Stream stream;`

But yes, I do think `Stream` is too generic. `AudioStream`, `AudioStreamBuilder` etc would be more readable. Also if we ever support MIDI we could have a `MidiStream`. If you agree I'll rename.

GettingStarted.md
dturner: Fixed

include/oboe/StreamBuilder.h
dturner: Fixed

include/oboe/StreamBuilder.h
dturner: Fixed

src/common/Utilities.cpp
dturner: Done for all `convert` methods

include/oboe/Definitions.h
dturner: Why don't I just remove `ErrorBase` from the list of possible values in the `Result` enum?

include/oboe/StreamBase.h
```diff
@@ -81,28 +83,30 @@ class OboeStreamBase {
...
-    int32_t                 mChannelCount = OBOE_UNSPECIFIED;
-    int32_t                 mSampleRate = OBOE_UNSPECIFIED;
-    int32_t                 mDeviceId = OBOE_UNSPECIFIED;
-    int32_t                 mBufferCapacityInFrames = OBOE_UNSPECIFIED;
-
-    oboe_sharing_mode_t     mSharingMode = OBOE_SHARING_MODE_SHARED;
-    oboe_audio_format_t     mFormat = OBOE_AUDIO_FORMAT_UNSPECIFIED;
-    oboe_direction_t        mDirection = OBOE_DIRECTION_OUTPUT;
-    oboe_performance_mode_t mPerformanceMode = OBOE_PERFORMANCE_MODE_NONE;
+    StreamCallback         *mStreamCallback = NULL;
```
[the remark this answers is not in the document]
dturner: Fixed

include/oboe/Definitions.h
mnaganov: Yes, that's perhaps not the best expression. But seeing a lot of "Error" prefixes doesn't make me happy either. Although, I've just checked Vulkan's 'Result' type, and they also have "error" prefixes everywhere. OK, let's not consider changing anything here for now.

include/oboe/Definitions.h
mnaganov: Sure. What I've meant is that if anyone will ever need to check against AAUDIO_ERROR_BASE, you can provide a convenience method.

include/oboe/StreamBuilder.h
dturner: Good spot. I think it makes sense for the client to retain ownership of the callback so that means using a `std::shared_ptr<StreamCallback>`. I've updated the code to do this. Would you mind taking a look.

include/oboe/Definitions.h
dturner: Done

include/oboe/Definitions.h
dturner: `ErrorBase` removed

src/common/Utilities.cpp
dturner: @philburk one for you

include/oboe/LatencyTuner.h
```diff
@@ -83,7 +83,7 @@ class LatencyTuner {
     // arbitrary number of calls to wait before bumping up the latency
     static constexpr int32_t kIdleCount = 8;
 
-    Stream               &mStream;
+    AudioStream               &mStream;
```
philburk: realign

src/opensles/StreamBuffered.cpp
```diff
@@ -44,17 +43,12 @@ Result StreamBuffered::open() {
         LOGD("StreamBuffered(): new FifoBuffer");
         mFifoBuffer = new FifoBuffer(getBytesPerFrame(), 1024); // TODO size?
         // Create a callback that reads from the FIFO
-        mInternalCallback = new AudioStreamBufferedCallback(this);
-        mStreamCallback = mInternalCallback;
-        LOGD("StreamBuffered(): mInternalCallback = %p", mInternalCallback);
+        mStreamCallback = std::make_shared<AudioStreamBufferedCallback>(this);
```
philburk: This makes me nervous.  Can this object be deleted properly if it holds a reference to itself?

philburk: Is "this" actually an AudioStreamBufferedCallback?

mnaganov: "this" is the argument for AudioStreamBufferedCallback's constructor. Thus, the constructed callback will hold a raw pointer to the stream, while the stream will be holding a refcounted pointer to the callback.

include/oboe/LatencyTuner.h
dturner: Done

src/opensles/StreamBuffered.cpp
dturner: Yes, it's the equivalent of doing `new AudioStreamBufferedCallback(this)` except that instead of a raw pointer the result is a `shared_ptr` which will ensure deletion when the refcount drops to zero.

src/common/Utilities.cpp
```diff
@@ -0,0 +1,133 @@
...
+    switch (mode) {
+        OBOE_CASE_ENUM(PerformanceMode::LowLatency);
+        OBOE_CASE_ENUM(PerformanceMode::None);
+        OBOE_CASE_ENUM(PerformanceMode::PowerSaving);
+        default:
+            return "Unrecognized performance mode";
+    }
+}
+
+const char *convertSharingModeToText(SharingMode mode) {
```
krocard: Because the input type is already in the signature, you may or may not want to remove it from the function name:

``` C++
const char *convertToText(AudioFormat format);
const char *convertToText(PerformanceMode mode);
const char *convertToText(SharingMode mode);
```

but for one it is a style preference secondly some people prefer to be explicit.
Of course there is also the template route that allows explicit and implicit input type specification:

``` C++
template <class From>
const char * convertToText(From);

template <>
const char* convertToText<AudioFormat>(AudioFormat format);
// ...

// Which can be use explicitly:
auto format = AudioFormat::I16;
cout << convertToText<AudioFormat>(format);
// or implicitly
cout << convertToText(format);
```
But this is not the current style of the API, plus it is a backward incompatible change that might not be worth it.

src/opensles/StreamBuffered.cpp
philburk: Cool. Thanks for the explanation.

FullGuide.md
```diff
@@ -28,11 +28,11 @@ The audio device attached to a stream determines whether the stream is for input
 
 A stream has a sharing mode:
 
-*   `OBOE_SHARING_MODE_EXCLUSIVE` (available on API 26+) means the stream has exclusive access to its audio device; the device cannot be used by any other audio stream. If the audio device is already in use, it might not be possible for the stream to have exclusive access. Exclusive streams provide the lowest possible latency, but they are also more likely to get disconnected. You should close exclusive streams as soon as you no longer need them, so that other apps can access the device.
-*   `OBOE_SHARING_MODE_SHARED` allows Oboe to mix audio. Oboe mixes all the shared streams assigned to the same device.
+*   `SharingMode::Exclusive` (available on API 26+) means the stream has exclusive access to its audio device; the device cannot be used by any other audio stream. If the audio device is already in use, it might not be possible for the stream to have exclusive access. Exclusive streams provide the lowest possible latency, but they are also more likely to get disconnected. You should close exclusive streams as soon as you no longer need them, so that other apps can access the device.
```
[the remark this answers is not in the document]
philburk: I am very open to a better term than "audio device". Calling it a "front end of this hardware" is still not very clear.  Maybe "audio resource"?  A phone can have two "front ends" for a device. One is used by AudioTrack and one is available for MMAP.

include/oboe/AudioStreamBase.h
```diff
@@ -14,11 +14,14 @@
...
+#ifndef OBOE_STREAM_BASE_H_
+#define OBOE_STREAM_BASE_H_
 
-#include "oboe/OboeStreamCallback.h"
-#include "oboe/OboeDefinitions.h"
+#include <memory>
+#include "oboe/AudioStreamCallback.h"
+#include "oboe/Definitions.h"
+
+namespace oboe {
```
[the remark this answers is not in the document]
philburk: The intention was to have an App statically link to Oboe.
I suppose there could be a problem if an app tried to use with a library that had it's own calls to an old version of Oboe.  But that seems unlikely. Most audio libraries just pass PCM in and out and leave the audio I/O to the main app..

src/common/Utilities.cpp
philburk: This CL breaks the API  anyway. We warned people it would break.  And we are modernizing. So now is the time if we want to do this kind of stuff.  Up to you Don.

include/oboe/AudioStreamBase.h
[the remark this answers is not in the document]
krocard: If Oboe is static linked, inline version namespace have a lot less attractivity. Do not bother with it.

FullGuide.md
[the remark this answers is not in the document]
philburk: Let's fix this doc issue in a later CL. We need to get this big refactoring done.

include/oboe/Definitions.h
```diff
@@ -0,0 +1,115 @@
...
+        I16 = AAUDIO_FORMAT_PCM_I16,
+        Float = AAUDIO_FORMAT_PCM_FLOAT,
+    };
+
+    enum class DataCallbackResult : aaudio_data_callback_result_t {
+        Continue = AAUDIO_CALLBACK_RESULT_CONTINUE,
+        Stop = AAUDIO_CALLBACK_RESULT_STOP,
+    };
+
+    enum class Result : aaudio_result_t {
```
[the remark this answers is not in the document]
dturner: Thanks. Added checks for all AAudio primitive data types.

src/common/Utilities.cpp
dturner: The implicit use of convertToText is really nice. Have implemented.

src/common/Utilities.cpp
```diff
@@ -0,0 +1,133 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+#include <unistd.h>
+#include "oboe/Definitions.h"
+#include "oboe/Utilities.h"
+
+#define OBOE_CASE_ENUM(name) case name: return #name
```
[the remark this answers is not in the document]
dturner: I tried this and it just outputs the type name e.g. "Result::OK" twice in a row. I think I need to static_cast `name` to an `int32_t` then concat with the return string - could be tough to implement in a macro.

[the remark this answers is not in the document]
mnaganov: Right, for Kevin's approach to work, "name" must be a preprocessor define, so it can be expanded into a value during macro expansion phase, like described here: https://gcc.gnu.org/onlinedocs/gcc-4.1.2/cpp/Stringification.html. Preprocessor knows nothing about the values of enums.

But actually, the tricky part that involves the use of preprocessor is getting the program symbol name, that is, the name of the enum constant. The value of a enum constant can be obtained at compile- or run-time.

So, for example, in your enum to string function you could use a stringstream, where you first put the value of the enum (unconditionally), and then use the stringification trick in order to add the symbol name of the enum.

src/common/Utilities.cpp
```diff
@@ -0,0 +1,133 @@
...
+        OBOE_CASE_ENUM(Result::ErrorNoMemory);
+        OBOE_CASE_ENUM(Result::ErrorNull);
+        OBOE_CASE_ENUM(Result::ErrorTimeout);
+        OBOE_CASE_ENUM(Result::ErrorWouldBlock);
+        OBOE_CASE_ENUM(Result::ErrorInvalidFormat);
+        OBOE_CASE_ENUM(Result::ErrorOutOfRange);
+        OBOE_CASE_ENUM(Result::ErrorNoService);
+        OBOE_CASE_ENUM(Result::ErrorInvalidRate);
+        default:
+            return "Unrecognized result";
```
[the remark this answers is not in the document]
dturner: I've added a TODO for this as it's a non-breaking change and I couldn't figure out exactly how this could be achieved easily without a lot of messing about with stringstream->string->c_str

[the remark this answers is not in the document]
krocard: This IS a breaking change. You have to return a string buffer that the client owned (eg: std::string). instead of returning a static char *

src/common/Utilities.cpp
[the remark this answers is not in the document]
krocard: Right, I though the name was an AAudio constant. There are solutions to convert the enum value to a compile time string, but it would add too much complexity. Hopefully in C++20, reflection will make all that much easier.
The alternative would be to construct the string at build time which would make it much easier:

```C++
template <class Enum, Enum enum>
const char* enumToStr(const char* name) {
    const static str = name
}
#define OBOE_CASE_ENUM(name) case name: return std::string(#name) + " (" + std::to_string(name) + ')';
```

But that changes the API, as a std::string has to be returned.

FullGuide.md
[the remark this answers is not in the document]
dturner: I used the concept of an audio device "endpoint" to explain this more clearly in #23.

src/common/Utilities.cpp
philburk: To avoid rounding in opposite directions because of the cast. But Andy has a better technique now. Will change.

PR #23, opened 2017-12-08
Clearer explanation for exclusive sharing mode

dturner:
I have added the concept of an audio device **endpoint**. Feedback welcome.


FullGuide.md
```diff
@@ -26,13 +26,14 @@ The audio device attached to a stream determines whether the stream is for input
...
 
-*   `SharingMode::Exclusive` (available on API 26+) means the stream has exclusive access to its audio device; the device cannot be used by any other audio stream. If the audio device is already in use, it might not be possible for the stream to have exclusive access. Exclusive streams provide the lowest possible latency, but they are also more likely to get disconnected. You should close exclusive streams as soon as you no longer need them, so that other apps can access the device.
+*   `SharingMode::Exclusive` (available on API 26+) means the stream has exclusive access to an endpoint on its audio device; the endpoint cannot be used by any other audio stream. If the exclusive endpoint is already in use, it might not be possible for the stream to obtain access to it. Exclusive streams provide the lowest possible latency by bypassing the mixer stage, but they are also more likely to get disconnected. You should close exclusive streams as soon as you no longer need them, so that other apps can access the device. Not all audio devices provide exclusive endpoints.
 *   `SharingMode::Shared` allows Oboe to mix audio. Oboe mixes all the shared streams assigned to the same device.
```
philburk: device or "endpoint"

FullGuide.md
```diff
@@ -26,13 +26,14 @@ The audio device attached to a stream determines whether the stream is for input
...
 
-You can explicitly request the sharing mode when you create a stream, although you are not guaranteed to receive that mode. By default,
-the sharing mode is `Shared`.
+You can explicitly request the sharing mode when you create a stream, although you are not guaranteed to receive that mode. By default, the sharing mode is `Shared`.
 
 ### Audio format
```
philburk: The diagram below is not exactly correct.  There is generally only ONE MMAP endpoint available per device. It can be used by one EXCLUSIVE stream OR multiple SHARED streams. But not both.
So if any app opens a SHARED MMAP stream then it can prevent another app from opening an EXCLUSIVE MMAP stream.

PR #29, opened 2017-12-18
Adding version information

dturner:
Here's how you can get Oboe version information: 

Major, minor and subminor version numbers: 

    oboe::Version::Major
    oboe::Version::Minor
    oboe::Version::Patch

Text representation (returns string literal "MAJOR.MINOR.SUBMINOR" e.g "0.9.0")

    oboe::Version::Text

Integer representation (returns single integer representing all 3 version numbers)

    oboe::Version::Number


src/common/Version.cpp
```diff
@@ -0,0 +1,39 @@
...
+ * limitations under the License.
+ */
+
+#include <sstream>
+#include "oboe/Version.h"
+
+namespace oboe {
+
+// Max digits in 32-bit unsigned int = 10, plus two periods, plus null terminator = 13
+constexpr int kMaxVersionStringLength = 13;
```
philburk: longest string would be "255.255.65535" + nul => 14

include/oboe/Version.h
```diff
@@ -0,0 +1,55 @@
...
+    // This is incremented when we make breaking API changes. Based loosely on https://semver.org/
+    static constexpr uint8_t MajorNumber = 0;
+
+    // This is incremented when we add backwards compatible functionality. Or set to zero when kVersionMajor is
+    // incremented
+    static constexpr uint8_t MinorNumber = 9;
+
+    // This is incremented when we make backwards compatible bug fixes. Or set to zero when kVersionMinor is
+    // incremented
+    static constexpr uint16_t SubMinorNumber = 0;
```
philburk: The word "Revision" is also common.

mnaganov: Or "patch", as in https://semver.org/

mnaganov: Or "patch", as in https://semver.org/

include/oboe/Version.h
```diff
@@ -0,0 +1,55 @@
...
+#ifndef OBOE_VERSIONINFO_H
+#define OBOE_VERSIONINFO_H
+
+namespace oboe {
+
+class Version {
+
+public:
+    // This is incremented when we make breaking API changes. Based loosely on https://semver.org/
+    static constexpr uint8_t MajorNumber = 0;
```
mnaganov: I would argue that specifying the package version number is probably one (and maybe only) good application for the preprocessor. The reason is that with preprocessor, you can easily produce the version string at the build time (using the stringification feature). Whereas your current implementation requires the C library for that.

Potentially, this can be done using a constexpr function, but then you will need to write int to string conversion logic yourself, and this seems like an overkill (see https://stackoverflow.com/a/26824971/4477684).

src/common/Version.cpp
```diff
@@ -0,0 +1,39 @@
...
+    static char text[kMaxVersionStringLength];
+    snprintf(text, kMaxVersionStringLength, "%d.%d.%d",
+             MajorNumber,
+             MinorNumber,
+             SubMinorNumber);
+    return text;
+}
+
+uint32_t Version::toInt(){
+    return MajorNumber << 24 | MinorNumber << 16 | SubMinorNumber;
```
mnaganov: Technically, those shifts depend on the data size of the version components.

include/oboe/Version.h
dturner: Yes I saw that whilst I was researching about whether we should switch to using `std::string`, it seems that you have to write some pretty advanced (read: difficult to read/maintain) C++ to produce compile-time string literals, something which is easy in C (with the down side of pointer/array duality).

src/common/Version.cpp
dturner: Good spot :) I was working under the false assumption that the number of characters for 8-bit integer + 8-bit integer + 16-bit integer was the same as that for a 32-bit integer.

include/oboe/Version.h
dturner: Changed to Patch to comply more closely with semver

src/common/Version.cpp
dturner: Removed this - it's no longer necessary when using #defines

include/oboe/Version.h
dturner: The static constants are now initialized using preprocessor directives. No need for any functions so I've removed the Version.cpp file.

src/common/Version.cpp
dturner: I've added comments specifying the data size for the #defines.

include/oboe/Version.h
```diff
@@ -0,0 +1,87 @@
...
+        OBOE_TOSTRING(OBOE_VERSION_PATCH)
+
+// Type: 32-bit unsigned int. See below for description.
+#define OBOE_VERSION_NUMBER ((OBOE_VERSION_MAJOR << 24) | (OBOE_VERSION_MINOR << 16) | OBOE_VERSION_PATCH)
+
+namespace oboe {
+
+class Version {
+
+public:
```
mnaganov: Can be made a struct now.

dturner: Done

PR #46, opened 2018-01-11
Rename ApiIndex to AudioApi, move into Definitions.h

dturner:


dturner:
A minor breaking change since the API method name has changed and the ApiIndex enum has been moved into Definitions.h

src/common/Utilities.cpp
```diff
@@ -161,5 +161,17 @@ const char *convertToText<StreamState>(StreamState state) {
     }
 }
 
+template<>
+const char *convertToText<AudioApi >(AudioApi audioApi) {
+
+    switch (audioApi) {
+        OBOE_CASE_ENUM(AudioApi::Unspecified);
```
philburk: Now that we are using smaller compound names. Would it better to return "Unspecified" instead of "AudioApi::Unspecified"? Maybe it does.

dturner: Yes `AudioApi::Unspecified == kUnspecified == 0`. Is that what you meant?

philburk: What is the char* that it returns? "AudioApi::Unspecified" or "Unspecified"?

dturner: I see what you mean now. It returns "AudioApi::Unspecfied". We could have a 2nd parameter to OBOE_CASE_ENUM which is the char* to return so the call would look like this: 

OBOE_CASE_ENUM(AudioApi::Unspecified, Unspecified)

Smaller name at the cost of repetition in the macro call. Your choice.

philburk: I think that will be more useful for developers. For example, when logging:
ALOG("using %s API", convertToText(audioApi)
it would be nice to print:
     "calling  AAudio API"
instead of
     "calling AudioApi::AAudio API"

Would this work:
    using AudioApi::Unspecified;
    OBOE_CASE_ENUM(Unspecified);

dturner: Unfortunately not. https://stackoverflow.com/questions/24212921/is-it-possible-to-alias-an-enum-class-enumerator. 

If we're going to make this change then I'd probably forego the macro and just do: 

```
case AudioApi::Unspecified: return "Unspecified";
```
OK, there's some redundancy but it's easy to understand what's going on rather than the "magic" inside the macro.

philburk: I agree. Good plan.

dturner: Done

philburk:
Looks good!

Issue #40, opened 2018-01-07
Crashes on many Android 8.0 devices


svenoaks:
I recently tried Oboe in my app for sound output for Android 8.0+ (only). I had to halt rollout because probably the majority of devices with Android 8.0 (none with 8.1) have repeating crashes probably making app unusable. I can't reproduce on API 26 or 27 emulator or a Nexus 5X which I upgraded and tested on both 8.0 and 8.1. When I removed Oboe for 8.0 in new release crashes went away.

The most common crashes not directly implicate Oboe, but lesser crashes on same devices do. Perhaps relevant information is I am building Oboe with old Android.mk system and with APP_PLATFORM = android-14 (though I think it is using 21 for Arm64 anyway automatically). Clang and c++_shared. NDK latest version in SDK manager. 16.1.4479499

Thinking about buying one of these devices and trying to get more information.

![1](https://user-images.githubusercontent.com/5206058/34654497-8c88efd0-f3ca-11e7-8d89-a5a8a3a1327f.png)
![2](https://user-images.githubusercontent.com/5206058/34654498-8c990b0e-f3ca-11e7-9ffe-caff7121560a.png)
![3](https://user-images.githubusercontent.com/5206058/34654499-8caa3c30-f3ca-11e7-9a4d-1b5cdf4abc69.png)
![4](https://user-images.githubusercontent.com/5206058/34654500-8ccd00b2-f3ca-11e7-8329-33372a3c3e7a.png)
![5](https://user-images.githubusercontent.com/5206058/34654501-8cd9a268-f3ca-11e7-9a70-28efb9d0e42c.png)
![6](https://user-images.githubusercontent.com/5206058/34654502-8cfd9632-f3ca-11e7-936a-f8a812738863.png)
![7](https://user-images.githubusercontent.com/5206058/34654503-8d0b317a-f3ca-11e7-9d38-27407b5c203e.png)
![8](https://user-images.githubusercontent.com/5206058/34654504-8d184266-f3ca-11e7-8c48-7a09da863e6a.png)

philburk:
Thanks for this report. We need to resolve this.

Some background:
On 8.0 devices Oboe will use AAudio by default because it is available.  AAudio on those 8.0 will use the older legacy audio data path. It only uses MMAP on Pixel devices on 8.1+.

Do you have a bugreport file that we can analyze?

If not, can you provide an APK that crashes so we can collect a bugreport.

[...]

PR #51, opened 2018-02-06
Do not use AAudio unless running on Android 8.1 or higher

mnaganov:
From https://github.com/google/oboe/issues/40, it seems that AAudio
on Android 8.0 isn't stable enough. As a safety measure, always
fall back to OpenSL ES on it.

### DIFF
--- a/src/aaudio/AudioStreamAAudio.cpp
+++ b/src/aaudio/AudioStreamAAudio.cpp
@@ -16,12 +16,16 @@
 
 #include <assert.h>
 #include <stdint.h>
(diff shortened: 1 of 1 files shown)


mnaganov:
Please note that I actually work for Google, just using my personal GitHub account. I've signed the CLA just in case, but I believe it's not really needed in this case.

dturner:
Tested on Pixel (emulator) running API 26. Result: OpenSL ES is used - PASS
Tested on Pixel 2 running API 27. Result: AAudio is used - PASS

+2 Approved

PR #55, opened 2018-02-09
oboe: add blocking read/write

philburk:
Use FIFO for OpenSL ES.
This will allow blocking read/write calls on a stream without using a callback.

### DIFF
--- a/src/aaudio/AudioStreamAAudio.cpp
+++ b/src/aaudio/AudioStreamAAudio.cpp
@@ -70,7 +70,6 @@ static void oboe_aaudio_error_callback_proc(
     }
 }
 
-
 namespace oboe {
 
 /*
@@ -301,6 +300,18 @@ int32_t AudioStreamAAudio::write(const void *buffer,
     }
 }
 
+int32_t AudioStreamAAudio::read(void *buffer,
+                                 int32_t numFrames,
+                                 int64_t timeoutNanoseconds)
+{
+    AAudioStream *stream = mAAudioStream.load();
+    if (stream != nullptr) {
+        return mLibLoader->stream_read(mAAudioStream, buffer, numFrames, timeoutNanoseconds);
+    } else {
+        return static_cast<int32_t>(Result::ErrorNull);
+    }
+}
+
 Result AudioStreamAAudio::waitForStateChange(StreamState currentState,
                                         StreamState *nextState,
                                         int64_t timeoutNanoseconds)
@@ -358,8 +369,7 @@ int32_t AudioStreamAAudio::getFramesPerBurst()
     }
 }
 
-int64_t AudioStreamAAudio::getFramesRead()
-{
+int64_t AudioStreamAAudio::getFramesRead() const {
     AAudioStream *stream = mAAudioStream.load();
     if (stream != nullptr) {
         return mLibLoader->stream_getFramesRead(stream);
@@ -367,8 +377,8 @@ int64_t AudioStreamAAudio::getFramesRead()
         return static_cast<int32_t>(Result::ErrorNull);
     }
 }
-int64_t AudioStreamAAudio::getFramesWritten()
-{
+
+int64_t AudioStreamAAudio::getFramesWritten() const {
     AAudioStream *stream = mAAudioStream.load();
     if (stream != nullptr) {
         return mLibLoader->stream_getFramesWritten(stream);
--- a/src/aaudio/AudioStreamAAudio.cpp
+++ b/src/aaudio/AudioStreamAAudio.cpp
@@ -27,10 +27,8 @@
 #include <sys/system_properties.h>
 #endif
 
-
 using namespace oboe;
 AAudioLoader *AudioStreamAAudio::mLibLoader = nullptr;
-
 
 // 'C' wrapper for the data callback method
 static aaudio_data_callback_result_t oboe_aaudio_data_callback_proc(
@@ -79,8 +77,7 @@ AudioStreamAAudio::AudioStreamAAudio(const AudioStreamBuilder &builder)
     : AudioStream(builder)
     , mFloatCallbackBuffer(nullptr)
     , mShortCallbackBuffer(nullptr)
-    , mAAudioStream(nullptr)
-{
+    , mAAudioStream(nullptr) {
     mCallbackThreadEnabled.store(false);
     LOGD("AudioStreamAAudio() call isSupported()");
     isSupported();
@@ -244,8 +241,7 @@ Result AudioStreamAAudio::convertApplicationDataToNative(int32_t numFrames) {
     return result;
 }
 
-Result AudioStreamAAudio::requestStart()
-{
+Result AudioStreamAAudio::requestStart() {
     std::lock_guard<std::mutex> lock(mLock);
     AAudioStream *stream = mAAudioStream.load();
     if (stream != nullptr) {
@@ -255,8 +251,7 @@ Result AudioStreamAAudio::requestStart()
     }
 }
 
-Result AudioStreamAAudio::requestPause()
-{
+Result AudioStreamAAudio::requestPause() {
     std::lock_guard<std::mutex> lock(mLock);
     AAudioStream *stream = mAAudioStream.load();
     if (stream != nullptr) {
@@ -276,8 +271,7 @@ Result AudioStreamAAudio::requestFlush() {
     }
 }
 
-Result AudioStreamAAudio::requestStop()
-{
+Result AudioStreamAAudio::requestStop() {
     std::lock_guard<std::mutex> lock(mLock);
     AAudioStream *stream = mAAudioStream.load();
     if (stream != nullptr) {
@@ -290,8 +284,7 @@ Result AudioStreamAAudio::requestStop()
 // TODO: Update to return tuple of Result and framesWritten (avoids cast)
 int32_t AudioStreamAAudio::write(const void *buffer,
                                      int32_t numFrames,
-                                     int64_t timeoutNanoseconds)
-{
+                                     int64_t timeoutNanoseconds) {
     AAudioStream *stream = mAAudioStream.load();
     if (stream != nullptr) {
         return mLibLoader->stream_write(mAAudioStream, buffer, numFrames, timeoutNanoseconds);
@@ -302,8 +295,7 @@ int32_t AudioStreamAAudio::write(const void *buffer,
 
 int32_t AudioStreamAAudio::read(void *buffer,
                                  int32_t numFrames,
-                                 int64_t timeoutNanoseconds)
-{
+                                 int64_t timeoutNanoseconds) {
     AAudioStream *stream = mAAudioStream.load();
     if (stream != nullptr) {
         return mLibLoader->stream_read(mAAudioStream, buffer, numFrames, timeoutNanoseconds);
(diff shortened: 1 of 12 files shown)


src/opensles/AudioStreamBuffered.cpp
```diff
@@ -25,23 +25,18 @@ namespace oboe {
  * AudioStream with a FifoBuffer
  */
 AudioStreamBuffered::AudioStreamBuffered(const AudioStreamBuilder &builder)
-        : AudioStream(builder)
-        , mFifoBuffer(nullptr)
```
mnaganov: Why not to change the field type to a smart pointer and forget about manual initialization / deletion:

std::unique_ptr<FifoBuffer> mFifoBuffer;

src/opensles/AudioStreamBuffered.cpp
```diff
@@ -25,23 +25,18 @@ namespace oboe {
...
     // If the caller does not provide a callback use our own internal
     // callback that reads data from the FIFO.
     if (getCallback() == nullptr) {
-        LOGD("AudioStreamBuffered(): new FifoBuffer");
-        // TODO: Fix memory leak here
+        LOGD("AudioStreamBuffered(): new FifoBuffer(bytesPerFrame=%d", getBytesPerFrame());
+        // FIFO is configured with the same format and channels as the stream.
         mFifoBuffer = new FifoBuffer(getBytesPerFrame(), 1024); // TODO size?
         // Create a callback that reads from the FIFO
         mInternalCallback = std::unique_ptr<AudioStreamBufferedCallback>(new AudioStreamBufferedCallback(this));
```
mnaganov: You can use std::unique_ptr::reset method http://en.cppreference.com/w/cpp/memory/unique_ptr/reset:

mInternalCallback.reset(new AudioStreamBufferedCallback(this))

philburk: Thanks. I cleaned up this section quite a bit.

src/opensles/AudioStreamBuffered.cpp
philburk: Done

src/aaudio/AudioStreamAAudio.cpp
```diff
@@ -282,6 +281,18 @@ int32_t AudioStreamAAudio::write(const void *buffer,
     }
 }
 
+int32_t AudioStreamAAudio::read(void *buffer,
+                                 int32_t numFrames,
+                                 int64_t timeoutNanoseconds)
+{
```
philburk: move to "{" previous line.

src/opensles/AudioStreamBuffered.cpp
```diff
@@ -25,57 +25,187 @@ namespace oboe {
...
+    }
+    return transfer((void *) buffer, numFrames, timeoutNanoseconds);
+}
+
+// Read data from the FIFO that was written by the callback.
+int32_t AudioStreamBuffered::read(void *buffer,
+                                  int32_t numFrames,
+                                  int64_t timeoutNanoseconds) {
+    if (getDirection() == Direction::Output) {
+        return (int32_t) Result::ErrorUnavailable; // TODO review, better error code?
```
mnaganov: ErrorInvalidState?

philburk: or maybe ErrorUnsupported

src/fifo/FifoBuffer.cpp
```diff
@@ -31,17 +32,19 @@ FifoBuffer::FifoBuffer(uint32_t bytesPerFrame, uint32_t capacityInFrames)
         : mFrameCapacity(capacityInFrames)
         , mBytesPerFrame(bytesPerFrame)
         , mStorage(NULL)
-        , mReadAtNanoseconds(0)
         , mFramesReadCount(0)
         , mFramesUnderrunCount(0)
         , mUnderrunCount(0)
 {
+    assert(bytesPerFrame > 0);
```
dturner: Is there a more elegant way we could handle these two assertions?

src/opensles/AudioStreamBuffered.cpp
```diff
@@ -25,57 +25,187 @@ namespace oboe {
...
-        // TODO: Fix memory leak here
-        mFifoBuffer = new FifoBuffer(getBytesPerFrame(), 1024); // TODO size?
-        // Create a callback that reads from the FIFO
-        mInternalCallback = std::unique_ptr<AudioStreamBufferedCallback>(new AudioStreamBufferedCallback(this));
-        mStreamCallback = mInternalCallback.get();
+    if (usingFIFO()) {
+        // FIFO is configured with the same format and channels as the stream.
+        int32_t capacity = getBufferCapacityInFrames();
+        if (capacity == oboe::kUnspecified) {
+            capacity = getFramesPerBurst() * 16; // arbitrary
```
dturner: Change 16 to konstant

src/opensles/AudioStreamBuffered.cpp
```diff
@@ -25,57 +25,187 @@ namespace oboe {
...
+
+
+    // This is called by the OpenSL ES callback to read or write the back end of the FIFO.
+DataCallbackResult AudioStreamBuffered::onDefaultCallback(void *audioData, int numFrames) {
+    int32_t framesTransferred  = 0;
+
+    if (getDirection() == oboe::Direction::Output) {
+        // Read from the FIFO and write to audioData
+        framesTransferred = mFifoBuffer->readNow(audioData, numFrames);
+        LOGV("%s() read %d / %d frames from FIFO", __func__, framesTransferred, numFrames);
```
dturner: Logging in callback, remove

src/opensles/AudioStreamBuffered.cpp
```diff
@@ -25,57 +25,187 @@ namespace oboe {
...
+    } else {
+        // Read from audioData and write to the FIFO
+        framesTransferred = mFifoBuffer->write(audioData, numFrames); // FIXME writeNow????
+        LOGV("%s() wrote %d / %d frames to FIFO", __func__, framesTransferred, numFrames);
```
dturner: Logging in callback, remove

src/opensles/AudioStreamBuffered.cpp
```diff
@@ -25,57 +25,187 @@ namespace oboe {
...
+int64_t AudioStreamBuffered::predictNextCallbackTime() {
+    if (mBackgroundRanAtNanoseconds == 0) {
+        return 0;
+    }
+    int64_t nanosPerBuffer = (kNanosPerSecond * mLastBackgroundSize) / getSampleRate();
+    const int64_t margin = 200 * kNanosPerMicrosecond; // arbitrary delay so we wake up just after
+    return mBackgroundRanAtNanoseconds + nanosPerBuffer + margin;
+}
+
+// TODO: This method should return a tuple of Result,int32_t where the
```
dturner: There's a few instances where returning a tuple would certainly be easier to understand, rather than having: 

positive int32_t = some interesting value for the caller
negative int32_t = error which must be cast back to a Result type

Any thoughts?

src/opensles/AudioStreamBuffered.cpp
dturner: I would categorize writing to an input stream or reading from an output stream as invalid operations, so ErrorInvalidOperation?

src/opensles/AudioStreamBuffered.cpp
mnaganov: Right, the current way is more appropriate for C. E.g. for sizes, there is a standard type is GNU C library called 'ssize_t' which either contains a positive size or negative error code.

In C++, I would prefer a dedicated type that can be checked as a boolean in a logical expression, and then either the value or an error code can be extracted from it. Here is a good example: https://www.kdab.com/tuple-pair-cpp-apis/  Look for "class value_or_error" in the article. Note that they use std::variant in the implementation, which is from c++17, but we don't actually need it, can just have two fields instead.

src/opensles/AudioStreamOpenSLES.cpp
```diff
@@ -129,10 +133,10 @@ SLresult AudioStreamOpenSLES::enqueueCallbackBuffer(SLAndroidSimpleBufferQueueIt
 
 SLresult AudioStreamOpenSLES::processBufferCallback(SLAndroidSimpleBufferQueueItf bq) {
     // Ask the callback to fill the output buffer with data.
-    Result result = fireCallback(mCallbackBuffer, mFramesPerCallback);
-    if (result != Result::OK) {
+    DataCallbackResult result = fireCallback(mCallbackBuffer, mFramesPerCallback);
+    if (result != DataCallbackResult::Continue) {
         LOGE("Oboe callback returned %d", result);
-        return SL_RESULT_INTERNAL_ERROR;
+        return SL_RESULT_INTERNAL_ERROR; // TODO How should we stop OpenSL ES.
```
dturner: We could use the SL player interface to set the play state to stopped. E.g. 

https://github.com/googlesamples/android-ndk/blob/master/audio-echo/app/src/main/cpp/audio_player.cpp#L235

Additionally, we could set the stream state to stopped.

src/opensles/AudioStreamBuffered.cpp
philburk: Hmmm. Maybe. The way that value_or_error template override * seems odd. You use *result to get the value and result.error() to get the error.

src/opensles/AudioStreamOpenSLES.cpp
philburk: Can that be called safely from inside an OpenSL callback?

src/opensles/AudioStreamBuffered.cpp
philburk: I like InvalidOperation. Currently all oboe errors are defined as AAUDIO_ERROR_*. We could add INVALID_OPERATION to aaudio and try keep the numbers in sync.

src/fifo/FifoBuffer.cpp
philburk: Maybe. I added them because this is a constructor and the FIFO cannot work if the values are incorrect. If they are zero then the FIFO seems to work but doesn't. This took a while to debug. I generally don't like asserts but they seem justified here.

src/opensles/AudioStreamBuffered.cpp
philburk: Done

src/opensles/AudioStreamBuffered.cpp
philburk: Done

src/opensles/AudioStreamBuffered.cpp
philburk: Done

src/opensles/AudioStreamBuffered.cpp
philburk: I implemented this. But I also notice that many oboe functions return a positive value or a negative error. Changing all of them would be disruptive.

src/fifo/FifoBuffer.cpp
dturner: Yes agreed, just checking they weren't left in for debugging purposes.

src/opensles/AudioStreamOpenSLES.cpp
dturner: 90% sure they can. We should probably add a test for this.

philburk: This is a separate issue from the read/write patch. 
I filed:   https://github.com/google/oboe/issues/61

Let's finish up this patch.

PR #62, opened 2018-02-14
Changing callback interface to use references rather than pointers

### LINKED ISSUE #31: AudioStreamCallback should use references rather than pointers

dturner:
Fixes #31. 

How much deeper should this change go? Do we want to modify [`AudioStreamAAudio::callOnAudioReady`](https://github.com/google/oboe/blob/master/src/aaudio/AudioStreamAAudio.h#L95) and other similar methods as well? 

Note: This is a breaking API change, so the hello-oboe app will need to be updated as well. Another example of when it'd be useful to have the sample in this repo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,4 +26,6 @@ We would love to receive your pull requests. Before we can though, please read t
 
 ## Version history
 
+- 14th February 2018 - v0.11 `AudioStreamCallback` methods use references rather than pointers
+- 18th January 2018 - v0.10 Input streams supported via callbacks (+ minor API changes/bug fixes)
 - 18th October 2017 - v0.9 Initial developer preview
--- a/include/oboe/AudioStreamCallback.h
+++ b/include/oboe/AudioStreamCallback.h
@@ -30,13 +30,13 @@ public:
     /**
      * A buffer is ready for processing.
      *
-     * @param oboeStream pointer to the associated stream
+     * @param oboeStream the associated stream
      * @param audioData buffer containing input data or a place to put output data
      * @param numFrames number of frames to be processed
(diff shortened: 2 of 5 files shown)


dturner:
Just a thought, by passing const references aren't we inhibiting the user defined implementation of onAudioReady from making changes to the stream (e.g changing the buffer size)?

mnaganov:
> Just a thought, by passing const references aren't we inhibiting the user defined implementation of onAudioReady from making changes to the stream (e.g changing the buffer size)?

That touches a large philosophical topic of "logical const-ness" and real immutability (e.g. when the object actually resides in ROM memory). If one is passing an object by a const reference then indeed inside the function he can only call methods marked as "const", but they potentially can make changes to the state of the physical stream, depending on how the object is designed.

However, if we want to emphasize that the passed in object can be changed by the function, then we need to consider it as an "input/output" parameter, and use a pointer:

https://google.github.io/styleguide/cppguide.html#Function_Parameter_Ordering

_... while output and input/output parameters will be pointers to non-const._

philburk:
@Mikhail, to clarify, it sounds like the two options are "const reference" or "pointer". Since we can't do const then you recommend a pointer.  Is "non-const reference" an option?

mnaganov:
@philburk Yes, that's correct. Non-const references are generally not allowed by the Google C++ style guide: _"we never allow non-const reference parameters except when required by convention, e.g., swap()"_

BTW, when you intend to reference me, you need to use @mnaganov, otherwise you are bothering some other user :)

dturner:
Agreed, closing. Although I learned a bit about refs vs pointers so that's something :)

PR #63, opened 2018-02-14
oboe: add ErrorOrValue dual return

philburk:
For returning an error code and a frame count, or other value, at the same time.

### DIFF
--- a/include/oboe/ErrorOrValue.h
+++ b/include/oboe/ErrorOrValue.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2016 The Android Open Source Project
+ * Copyright (C) 2018 The Android Open Source Project
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -29,8 +29,8 @@ public:
             , mError(error) {}
 
     explicit ErrorOrValue(T value)
-            : mValue(value < 0 ? 0 : value)
-            , mError(value < 0 ? static_cast<Result>(value) : oboe::Result::OK) {}
+            : mValue(value)
+            , mError(oboe::Result::OK) {}
 
     oboe::Result error() const {
         return mError;
@@ -41,12 +41,21 @@ public:
     }
 
     /**
-     * Quick way to check for an error.
-     * @return true if an error occurred  // TODO does this seem backwards?
+     * @return true if OK
      */
-    explicit operator bool() const { return mError != oboe::Result::OK; }
+    explicit operator bool() const { return mError == oboe::Result::OK; }
 
-    bool operator !() const { return mError == oboe::Result::OK; }
+    /**
+     * Quick way to check for an error.
+     *
+     * The caller could write something like this:
+     * <code>
+     *     if (!result) { printf("Got error %s\n", convertToText(result.error())); }
+     * </code>
+     *
+     * @return true if an error occurred
+     */
+    bool operator !() const { return mError != oboe::Result::OK; }
 
 private:
     const T             mValue;
--- a/include/oboe/ErrorOrValue.h
+++ b/include/oboe/ErrorOrValue.h
@@ -1,0 +1,58 @@
+/*
+ * Copyright (C) 2016 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


include/oboe/ErrorOrValue.h
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright (C) 2016 The Android Open Source Project
```
mnaganov: 2018

include/oboe/ErrorOrValue.h
```diff
@@ -0,0 +1,58 @@
...
+        return mError;
+    }
+
+    T value() const {
+        return mValue;
+    }
+
+    /**
+     * Quick way to check for an error.
+     * @return true if an error occurred  // TODO does this seem backwards?
```
mnaganov: To me, "backwards" seems more natural:

`if (!write(...)) { // handle error }`

Thus, operator! should return 'true' if there was an error.

include/oboe/ErrorOrValue.h
```diff
@@ -0,0 +1,58 @@
...
+
+template <typename T>
+class ErrorOrValue {
+public:
+    explicit ErrorOrValue(oboe::Result error)
+            : mValue{}
+            , mError(error) {}
+
+    explicit ErrorOrValue(T value)
+            : mValue(value < 0 ? 0 : value)
```
mnaganov: That's not generic. What if 'T' is bool, or a string? In fact, it's the responsibility of the function that returns ErrorOrValue to determine the error condition, and use the appropriate type to pass to the constructor. If the constructor that takes 'T' is used, ErrorOrValue should assume that it was a success.

include/oboe/ErrorOrValue.h
```diff
@@ -0,0 +1,58 @@
...
+            , mError(value < 0 ? static_cast<Result>(value) : oboe::Result::OK) {}
+
+    oboe::Result error() const {
+        return mError;
+    }
+
+    T value() const {
+        return mValue;
```
mnaganov: Should check if there is no error, and emit a log message.

src/aaudio/AudioStreamAAudio.cpp
```diff
@@ -282,25 +282,29 @@ Result AudioStreamAAudio::requestStop() {
...
-int32_t AudioStreamAAudio::write(const void *buffer,
+ErrorOrValue<int32_t>   AudioStreamAAudio::write(const void *buffer,
                                      int32_t numFrames,
                                      int64_t timeoutNanoseconds) {
     AAudioStream *stream = mAAudioStream.load();
     if (stream != nullptr) {
-        return mLibLoader->stream_write(mAAudioStream, buffer, numFrames, timeoutNanoseconds);
+        int32_t result = mLibLoader->stream_write(mAAudioStream, buffer,
+                                                  numFrames, timeoutNanoseconds);
+        return ErrorOrValue<int32_t>(result);
```
mnaganov: As I've noted, it's the responsibility of `AudioStreamAAudio::write` to implement the logic that considers whether the value of `result` indicates an error or not. `ErrorOrValue` must not know these details.

include/oboe/ErrorOrValue.h
philburk: Done

include/oboe/ErrorOrValue.h
philburk: Done

include/oboe/ErrorOrValue.h
philburk: Done

src/aaudio/AudioStreamAAudio.cpp

include/oboe/ErrorOrValue.h
philburk: Why?

mnaganov: Might be useful if a programmer claims that "oboe doesn't work," and then from the logs it turns out they are not checking the result, and there was an error. Treble for example has even stronger checks where retrieving the value crashes if the error status hasn't been checked before value retrieval. But that feels a bit too harsh, a simple log should be enough.

philburk: @mnaganov - I do not understand this request. Do you want to log every access to the value?!

mnaganov: Only if there was an error (mError != OK). Accessing the value if there was an error indicates a problem with client code.

src/opensles/AudioStreamBuffered.cpp
```diff
@@ -98,23 +98,23 @@ int64_t AudioStreamBuffered::predictNextCallbackTime() {
 // TODO: Consider returning an error_or_value struct instead.
```
dturner: Nit: Remove this comment now method returns ErrorOrValue

include/oboe/ErrorOrValue.h
philburk: I see. Good idea. https://github.com/google/oboe/issues/76

PR #64, opened 2018-02-15
oboe: add general purpose getSdkVersion() query

philburk:
Use this for deciding whether to support AAudio.

### DIFF
--- a/src/aaudio/AudioStreamAAudio.cpp
+++ b/src/aaudio/AudioStreamAAudio.cpp
@@ -89,12 +89,6 @@ AudioStreamAAudio::~AudioStreamAAudio() {
 }
 
 bool AudioStreamAAudio::isSupported() {
-    const int SDK_8_1 = 27; // OC-MR1
-    if (getSdkVersion() < SDK_8_1) {
-        // See https://github.com/google/oboe/issues/40,
-        // AAudio is not stable enough on Android 8.0.
-        return false;
-    }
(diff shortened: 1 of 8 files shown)


src/aaudio/AudioStreamAAudio.cpp
```diff
@@ -83,24 +83,14 @@ AudioStreamAAudio::AudioStreamAAudio(const AudioStreamBuilder &builder)
...
-  if (__system_property_get("ro.build.version.sdk", sdk) != 0) {
-      return atoi(sdk) >= 27;  // SDK Level 27 is Android Oreo 8.1
-  }
-#endif
-  return false;
-}
-
 bool AudioStreamAAudio::isSupported() {
-    if (!isRunningOnAndroid8_1OrHigher()) {
+    const int SDK_8_1 = 27; // OC-MR1
```
mnaganov: Please rename to "ANDROID_8_1". People usually associate "SDK" with API versions, so "SDK 8.1" looks strange.

philburk:
@dturner - Should this be getAndroidApiVersion() instead of getSdkVersion() ?

mnaganov:
In android.os.Build, this number is called "The SDK version of the software currently running on this hardware device." They also call it "API level" (see https://developer.android.com/guide/topics/manifest/uses-sdk-element.html for an example). So if you don't like 'getSdkVersion', the correct alternative is 'getAndroidApiLevel'.

PR #74, opened 2018-03-27
Set deviceId to kUnspecified for OpenSL ES.

### LINKED ISSUE #54: setDeviceId does not change anything
Hello! Thanks a lot for this project. I tried it for streaming input and output, and it works great. But I wanted to change mic source for input stream by setDeviceId(int sourceFromAudioManager) and it did not work, no effect for different values. Ids were given   from AudioManager.java (api 23). For my app this behavior is very clitical. Are there any suggestions ?
P.S. Sory for my english :)

philburk:
Fixes #54

### DIFF
--- a/include/oboe/AudioStreamBuilder.h
+++ b/include/oboe/AudioStreamBuilder.h
@@ -179,6 +179,9 @@ public:
      *
      * By default, the primary device will be used.
      *
+     * Note that when using OpenSL ES, this will be ignored and the subsequent
+     * stream will have deviceId kUnspecified.
+     *
      * @param deviceId device identifier or kUnspecified
      * @return pointer to the builder so calls can be chained
      */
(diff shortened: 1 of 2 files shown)


include/oboe/AudioStreamBuilder.h
```diff
@@ -179,6 +179,9 @@ class AudioStreamBuilder : public AudioStreamBase {
      *
      * By default, the primary device will be used.
      *
+     * Note that when using OpenSL ES, this will be ignored and the subsequent
```
mnaganov: subsequent -> created?

mnaganov: Is there an easy way for the client to find out whether Oboe will be using AAudio or OpenSL ES? I haven't found one.

Since the result of choosing a wrong device will be audible to end user, I suggest providing a better service here other than simply ignoring client's requests.

We can introduce "capabilities" that say which operations are possible and which are not. Though, that looks like a big change so can be addressed later. It's actually even better than simply telling the app if Oboe uses AAudio or OpenSL ES because there can be "partial" cases like we have with MMap mode.

philburk:
For some reason GitHub will not let me rebase locally. I only want to change one word in a comment. So I am going to merge this as is so we can move forward.

PR #75, opened 2018-03-29
Add RhythmGame sample 

dturner:
Adding a new sample which shows how to create a simple musical game. Please read the README before reviewing: 

https://github.com/google/oboe/tree/rhythmgame/samples/RhythmGame
 
The main files which need reviewing are: 

samples/RhythmGame/README.md
samples/RhythmGame/src/main/cpp/*


samples/RhythmGame/src/main/cpp/utils/LockFreeQueue.h
```diff
@@ -0,0 +1,142 @@
...
+    /**
+     * Implementation details:
+     *
+     * We have 2 counters: readCounter and writeCounter. Each will increment until it reaches
+     * INDEX_TYPE_MAX, then wrap to zero. Unsigned integer overflow is defined behaviour in C++.
+     *
+     * Each time we need to access our data array we call mask() which gives us the index into the
+     * array. This approach avoids having a "dead item" in the buffer to distinguish between full
+     * and empty states. It also allows us to have a size() method which is easily calculated.
+     */
```
mnaganov: You also need to mention that the implementation expects only 1 reader thread and 1 writer thread. Otherwise you'll have much more complications.

samples/RhythmGame/src/main/cpp/utils/LockFreeQueue.h
```diff
@@ -0,0 +1,142 @@
...
+         *
+         * But because we're returning an unsigned type return value will be as follows:
+         *
+         * returnValue = INDEX_TYPE_MAX - (write - read)
+         *
+         * e.g. if write is 0, read is 150 and the INDEX_TYPE is uint8_t where the max value is
+         * 255 the return value will be (255 - (0 - 150)) = 105.
+         *
+         */
+        return writeCounter - readCounter;
```
mnaganov: There is a concurrency problem with code that attempts to use both counters at the "same time"—here and in 'isEmpty.' Since two consecutive reads from atomic variables is not atomic, you can end up with inconsistent results.

It's safer to store both counters in a single atomic variable of double size, and use atomic 'read-modify-update' operations on it. The drawback is potential cache line sharing between two cores, but since your main goal here is just avoiding locking and not achieving super performance, it's fine.

samples/RhythmGame/src/main/cpp/utils/UtilityFunctions.cpp
```diff
@@ -0,0 +1,79 @@
...
+        return TapResult::Late;
+    }
+}
+
+
+void renderEvent(TapResult r){
+
+    switch (r) {
+        case TapResult::Success:
+            SetGLScreenColor(GREEN);
```
mnaganov: These colors should also be part of the game color scheme.

samples/RhythmGame/src/main/cpp/utils/UtilityFunctions.h
```diff
@@ -0,0 +1,66 @@
...
+
+#ifndef RHYTHMGAME_TIMEFUNCTIONS_H
+#define RHYTHMGAME_TIMEFUNCTIONS_H
+
+#include <stdint.h>
+
+
+constexpr int kMillisecondsInSecond = 1000;
+constexpr int kMicrosecondsInMillisecond = 1000;
+constexpr int kNanosecondsInMillisecond = 1000000;
```
mnaganov: These constant need to be int64_t otherwise you can easily get overflows during conversions on a 32-bit architecture.

samples/RhythmGame/src/main/cpp/utils/logging.h
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright 2018 Google Inc.
```
mnaganov: Again Google.

samples/RhythmGame/src/main/cpp/utils/logging.h
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+#ifndef ANDROID_LOGGING_H
+#define ANDROID_LOGGING_H
+
+#include <stdio.h>
+#include <android/log.h>
+#include <vector>
+
+#define LOGD(...) ((void)__android_log_print(ANDROID_LOG_DEBUG, "rhythm_game", __VA_ARGS__))
```
mnaganov: Create a define for the tag name.

samples/RhythmGame/src/main/java/com/donturner/rhythmgame/GameSurfaceView.java
```diff
@@ -0,0 +1,82 @@
...
+        else*/
+        setEGLContextClientVersion(2);
+        mRenderer = new RendererWrapper();
+        // Set the Renderer for drawing on the GLSurfaceView
+        setRenderer(mRenderer);
+    }
+
+    @Override
+    public void surfaceDestroyed(SurfaceHolder holder) {
+        super.surfaceDestroyed(holder);
```
mnaganov: Does it make sense to run our code first, and then the base class code in this case?

samples/RhythmGame/src/main/java/com/donturner/rhythmgame/GameSurfaceView.java
```diff
@@ -0,0 +1,82 @@
...
+
+    public static native void native_onTouchInput(int eventType, long timeSinceBootMs, int pixel_x, int pixel_y);
+    public static native void native_surfaceDestroyed();
+    private final RendererWrapper mRenderer;
+
+    public GameSurfaceView(Context context) {
+        super(context);
+
+        // Create an OpenGL ES context
+        /*if(BuildConfig.FLAVOR.equals("GL_3"))
```
mnaganov: Please remove commented out code for clarity.

samples/RhythmGame/src/main/java/com/donturner/rhythmgame/RendererWrapper.java
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright 2017 Google Inc.
```
mnaganov: 2018, AOSP?

samples/RhythmGame/test/testLockFreeQueue.cpp
```diff
@@ -0,0 +1,161 @@
...
+ * SIZE:
+ *  - correct zero
+ *  - correct once
+ *  - correct many
+ *
+ * CANNOT TEST:
+ *  - read/writes correct after wraparound (takes too long to run)
+ *  - Constructing queue with non power of 2 capacity results in assert failure (needs death test)
+ *  - Constructing queue with signed type results in assert failure (needs death test)
+ *  - Multi-threaded read/writes (complicated to test)
```
mnaganov: MT is essential to test here, otherwise there is a big risk of running into very confusing problems at run time only.

samples/RhythmGame/src/main/cpp/Game.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+    mClap = SoundRecording::loadFromAssets(mAssetManager, "CLAP.raw");
+    mBackingTrack = SoundRecording::loadFromAssets(mAssetManager, "FUNKY_HOUSE.raw" );
+    mBackingTrack->setPlaying(true);
+    mBackingTrack->setLooping(true);
+
+    mMixer.addTrack(mClap);
+    mMixer.addTrack(mBackingTrack);
+
+    mClapEvents.push(0);
```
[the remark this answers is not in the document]
dturner: Done

samples/RhythmGame/src/main/cpp/Game.cpp
```diff
@@ -0,0 +1,111 @@
...
+    mClapEvents.push(24000);
+    mClapEvents.push(48000);
+
+    mClapWindows.push(96000);
+    mClapWindows.push(120000);
+    mClapWindows.push(144000);
+
+    AudioStreamBuilder builder;
+    builder.setFormat(AudioFormat::I16);
```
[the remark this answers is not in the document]
dturner: True but chaining the calls makes the code less maintainable because it's slightly more difficult to add and remove stream properties. Is there any advantage, other than not prefixing with `builder.`?

samples/RhythmGame/src/main/cpp/Game.cpp
```diff
@@ -0,0 +1,111 @@
...
+    builder.setChannelCount(2);
+    builder.setSampleRate(48000);
+    builder.setCallback(this);
+
+    builder.setPerformanceMode(PerformanceMode::LowLatency);
+    builder.setSharingMode(SharingMode::Exclusive);
+
+    builder.openStream(&mAudioStream);
+
+    mAudioStream->setBufferSizeInFrames(mAudioStream->getFramesPerBurst() * 2);
```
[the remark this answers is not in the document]
dturner: I was deliberately skipping them when I live coded this example, however I shouldn't in this official sample. Fixed.

samples/RhythmGame/src/main/cpp/Game.cpp
```diff
@@ -0,0 +1,111 @@
...
+}
+
+void Game::tap(int64_t eventTimeAsUptime) {
+    mClap->setPlaying(true);
+
+    int64_t nextClapWindowFrame;
+    if (mClapWindows.pop(nextClapWindowFrame)){
+
+        int64_t frameDelta = nextClapWindowFrame - mCurrentFrame;
+        int64_t timeDelta = convertFramesToMillis(frameDelta, 48000);
```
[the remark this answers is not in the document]
dturner: Done

samples/RhythmGame/src/main/cpp/Game.cpp
```diff
@@ -0,0 +1,111 @@
...
+    }
+}
+
+void Game::tick(){
+
+    TapResult r;
+    if (mUiEvents.pop(r)) {
+        renderEvent(r);
+    } else {
+        SetGLScreenColor(GREY);
```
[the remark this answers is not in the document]
dturner: Done

samples/RhythmGame/src/main/cpp/Game.h
```diff
@@ -0,0 +1,58 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef RHYTHMGAME_GAME_H
+#define RHYTHMGAME_GAME_H
+
+#include <android/asset_manager.h>
```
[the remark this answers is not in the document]
dturner: Done

samples/RhythmGame/src/main/cpp/Game.h
```diff
@@ -0,0 +1,58 @@
...
+#include <audio/SoundRecording.h>
+#include <audio/Mixer.h>
+#include <utils/LockFreeQueue.h>
+#include <utils/UtilityFunctions.h>
+
+using namespace oboe;
+
+class Game : public AudioStreamCallback {
+public:
+    Game(AAssetManager *assetManager);
```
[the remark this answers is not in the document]
dturner: Done

samples/RhythmGame/src/main/cpp/Game.h
```diff
@@ -0,0 +1,58 @@
...
+
+    void start();
+    void onSurfaceCreated();
+    void onSurfaceDestroyed();
+    void onSurfaceChanged(int widthInPixels, int heightInPixels);
+    void tick();
+    void tap(int64_t eventTimeAsUptime);
+
+    DataCallbackResult
```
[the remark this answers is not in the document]
dturner: Done

samples/RhythmGame/src/main/cpp/Game.h
```diff
@@ -0,0 +1,58 @@
...
+    onAudioReady(AudioStream *oboeStream, void *audioData, int32_t numFrames) override;
+
+private:
+    AAssetManager *mAssetManager;
+    AudioStream *mAudioStream;
+    SoundRecording *mClap;
+    SoundRecording *mBackingTrack;
+    Mixer mMixer;
+
+    LockFreeQueue<int64_t, 1024> mClapEvents;
```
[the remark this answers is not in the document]
dturner: Done

samples/RhythmGame/src/main/cpp/audio/RenderableAudio.h
```diff
@@ -0,0 +1,30 @@
...
+#ifndef RHYTHMGAME_AUDIORENDERER_H
+#define RHYTHMGAME_AUDIORENDERER_H
+
+
+#include <cstdint>
+
+class RenderableAudio {
+
+public:
+    virtual void renderAudio(int16_t *audioData, int32_t numFrames) = 0;
```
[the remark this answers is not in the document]
dturner: Good spot, thanks. Done.

samples/RhythmGame/src/main/cpp/audio/SoundRecording.cpp
```diff
@@ -0,0 +1,75 @@
+/*
+ * Copyright 2017 The Android Open Source Project
```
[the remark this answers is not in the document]
dturner: Done

samples/RhythmGame/src/main/cpp/audio/SoundRecording.cpp
```diff
@@ -0,0 +1,75 @@
...
+        }
+
+        for (int i = 0; i < numFrames; ++i) {
+            for (int j = 0; j < mChannelCount; ++j) {
+                targetData[(i*mChannelCount)+j] = mData[(mReadFrameIndex*mChannelCount)+j];
+            }
+            mReadFrameIndex++;
+
+            // Handle wraparound
+            if (mReadFrameIndex >= mTotalFrames) mReadFrameIndex = 0;
```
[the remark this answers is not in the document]
dturner: Done

samples/RhythmGame/src/main/cpp/audio/SoundRecording.h
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright 2017 The Android Open Source Project
```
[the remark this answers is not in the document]
dturner: Done

samples/RhythmGame/src/main/cpp/audio/SoundRecording.h
```diff
@@ -0,0 +1,53 @@
...
+    void resetPlayHead() { mReadFrameIndex = 0; };
+    void setPlaying(bool isPlaying) { mIsPlaying = isPlaying; resetPlayHead(); };
+    void setLooping(bool isLooping) { mIsLooping = isLooping; };
+
+    static SoundRecording * loadFromAssets(AAssetManager *assetManager, const char * filename);
+
+private:
+    int32_t mChannelCount = 2; // TODO: move this into a konstant and maybe add as parameter to ctor
+    int32_t mReadFrameIndex = 0;
+    const int16_t* mData;
```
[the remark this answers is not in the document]
dturner: Done

samples/RhythmGame/src/main/cpp/native-lib.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include <android/asset_manager_jni.h>
+#include <android/input.h>
+#include <utils/logging.h>
+#include <memory>
+
+#include "Game.h"
+
+extern "C" {
+
+std::shared_ptr<Game> game;
```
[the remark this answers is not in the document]
dturner: No I most certainly could not ;-) Changed to unique_ptr

samples/RhythmGame/src/main/cpp/ui/OpenGLFunctions.cpp
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright 2017 Google Inc.
```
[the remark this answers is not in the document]
dturner: Done

samples/RhythmGame/src/main/cpp/native-lib.cpp
```diff
@@ -0,0 +1,51 @@
+#include <jni.h>
```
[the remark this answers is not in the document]
dturner: Done

samples/RhythmGame/src/main/cpp/ui/OpenGLFunctions.cpp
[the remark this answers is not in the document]
dturner: Changed

samples/RhythmGame/src/main/cpp/ui/OpenGLFunctions.h
```diff
@@ -0,0 +1,57 @@
+/*
+ * Copyright 2018 Google Inc.
```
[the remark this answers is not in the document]
dturner: Changed

samples/RhythmGame/src/main/cpp/utils/LockFreeQueue.h
dturner: This is mentioned in the comment for the class (line 24), I'll explicitly mention it here as well as it's quite easy to miss.

samples/RhythmGame/src/main/cpp/utils/LockFreeQueue.h
dturner: Now that is clever.

samples/RhythmGame/src/main/cpp/utils/UtilityFunctions.cpp
dturner: Done

samples/RhythmGame/src/main/cpp/utils/UtilityFunctions.h
dturner: Done

samples/RhythmGame/src/main/cpp/utils/logging.h
dturner: Done

samples/RhythmGame/src/main/cpp/utils/logging.h
dturner: Done

samples/RhythmGame/src/main/java/com/donturner/rhythmgame/GameSurfaceView.java
dturner: Yes, it probably does. Changed.

samples/RhythmGame/src/main/java/com/donturner/rhythmgame/GameSurfaceView.java
dturner: Done

samples/RhythmGame/src/main/java/com/donturner/rhythmgame/RendererWrapper.java
dturner: Done

samples/RhythmGame/src/main/cpp/Game.cpp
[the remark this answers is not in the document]
mnaganov: Omitting 'builder' removes clutter, so the code is easier to read. If you choose the first parameter to be something that is always needed to be set, and use a line per chained call, then adding / removing setters is no more difficult than in your version. Compare:

builder.setA();
builder.setB();
builder.setC();

builder.setA()
  ->setB()
  ->setC();

[Personally I would prefer setters that return "*this", so it's always "." prefix, but I guess we are stuck with what we have.]

But finally, up to you—there is no style guide for this.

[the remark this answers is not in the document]
dturner: When you add a setter you need to remove the trailing `;` from the last line (in this case `->setC();` so there is some marginal extra overhead. Also, I find the difference between the initial `.` and subsequent `->`s to be more visually cluttering (and slightly more confusing: "oh wait, a set method returns a pointer?") than the multiple instances of `builder.`. Not that bothered either way, but will probably leave as is.

samples/RhythmGame/src/main/cpp/Game.h
```diff
@@ -19,16 +19,31 @@
...
+constexpr int kMaxQueueItems = 1024; // Must be power of 2
+constexpr ScreenColor kScreenBackgroundColor = GREY;
+constexpr ScreenColor kTapSuccessColor = GREEN;
+constexpr ScreenColor kTapEarlyColor = ORANGE;
+constexpr ScreenColor kTapLateColor = PURPLE;
+
+// This defines the size of the tap window in milliseconds. For example, if defined at 100ms the
+// player will have 100ms before and after the centre of the tap window to tap on the screen and
+// be successful
+constexpr int kWindowCenterOffset = 100;
```
mnaganov: Please add "Ms" suffix.

samples/RhythmGame/src/main/cpp/Game.h
```diff
@@ -19,16 +19,31 @@
...
-#include <utils/LockFreeQueue.h>
-#include <utils/UtilityFunctions.h>
+
+#include "audio/Mixer.h"
+#include "audio/SoundRecording.h"
+#include "ui/OpenGLFunctions.h"
+#include "utils/LockFreeQueue.h"
+#include "utils/UtilityFunctions.h"
+
+constexpr int kSampleRate = 48000; // Fixed sample rate, see README
```
mnaganov: Please add "Hz" suffix.

samples/RhythmGame/src/main/cpp/audio/RenderableAudio.h
```diff
@@ -14,17 +14,18 @@
...
+#ifndef RHYTHMGAME_RENDERABLEAUDIO_H
+#define RHYTHMGAME_RENDERABLEAUDIO_H
 
 
 #include <cstdint>
 
 class RenderableAudio {
 
 public:
+    virtual ~RenderableAudio(){};
```
mnaganov: nit: It is recommended to put "= default" instead of using empty body.

samples/RhythmGame/src/main/cpp/Game.cpp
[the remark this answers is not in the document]
mnaganov: Sure. I agree that returning a pointer from the setters hurts the looks of client code.

samples/RhythmGame/src/main/cpp/utils/LockFreeQueue.h
dturner: Been giving this quite a bit of thought. Whilst it's correct that "two consecutive reads from atomic variables is not atomic" I don't think it's a problem here because both `isFull()` and `isEmpty()` are used pessimistically. 

An example.

1. The queue starts full
1. Writer thread tries to push an item onto the queue by calling `push()`
1. Writer thread calls `isFull()`, which via `size()` performs the non-atomic operation by reading both `readCounter` and `writeCounter`.
1. At this exact same time the reader thread attempts to pop an item off the queue by calling `pop()`.
1. Reader thread calls `isEmpty()`, which via `size()` performs the non-atomic operation by reading both `readCounter` and `writeCounter`. 

This is a race with 2 outcomes: 

Outcome 1: Push fails

1. The writer thread finishes calling `size()` before the reader thread has updated `readCounter` in which case `isFull()` is true and the push fails. This is an acceptible outcome because the concurrent pop hasn't yet completed.

Outcome 2: Push succeeds

1. The reader thread finishes calling `size()` before the writer thread and updates `readCounter`
2. On the writer thread `isFull()` returns false (there is one space in the queue now) and the push succeeds

Am I missing something?

samples/RhythmGame/test/testLockFreeQueue.cpp
dturner: Done

samples/RhythmGame/src/main/cpp/Game.h
dturner: Done

samples/RhythmGame/src/main/cpp/Game.h
dturner: Done

samples/RhythmGame/src/main/cpp/audio/RenderableAudio.h
dturner: Done. nits = better code 👍

samples/RhythmGame/src/main/cpp/utils/LockFreeQueue.h
mnaganov: What about the situation when the writer thread initiates `push', updates `writeCounter', then gets preempted. `isEmpty' now returns `true', but the data isn't in the buffer yet, so the reader receives rubbish. To fix that, `push' must first do a write, and only then 'publish' the result by increasing the counter.

A similar problem with `pop'—after it has increased `readCounter', it enables the writer thread to overwrite the data it is about to read (because `ifFull' will now be returning `false'). So again, `pop' first must read the data, and only then 'publish' this fact.

Apologies for distracting your attention on the non-problem of pair inconsistency, it just was the first thing that had struck me.

mnaganov: Don, I think you can take a look at the implementation here: https://www.di.ens.fr/~guatto/papers/sbac13.pdf
What you are implementing is called "Lamport Queue," and this article provides a basic implementation using C11 atomics. You can continue to use C++11 atomic with default sequential consistency mode, it will just be even less effective than their implementation, but that should not be a problem here.

dturner: Good point, and thank you greatly for taking the time to review. 

I thought that the postincrement operator would only increment the counter *after* the rest of the operators have been evaluated, however this is incorrect since the `a++` operator has a [precedence of 2](http://en.cppreference.com/w/cpp/language/operator_precedence) and `=` has a precedence of 16.  

So if I change the following line in `push`: 

    buffer[mask(writeCounter++)] = item;

To

    buffer[mask(writeCounter)] = item;
    ++writeCounter;

This should mean that data is only 'published' after it has actually been written, and vice versa for `pop` - does that solve the problem? 

Thanks for the Lamport Queue link. Looks pretty good, may use if this doesn't pan out.

One other question: Can you think of a way I could get a failing test for the scenario you mentioned? My current one passes every time, presumably because the chance of the writer thread being preempted immediately after `a++` is very small.

dturner: Thanks Mikhail. Code updated accordingly. Happy for me to merge?

PR #84, opened 2018-04-29
oboe: add calculateLatencyMillis()

### LINKED ISSUE #69: Add method to obtain latency of a stream
This was moved from an internal Android bug tracker.

For many developers knowing the latency of a stream is more important than obtaining very low latency. 

For example, in karaoke apps, the latency of the recording and playback streams must be known in order to synchronise the user's recorded voice over the backing track. 

Another example is recording in Audio Workstation apps (DAWs), the recorded audio must be synchronised with the current track. Doing this requires knowing the latency of the recording and playback streams.  

We could add the following method to a utility.

int32_t  getLatency()

which returns the latency of the stream in microseconds or nanoseconds (milliseconds does not have sufficient precision).

philburk:
Based on getTimestamp().

Fixes #69

### DIFF
--- a/include/oboe/AudioStream.h
+++ b/include/oboe/AudioStream.h
@@ -161,19 +161,24 @@ public:
     /**
      * Calculate the latency of a stream based on getTimestamp().
      *
-     * Latency is the time it takes for a given frame to travel from the
-     * app to the edge of the device or vice versa.
+     * Output latency is the time it takes for a given frame to travel from the
+     * app to some type of digital-to-analog converter. If the DAC is external, for example
+     * in a USB interface or a TV connected by HDMI, then there may be additional latency
+     * that the Android device is unaware of.
      *
-     * Note that the latency of an OUTPUT stream will increase when you write data to it
-     * and then decrease over time.
+     * Input latency is the time it takes to a given frame to travel from an analog-to-digital
+     * converter (ADC) to the app.
      *
-     * The latency of an INPUT stream will decrease when you read data from it
-     * and then increase over time.
+     * Note that the latency of an OUTPUT stream will increase abruptly when you write data to it
+     * and then decrease slowly over time as the data is consumed.
+     *
+     * The latency of an INPUT stream will decrease abruptly when you read data from it
+     * and then increase slowly over time as more data arrives.
      *
      * The latency of an OUTPUT stream is generally higher than the INPUT latency
-     * because an tries to keep the OUTPUT buffer full and the INPUT buffer empty.
+     * because an app generally tries to keep the OUTPUT buffer full and the INPUT buffer empty.
      *
-     * @return The latency in millisecondssec and Result::OK, or a negative error.
+     * @return The latency in milliseconds and Result::OK, or a negative error.
      */
     virtual ErrorOrValue<double> calculateLatencyMillis() {
         return ErrorOrValue<double>(Result::ErrorUnimplemented);
--- a/include/oboe/AudioStream.h
+++ b/include/oboe/AudioStream.h
@@ -158,9 +158,30 @@ public:
 
     virtual int64_t getFramesRead() const { return mFramesRead; }
 
+    /**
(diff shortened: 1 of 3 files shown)


include/oboe/AudioStream.h
```diff
@@ -158,9 +158,30 @@ class AudioStream : public AudioStreamBase {
 
     virtual int64_t getFramesRead() const { return mFramesRead; }
 
+    /**
+     * Calculate the latency of a stream based on getTimestamp().
+     *
+     * Latency is the time it takes for a given frame to travel from the
+     * app to the edge of the device or vice versa.
+     *
+     * Note that the latency of an OUTPUT stream will increase when you write data to it
```
mnaganov: Do you mean the results provided by this function, or the actual latency (e.g. measured using a dedicated rig)?

include/oboe/AudioStream.h
```diff
@@ -158,9 +158,30 @@ class AudioStream : public AudioStreamBase {
...
+     * and then decrease over time.
+     *
+     * The latency of an INPUT stream will decrease when you read data from it
+     * and then increase over time.
+     *
+     * The latency of an OUTPUT stream is generally higher than the INPUT latency
+     * because an tries to keep the OUTPUT buffer full and the INPUT buffer empty.
```
mnaganov: "an tries to keep" needs to be fixed.

src/aaudio/AudioStreamAAudio.cpp
```diff
@@ -390,4 +390,46 @@ Result AudioStreamAAudio::getTimestamp(clockid_t clockId,
...
+        // Calculate the number of frames between app and hardware
+        int64_t frameIndexDelta = appFrameIndex - hardwareFrameIndex;
+
+        // Calculate the time which the next frame will be or was presented
+        int64_t frameTimeDelta = (frameIndexDelta * oboe::kNanosPerSecond) / getSampleRate();
+        int64_t appFrameHardwareTime = hardwareFrameHardwareTime + frameTimeDelta;
+
+        // The current latency is the difference in time between when the current frame is at
+        // the app and when it is at the hardware.
+        int64_t latencyNanos = (isOutput)
```
mnaganov: Change the type to `double` so you don't have to coerce later.

src/aaudio/AudioStreamAAudio.cpp
```diff
@@ -390,4 +390,46 @@ Result AudioStreamAAudio::getTimestamp(clockid_t clockId,
     }
 }
 
+ErrorOrValue<double> AudioStreamAAudio::calculateLatencyMillis() {
+    AAudioStream *stream = mAAudioStream.load();
+    if (stream != nullptr) {
```
mnaganov: I suggest doing an early bailout if the stream is null instead of having an extra indent for the main logic.

include/oboe/AudioStream.h
```diff
@@ -158,9 +158,30 @@ class AudioStream : public AudioStreamBase {
 
     virtual int64_t getFramesRead() const { return mFramesRead; }
 
+    /**
+     * Calculate the latency of a stream based on getTimestamp().
+     *
+     * Latency is the time it takes for a given frame to travel from the
+     * app to the edge of the device or vice versa.
```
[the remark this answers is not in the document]
philburk: The "edge" could be a DAC or a USB port or an HDMI port.  I would like to find a better word than edge.

include/oboe/AudioStream.h
philburk: Both. The actual output latency is a sawtooth function that jumps up when you write data then slowly falls as the data is consumed.

include/oboe/AudioStream.h
philburk: Done

src/aaudio/AudioStreamAAudio.cpp

src/aaudio/AudioStreamAAudio.cpp

Issue #81, opened 2018-04-10
implement getFramesRead() for OpenSL ES


philburk:
Maybe use SLPlayItf::getPosition() and a MonotonicCounter.

PR #89, opened 2018-05-10
Implement getFramesRead for OpenSL ES OUTPUT

philburk:
and getFramesWritten for OpenSL ES INPUT.
Query position in milliseconds from OpenSL ES.
Use MonotonicCounter to build 64-bit frame counter from
32-bit internal counter.

Fixes Issue #81

### DIFF
--- a/src/common/MonotonicCounter.h
+++ b/src/common/MonotonicCounter.h
@@ -1,0 +1,110 @@
+/*
+ * Copyright 2016 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef COMMON_MONOTONIC_COUNTER_H
+#define COMMON_MONOTONIC_COUNTER_H
+
+#include <cstdint>
+
+/**
+ * Maintain a 64-bit monotonic counter.
+ * Can be used to track a 32-bit counter that wraps or gets reset.
+ *
+ * Note that this is not atomic and has no interior locks.
+ * A caller will need to provide their own exterior locking
+ * if they need to use it from multiple threads.
+ */
+class MonotonicCounter {
+
+public:
+    MonotonicCounter() {};
+    virtual ~MonotonicCounter() {};
+
+    /**
+     * @return current value of the counter
+     */
+    int64_t get() const {
+        return mCounter64;
+    }
+
+    /**
+     * set the current value of the counter
+     */
+    void set(int64_t counter) {
+        mCounter64 = counter;
+    }
+
+    /**
+     * Advance the counter if delta is positive.
+     * @return current value of the counter
+     */
+    int64_t increment(int64_t delta) {
+        if (delta > 0) {
+            mCounter64 += delta;
+        }
+        return mCounter64;
+    }
+
+    /**
+     * Advance the 64-bit counter if (current32 - previousCurrent32) > 0.
+     * This can be used to convert a 32-bit counter that may be wrapping into
+     * a monotonic 64-bit counter.
+     *
+     * This counter32 should NOT be allowed to advance by more than 0x7FFFFFFF between calls.
+     * Think of the wrapping counter like a sine wave. If the frequency of the signal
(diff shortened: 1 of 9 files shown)


src/common/MonotonicCounter.h
```diff
@@ -0,0 +1,110 @@
...
+     */
+    void roundUp64(int32_t period) {
+        if (period > 0) {
+            int64_t numPeriods = (mCounter64 + period - 1) / period;
+            mCounter64 = numPeriods * period;
+        }
+    }
+
+private:
+    int64_t mCounter64 = 0;
```
philburk: Maybe should be atomic.

src/opensles/AudioInputStreamOpenSLES.cpp
```diff
@@ -205,10 +206,17 @@ Result AudioInputStreamOpenSLES::requestStop() {
         result = Result::ErrorInvalidState; // TODO review
     } else {
         setState(StreamState::Stopping);
+        mPositionMillis.reset32(); // OpenSL ES resets its millisecond position when stopped.
     }
     return result;
 }
 
+int64_t AudioInputStreamOpenSLES::getFramesWritten() const {
+    int64_t millis64 = mPositionMillis.get();
```
philburk: Extract common function and call from OutputStream

src/opensles/AudioOutputStreamOpenSLES.cpp
```diff
@@ -210,10 +215,26 @@ Result AudioOutputStreamOpenSLES::requestStop() {
         result = Result::ErrorInvalidState; // TODO review
     } else {
         setState(StreamState::Stopping);
+        mPositionMillis.reset32(); // OpenSL ES resets its millisecond position when stopped.
+        int64_t framesWritten = getFramesWritten();
+        if (framesWritten >= 0) {
+            setFramesRead(framesWritten);
```
philburk: This is supposed to prevent drift but does not seem to be working.

src/common/MonotonicCounter.h
```diff
@@ -0,0 +1,110 @@
+/*
+ * Copyright 2016 The Android Open Source Project
```
mnaganov: 2018

src/common/MonotonicCounter.h
```diff
@@ -0,0 +1,110 @@
...
+     * This is used if your counter32 has been reset to zero.
+     */
+    void reset32() {
+        mCounter32 = 0;
+    }
+
+    /**
+     * Round 64-bit counter up to a multiple of the period.
+     *
+     * @param period might be, for example, a buffer capacity
```
mnaganov: Specify that it has to be a positive integer.

src/common/MonotonicCounter.h
mnaganov: Atomicity isn't for free--there are costs on synchronizing CPU caches. Also, in `update32` we seem to update both 32-bit and 64-bit counters in one transaction. This will require making the whole pair to be atomic.

src/common/MonotonicCounter.h
philburk: The code was written in 2016. I copied it from inside AAudio.

src/common/MonotonicCounter.h
mnaganov: Make a struct and declare it atomic. Depending on different conditions it may use a mutex internally, you can check that with `is_lock_free` function: http://en.cppreference.com/w/cpp/atomic/atomic/is_lock_free

For two counters, it should usually be lock-free, unless it's a 32-bit ARM CPU and the fields alignment is wrong.

philburk: Mikhail and I talked. The caller is only updating the internal data from one thread at a time. So I will document that this is not thread-safe and require caller to guarantee synchronization.

src/opensles/AudioInputStreamOpenSLES.cpp
philburk: Done

src/common/MonotonicCounter.h
philburk: Done

src/common/MonotonicCounter.h
philburk: Done

src/opensles/AudioOutputStreamOpenSLES.cpp
philburk: Is no longer drifting.

src/opensles/AudioStreamOpenSLES.cpp
```diff
@@ -170,3 +170,9 @@ SLresult AudioStreamOpenSLES::registerBufferQueueCallback() {
 int32_t AudioStreamOpenSLES::getFramesPerBurst() {
     return mFramesPerBurst;
 }
+
+int64_t AudioStreamOpenSLES::getFramesProcessedByServer() const {
+    int64_t millis64 = mPositionMillis.get();
+    int64_t framesRead = millis64 * getSampleRate() / kMillisPerSecond;
```
philburk: change to framesProcessed

PR #98, opened 2018-05-24
Add MegaDrone sample

dturner:
I presented a version of this code at AES 144 in Milan on Weds 23rd May. There were several requests to publish the code. This seems like the logical place. 

@mnaganov I would really welcome your review on the use of `std::shared_ptr` in `Synth.h` and `Synth.cpp`


dturner:
BTW worth reading the readme first: https://github.com/google/oboe/blob/megadrone/samples/MegaDrone/README.md

samples/MegaDrone/src/main/cpp/Mixer.h
```diff
@@ -0,0 +1,57 @@
...
+
+template <typename T>
+class Mixer : public RenderableAudio<T> {
+
+public:
+
+    void renderAudio(T *audioData, int32_t numFrames) {
+
+        // Zero out the incoming container array
+        for (int j = 0; j < numFrames; ++j) {
```
mnaganov: If this is performance critical then `memset(audioData, 0, sizeof(T) * numFrames)` is preferred. Works OK for both integers and floats.

samples/MegaDrone/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+add_library( megadrone SHARED
+        src/main/cpp/native-lib.cpp
+        src/main/cpp/Synth.cpp)
+
+target_link_libraries( megadrone log oboe )
+
+# Make the Oboe public headers available to our app
+include_directories (${OBOE_DIR}/include)
+
+set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -Ofast")
```
philburk: Might as well optimize _RELEASE as well. -Ofast is faster than -O2.

samples/MegaDrone/src/main/cpp/Mixer.h
```diff
@@ -0,0 +1,57 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef RHYTHMGAME_MIXER_H
+#define RHYTHMGAME_MIXER_H
+
+#include "RenderableAudio.h"
+
+constexpr int32_t kBufferSize = 192*10; // Temporary buffer is used for mixing
```
philburk: add spaces

samples/MegaDrone/src/main/cpp/Synth.cpp
```diff
@@ -0,0 +1,96 @@
...
+#include "MonoToStereo.h"
+
+void Synth::start() {
+
+    AudioStreamBuilder builder;
+    builder.setCallback(this);
+    builder.setPerformanceMode(PerformanceMode::LowLatency);
+    builder.setSharingMode(SharingMode::Exclusive);
+
+    builder.openStream(&mStream);
```
philburk: check return code. It can easily fail.

samples/MegaDrone/src/main/cpp/MonoToStereo.h
```diff
@@ -0,0 +1,45 @@
...
+public:
+
+    MonoToStereo(RenderableAudio<T> *input) : mInput(input){};
+
+    void renderAudio(T *audioData, int32_t numFrames) override {
+
+        // We assume that audioData has sufficient frames to hold the stereo output, so copy each
+        // frame in the input to the output twice
+        // e.g. 123 => 112233
+        for (int i = 0; i < numFrames; ++i) {
```
mnaganov: This seems to be rather ineffective. Why not to render entire `numFrames` fragment first, and then do the expansion moving backwards:

       r          r          r
    [123xxx] -> [123x33] -> [122233] -> [112233]
         ww        ww        ww

samples/MegaDrone/src/main/cpp/Oscillator.h
```diff
@@ -0,0 +1,96 @@
...
+                    audioData[i] = -mAmplitude;
+                } else {
+                    audioData[i] = mAmplitude;
+                }
+
+                mPhase += mPhaseIncrement;
+                if (mPhase > kTwoPi) mPhase -= kTwoPi;
+            }
+        } else {
+            for (int i = 0; i < numFrames; ++i) {
```
mnaganov: Again, consider using `memset`.

samples/MegaDrone/src/main/cpp/Synth.h
```diff
@@ -0,0 +1,56 @@
...
+    std::array<Oscillator<int16_t>, kNumOscillators> mOscsInt16;
+    std::shared_ptr<Mixer<int16_t>> mMixerInt16;
+    std::shared_ptr<RenderableAudio<int16_t>> mOutputStageInt16;
+
+    // Rendering objects for floating point output
+    std::array<Oscillator<float>, kNumOscillators> mOscsFloat;
+    std::shared_ptr<Mixer<float>> mMixerFloat;
+    std::shared_ptr<RenderableAudio<float>> mOutputStageFloat;
+
+
```
philburk: extra lines

samples/MegaDrone/src/main/cpp/Oscillator.h
```diff
@@ -0,0 +1,96 @@
...
+                audioData[i] = 0;
+            }
+        }
+    };
+
+private:
+    std::atomic<bool> mIsWaveOn{false};
```
mnaganov: nit: inconsistent style, should be `{ false }`

samples/MegaDrone/src/main/java/com/example/oboe/megadrone/MainActivity.java
```diff
@@ -0,0 +1,58 @@
...
+    static {
+        System.loadLibrary("megadrone");
+    }
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_main);
+
+        startEngine();
```
philburk: maybe startEngine() in onStart() and stopEngine() in onStop().

samples/MegaDrone/src/main/cpp/Synth.h
```diff
@@ -0,0 +1,56 @@
...
+    void tap(bool isOn);
+
+    DataCallbackResult
+    onAudioReady(AudioStream *oboeStream, void *audioData, int32_t numFrames) override;
+
+private:
+    AudioStream *mStream;
+
+    // Rendering objects for 16-bit integer output
+    std::array<Oscillator<int16_t>, kNumOscillators> mOscsInt16;
```
mnaganov: Have you considering making Synth a template class parametrized by the data type? This duplication of fields and the corresponding logic can be avoided.

samples/MegaDrone/src/main/cpp/Synth.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+    const float baseOscFrequency = 116.0;
+    const float divisor = 33;
+    const float amplitude = 0.009;
+
+    if (mStream->getFormat() == AudioFormat::Float){
+
+        mMixerFloat = std::make_unique<Mixer<float>>();
```
mnaganov: `make_shared` ?

samples/MegaDrone/src/main/cpp/Mixer.h
```diff
@@ -0,0 +1,57 @@
...
+        }
+    }
+
+    void addTrack(RenderableAudio<T> *renderer){
+        mTracks[mNextFreeTrackIndex++] = renderer;
+    }
+
+private:
+    T mixingBuffer[kBufferSize]; // TODO: smart pointer
+    RenderableAudio<T>* mTracks[kMaxTracks]; // TODO: this might be better as a linked list for easy track removal
```
mnaganov: I guess, this needs to be an array of `shared_ptr` if you want `renderAudio` to run safely even if the original owner of the inputs has been destroyed. Also note that in this case `addTrack` needs to take a `shared_ptr`. See this note in http://en.cppreference.com/w/cpp/memory/shared_ptr:

> The ownership of an object can only be shared with another shared_ptr by copy constructing or copy assigning its value to another shared_ptr. Constructing a new shared_ptr using the raw underlying pointer owned by another shared_ptr leads to undefined behavior.

This also applies to other classes that store pointers to `RenderableAudio`.

samples/MegaDrone/CMakeLists.txt
dturner: Done

samples/MegaDrone/src/main/cpp/Mixer.h
dturner: Sorry, not entirely sure where you mean. After the `;`? After `"RenderableAudio.h"`

samples/MegaDrone/src/main/cpp/Synth.h
dturner: Good idea. I have converted Synth into a template class, and refactored the stream creation code into the `AudioEngine` class. `AudioEngine` has pointers to a float and an int16 version of `Synth`. In 3 methods it has to perform a check to see which one is currently in use. I can't figure out a way of avoiding this.

samples/MegaDrone/src/main/cpp/Synth.cpp
dturner: No longer relevant - code refactored

samples/MegaDrone/src/main/cpp/Mixer.h
dturner: Nice! Done.

samples/MegaDrone/src/main/cpp/Oscillator.h
dturner: Done

samples/MegaDrone/src/main/cpp/Oscillator.h
dturner: Done

samples/MegaDrone/src/main/cpp/Synth.h
dturner: Refactored the code so no longer relevant

samples/MegaDrone/src/main/java/com/example/oboe/megadrone/MainActivity.java
dturner: Done. However, I'm getting a segmentation fault (Pixel XL, Android 8.1) when `startEngine()` is called for a 2nd time (after bringing the app out of the background), even though, or maybe because `mStream` has been closed previously. Any ideas?

samples/MegaDrone/src/main/cpp/AudioEngine.h
```diff
@@ -0,0 +1,47 @@
...
+
+    DataCallbackResult
+    onAudioReady(AudioStream *oboeStream, void *audioData, int32_t numFrames) override;
+
+    void stop();
+
+private:
+
+    AudioStream *mStream = nullptr;
+    std::unique_ptr<Synth<float>> mSynthFloat;
```
mnaganov: You can get rid of dual variables and method calls using runtime polymorphism. Declare an interface `ISynth` that has `setWaveOn` and `renderAudio(void*, int)` abstract methods. Make `Synth` template inherit it. Use `ISynth` here instead.

samples/MegaDrone/src/main/java/com/example/oboe/megadrone/MainActivity.java
mnaganov: Could you please share the crash stack? Also maybe add some logging to verify your hypothesis about the stream.

samples/MegaDrone/src/main/cpp/AudioEngine.h
dturner: Of course! This is so much cleaner, thanks!

samples/MegaDrone/src/main/cpp/Oscillator.h
```diff
@@ -0,0 +1,92 @@
...
+constexpr double kDefaultFrequency = 440.0;
+constexpr int32_t kDefaultSampleRate = 48000;
+constexpr double kPi = M_PI;
+constexpr double kTwoPi = kPi * 2;
+
+template <typename T>
+class Oscillator : public RenderableAudio<T> {
+
+public:
+    ~Oscillator(){};
```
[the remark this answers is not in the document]
dturner: Thanks. I've removed this destructor as it was redundant.

PR #103, opened 2018-06-07
Adding loopback example ( echo ) sample

ggfan:


samples/echo/src/main/cpp/AudioEffect.cpp
```diff
@@ -0,0 +1,277 @@
...
+ * @param samples points to PCM audio buffer
+ * @param sampleCount is total samples pointed by samples
+ * @param channelCount channels for PCM audio pointed by samples
+ * @param freq is PCM audio frequency (48000hz for this sample)
+ */
+void  AudioMixer::addStream(std::unique_ptr<int16_t[]>samples, size_t sampleCount,
+          int32_t sampleRate, int32_t channelCount, oboe::AudioFormat format){
+  if (busy_) {
+    LOGW("filtering in progress, filter configuration is IGNORED");
+    return;
```
mnaganov: It would be better to have a return code to indicate this.

samples/echo/src/main/cpp/AudioEffect.cpp
```diff
@@ -0,0 +1,277 @@
...
+  feedbackFactor_ = static_cast<int32_t>(decay_ * kFloatToIntMapFactor);
+  liveAudioFactor_ = kFloatToIntMapFactor - feedbackFactor_;
+  allocateBuffer();
+}
+
+/**
+ * Destructor
+ */
+AudioDelay::~AudioDelay() {
+  if(buffer_) delete static_cast<uint8_t*>(buffer_);
```
mnaganov: Two notes:

- `delete` actually checks whether the pointer is `nullptr`, so no need to do a check here;

- if the type of buffer is actually known (uint8_t), why store it as `void*`? You could instead store it as `std::unique_ptr<uint8_t[]>` and avoid manual deletion.

samples/echo/src/main/cpp/AudioEffect.cpp
```diff
@@ -0,0 +1,277 @@
...
+  std::lock_guard<std::mutex> lock(lock_);
+
+  if(buffer_) {
+    delete static_cast<uint8_t*>(buffer_);
+    buffer_ = nullptr;
+  }
+
+  delay_  = delay;
+  allocateBuffer();
+  return buffer_ != nullptr;
```
mnaganov: I don't see how `buffer_` could be `nullptr` here. The current implementation of `allocateBuffer` will crash the app if any problem occurs.

samples/echo/src/main/cpp/AudioEffect.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+  // process every sample
+  int32_t sampleCount = channelCount * numFrames;
+  int16_t* samples =  & static_cast<int16_t*>(buffer_)[curPos_];
+  int32_t curSample;
+  for (size_t idx = 0; idx < sampleCount; idx++) {
+#if 1
+    curSample = (samples[idx] * feedbackFactor_ +
+                liveAudio[idx] * liveAudioFactor_) / kFloatToIntMapFactor;
+#else
```
mnaganov: Remove this?

samples/echo/src/main/cpp/EchoAudioEngine.h
```diff
@@ -0,0 +1,97 @@
+/*
+ * Copyright 2017 The Android Open Source Project
```
mnaganov: 2018

samples/echo/src/main/cpp/EchoAudioEngine.h
```diff
@@ -0,0 +1,97 @@
...
+  int32_t playbackDeviceId_ = oboe::kUnspecified;
+  oboe::AudioFormat format_ = oboe::AudioFormat::I16;
+  int32_t sampleRate_ = kLoopbackSampleRate;
+  int32_t inputChannelCount_ = kStereoChannelCount;
+  int32_t outputChannelCount_ = kStereoChannelCount;
+  oboe::AudioStream *recordingStream_ = nullptr;
+  oboe::AudioStream *playStream_ = nullptr;
+  int32_t framesPerBurst_;
+  std::mutex restartingLock_;
+  std::unique_ptr<AudioMixer> mixerEffect_ = nullptr;
```
mnaganov: No need to initialize `unique_ptr` with a null pointer, it will do it by itself.

samples/echo/src/main/cpp/jni_bridge.cpp
```diff
@@ -0,0 +1,174 @@
+/**
+ * Copyright 2017 The Android Open Source Project
```
mnaganov: 2018

samples/echo/src/main/cpp/AudioEffect.cpp
```diff
@@ -0,0 +1,277 @@
...
+ * Insert a raw PCM audio buffer to blend with live audio
+ *
+ * @param samples points to PCM audio buffer
+ * @param sampleCount is total samples pointed by samples
+ * @param channelCount channels for PCM audio pointed by samples
+ * @param freq is PCM audio frequency (48000hz for this sample)
+ */
+void  AudioMixer::addStream(std::unique_ptr<int16_t[]>samples, size_t sampleCount,
+          int32_t sampleRate, int32_t channelCount, oboe::AudioFormat format){
+  if (busy_) {
```
mnaganov: What would happen if `busy_` flips to `true` right after this check?

ggfan: application needs to retry the command. I will change it to return bool to  indicate the status. 
otherwise have to add a mutex ( it is not that bad either, the delay is using that method ) to let process() try and lock; on addStream() will blocking lock.

let me know for other ways I could borrow

ggfan: I see what you mean: yeah there is a brief window of contention. I will add lock to it

samples/common/audio_common.cpp
```diff
@@ -0,0 +1,120 @@
+/*
+ * Copyright 2017 The Android Open Source Project
```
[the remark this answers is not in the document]
ggfan: done

samples/common/audio_common.cpp
```diff
@@ -0,0 +1,120 @@
...
+#include "audio_common.h"
+#include <logging_macros.h>
+
+static const oboe::AudioFormat audioFormatEnum[] = {
+    oboe::AudioFormat::Invalid,
+    oboe::AudioFormat::Unspecified,
+    oboe::AudioFormat::I16,
+    oboe::AudioFormat::Float,
+};
+static const int32_t audioFormatCount = sizeof(audioFormatEnum)/
```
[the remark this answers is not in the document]
ggfan: done

samples/common/audio_common.cpp
```diff
@@ -0,0 +1,120 @@
...
+const char* FormatToString(oboe::AudioFormat format) {
+    for (int32_t i = 0; i < audioFormatCount; ++i) {
+        if (audioFormatEnum[i] == format)
+            return audioFormatStr[i];
+    }
+    return "UNKNOW_AUDIO_FORMAT";
+}
+
+void PrintAudioStreamInfo(const oboe::AudioStream *stream) {
+#define STREAM_CALL(func) (stream)->func()
```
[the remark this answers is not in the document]
ggfan: ok, changed.

samples/common/audio_common.cpp
```diff
@@ -0,0 +1,120 @@
...
+static const int32_t audioFormatCount = sizeof(audioFormatEnum)/
+                                        sizeof(audioFormatEnum[0]);
+
+static const uint32_t sampleFormatBPP[] = {
+    0xffff,
+    0xffff,
+    16, //I16
+    32, //FLOAT
+};
+uint16_t SampleFormatToBpp(oboe::AudioFormat format) {
```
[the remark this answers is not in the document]
ggfan: default 0 is good one, will change it.  The switch-case is also good.  The only reason is  that the code will still work for a many items, as well as just 2 items: it is a copy and past things for code, just change the data. 

Probably change to c++'s auto& i : array would make more sense, I will give it a shot, you could have a look and decide after I upload it.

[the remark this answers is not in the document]
ggfan: yeah, actually lot of the things are not needed as oboe::ConvertToText() takes care of lot of things.  changed to new style as you pointed out ( will be able to see it in the next upload )

samples/common/audio_common.h
```diff
@@ -0,0 +1,56 @@
...
+
+uint16_t SampleFormatToBpp(oboe::AudioFormat format);
+/*
+ * GetSystemTicks(void):  return the time in micro sec
+ */
+__inline__ uint64_t GetSystemTicks(void) {
+    struct timeval Time;
+    gettimeofday( &Time, NULL );
+
+    return (static_cast<uint64_t>(1000000) * Time.tv_sec + Time.tv_usec);
```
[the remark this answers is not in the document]
ggfan: done

samples/echo/src/main/cpp/AudioEffect.cpp
ggfan: done: changed  to blocking lock too: UI could wait for 2 ms.

samples/echo/src/main/cpp/AudioEffect.cpp
ggfan: the main reason of void* is, someday, have a generic filter class as a base class for different type of format. void* would be for that. It is up to the derived class to cast it back to its own thing.  This code is far from that, I will delete it.

samples/echo/src/main/cpp/AudioEffect.cpp
ggfan: yeah, done!

samples/echo/src/main/cpp/EchoAudioEngine.h
ggfan: done

samples/echo/src/main/cpp/EchoAudioEngine.h
ggfan: done

samples/echo/src/main/cpp/jni_bridge.cpp
ggfan: done

ggfan:
I "think" @mnaganov code reviews have been all address, may you kindly check it out?  thank you so much!

ggfan:
+@tjohns

samples/common/audio_common.cpp
```diff
@@ -0,0 +1,72 @@
...
+      case oboe::AudioFormat::I16:
+        return 16;
+      case oboe::AudioFormat::Float:
+        return 32;
+      default:
+        return 0;
+    }
+}
+
+void PrintAudioStreamInfo(const oboe::AudioStream *stream) {
```
dturner: This is a useful method for everyone using Oboe. I would consider moving it into `/common/Utilities.cpp` and changing the method signature to: 

    template<>
    const char *convertToText<AudioStream*>(AudioStream *stream)

samples/common/audio_common.cpp
```diff
@@ -0,0 +1,72 @@
...
+        LOGI("FramesReadByApp: %d", (int32_t)stream->getFramesRead());
+        LOGI("FramesWriteByDevice: %d", (int32_t)stream->getFramesWritten());
+    }
+}
+
+int64_t timestamp_to_nanoseconds(timespec ts) {
+  return (ts.tv_sec * oboe::kNanosPerSecond) + ts.tv_nsec;
+}
+
+int64_t get_time_nanoseconds(clockid_t clockid) {
```
dturner: For a follow up CL, we already have a method for this in `AudioClock` but it's not public. We should consolidate all "get time" methods into a single, publicly accessible method. https://github.com/google/oboe/issues/112

samples/common/audio_common.cpp
```diff
@@ -0,0 +1,72 @@
...
+  timespec ts;
+  clock_gettime(clockid, &ts);
+  return timestamp_to_nanoseconds(ts);
+}
+
+void ConvertMonoToStereo(int16_t *buffer, int32_t numFrames) {
```
dturner: Again, for a follow up CL. I have a class similar to this in the MegaDrone sample, needs consolidating. https://github.com/google/oboe/issues/113

samples/echo/build.gradle
```diff
@@ -0,0 +1,39 @@
+apply plugin: 'com.android.application'
+
+android {
+    compileSdkVersion 27
+
+    defaultConfig {
+        applicationId 'com.google.sample.oboe.echo'
+        minSdkVersion 24
```
dturner: Oboe samples should all have a minSdk of 16. They should also be tested on a device or emulator running API 16.

samples/echo/build.gradle
```diff
@@ -0,0 +1,39 @@
...
+        targetSdkVersion 27
+        versionCode 1
+        versionName '1.0'
+        ndk {
```
dturner: x86 build variants are useful when testing on emulator

samples/echo/build.gradle
```diff
@@ -0,0 +1,39 @@
...
+            abiFilters 'armeabi-v7a', 'arm64-v8a'
+        }
+        externalNativeBuild {
+            cmake {
+                arguments '-DANDROID_STL=c++_shared', '-DANDROID_TOOLCHAIN=clang',
```
dturner: Why are these needed?

dturner: Here's my build.gradle for MegaDrone: https://github.com/google/oboe/blob/master/samples/MegaDrone/build.gradle

samples/echo/proguard-rules.pro
```diff
@@ -0,0 +1,17 @@
+# Add project specific ProGuard rules here.
```
dturner: Do we need this file?

dturner:
Firstly, thanks very much for porting this over - it works great on my Pixel XL (haven't tried it on any others yet). 

Some feedback: 

- I'm not sure why music is played at the same time? Is this to demonstrate mixing audio streams? If so, this is already covered in the MegaDrone and RhythmGame samples (in quite a lot of depth). 
- Rather than having a "delay" effect, it'd be more effective to have a "real-time" effect, such as auto-tune, reverb, flange, wah etc. This is because we're trying to show the minimum possible latency and delay artificially introduces latency into the signal path. 
- Is there a reason for being able to select the audio API? Is this to demonstrate something in particular?

samples/echo/src/main/cpp/AudioEffect.cpp
```diff
@@ -0,0 +1,291 @@
...
+#define CLAMP_FOR_I16(x) \
+  ((x) > SHRT_MAX ? SHRT_MAX : ((x) < SHRT_MIN ? SHRT_MIN : (x)))
+
+/*
+ * Mixing Audio in integer domain to avoid FP calculation
+ *   (FG * ( MixFactor * 16 ) + BG * ( (1.0f-MixFactor) * 16 )) / 16
+ */
+static const int32_t kFloatToIntMapFactor = 128;
+
+AudioMixer::AudioMixer() : AudioFormat(48000, 2, oboe::AudioFormat::I16) {
```
dturner: Take a look at the files here: https://github.com/google/oboe/tree/master/samples/MegaDrone/src/main/cpp. The mixing and oscillator objects implement a common interface `RenderableAudio`. We should try to put these into a common folder which is shared between the samples. Could be done in a follow up CL though.

dturner:
Couple of other things: 

- Sample needs a decent README
- The samples root README also needs updating

samples/echo/build.gradle
ggfan: yeah got this part: some portion of the UI code needs to be changed and done: will get a device to do a real test: failed to create recording string on api-16 + emulator, could be some bugs

samples/echo/build.gradle
ggfan: got that part, uploaded with the change

samples/echo/build.gradle
ggfan: the default is NOT libc, it is clang.  we should make new samples to use the latest one.

unfortunately on api-16, there is no libc, so application compiled on libc++ will not be able to run. default gnustl is still needs to be used

samples/echo/proguard-rules.pro
ggfan: not really as it is empty: studio generated template file. deleted

ggfan:
The goal is to make one app to have some processing to demonstrate live audio + processing; you have been trying to move developers to use oboe, with this one set up, we could direct developers to this repo when asking questions for openSL ES on other places: it is real one and it works. Helpful in that regard

We certainly could add the processing of complicated filters, that would be next stage ( add in a real complicate filter ) to show it really could work with heavy duty duplex audio.

Selecting API part is optional: it helped in the past for the debugging, I thought developer would like to know and compare -- both should work, and performance might vary. It would emphasize that oboe could support both with the same code, but it could either way, I am fine to remove it from UI, let me know.

ggfan:
removed echo and mixing, only leaves live audio ( loopback ) in it, please review: hope to get it check in soon.

for the sample name, it could be

- as is ( name is a little misleading as it is not echoEffect )
- loopback ( align with CTS of AAudio )
- liveEffect ( as Don proposed  )

thank you again!

samples/LiveEffect/README.md
```diff
@@ -0,0 +1,25 @@
...
+Screenshots
+-----------
+
+![Screenshot](screenshot.png)
+
+
+### Stream Configurations
+- 48kHz
+- oboe::I16
+- steroe or mono
```
dturner: stereo

samples/LiveEffect/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+
+### Caveats
+When first time starting audio devices, the stream may not be stable.
+The symtom is the strange callback pattern. This sample waits half a second
```
dturner: spelling: symptom. Also, this could be worded better.

samples/LiveEffect/src/main/cpp/LiveEffectEngine.cpp
```diff
@@ -42,38 +40,38 @@ EchoAudioEngine::~EchoAudioEngine() {
...
+  if (isEffectOn_)
     return false;
 
   audioApi_ = api;
   return true;
 }
-void EchoAudioEngine::setEchoOn(bool isEchoOn) {
-  if (isEchoOn != isEchoOn_) {
-    isEchoOn_ = isEchoOn;
+void LiveEffectEngine::setEeffectOn(bool isOn) {
```
dturner: typo: setEffectOn

samples/LiveEffect/src/main/cpp/LiveEffectEngine.h
```diff
@@ -14,24 +14,25 @@
...
+#include <oboe/Oboe.h>
 #include <jni.h>
+#include <thread>
 #include <string>
-#include "audio_common.h"
-#include "AudioEffect.h"
 
+constexpr int kMonoChannelCount = 1;
+constexpr int kStereoChannelCount = 2;
 const int32_t kLoopbackSampleRate = 48000;
```
dturner: The sample rate shouldn't be hardcoded, it should be obtained from the output stream, then used to create the inputstream.

samples/README.md
```diff
@@ -6,6 +6,7 @@ These samples demonstrate how to use the Oboe library:
 sine wave when you tap the screen
 1. [RhythmGame](RhythmGame): A simple rhythm game where you copy the clap patterns you hear by tapping on the screen
 1. [MegaDrone](MegaDrone): A one hundred oscillator synthesizer, demonstrates low latency and CPU performance
+1. [LiveEffect](LiveEffect): loops back audio from input stream to output stream to demonstrate duplex capability
```
dturner: Thing is, it's not really a "loop back" as this implies a signal path which circles back on itself. Instead what we have is: 

recording at microphone -> app processing -> playback at speaker

Needs rewording

ggfan:
Let me go through it and get them fixed

samples/LiveEffect/README.md
ggfan: done

samples/LiveEffect/README.md
ggfan: done

samples/LiveEffect/src/main/cpp/LiveEffectEngine.cpp
ggfan: done

samples/LiveEffect/src/main/cpp/LiveEffectEngine.cpp
```diff
@@ -28,12 +27,11 @@
  */
 const float kSystemWarmupTime = 0.5f;
 
-EchoAudioEngine::EchoAudioEngine() {
+LiveEffectEngine::LiveEffectEngine() {
   assert(outputChannelCount_ == inputChannelCount_);
```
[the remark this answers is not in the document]
ggfan: Ok, clang-format -style=google is different then. I will change them along the way.

samples/LiveEffect/src/main/cpp/LiveEffectEngine.h
```diff
@@ -14,24 +14,25 @@
...
 
-#include <thread>
+#include <oboe/Oboe.h>
 #include <jni.h>
+#include <thread>
 #include <string>
-#include "audio_common.h"
-#include "AudioEffect.h"
 
+constexpr int kMonoChannelCount = 1;
```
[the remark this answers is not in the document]
ggfan: create a bug for it: https://github.com/google/oboe/issues/119.  should it be enum or just const, change it at that time might be good. Good point

samples/LiveEffect/src/main/cpp/jni_bridge.cpp
```diff
@@ -0,0 +1,117 @@
...
+ * limitations under the License.
+ */
+
+#include <jni.h>
+#include <logging_macros.h>
+#include "LiveEffectEngine.h"
+
+static LiveEffectEngine *engine = nullptr;
+
+static const int OBOE_API_AAUDIO = 0;
```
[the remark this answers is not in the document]
ggfan: done

samples/LiveEffect/src/main/cpp/LiveEffectEngine.h
ggfan: The usage would be better:
  app asks for the sample rate
  app checks for the sample rate really granted
  if not granted with different rate, do something else

The reasoning would be: if app wants to mix audio, it would already have certain sample rate in preference; it may not be exactly the same as the platform's default sample rate. If platform could support app's sample rate, app would not have to re-sample. It is more app friendly rather than leave all work to app. In this sample, it assert that the sample rate is what asked for. What do you think?

samples/README.md
ggfan: yeah, that is right no "back", just a loop: drop the "back". for loopback I thought Phils cts coins the same term, am I correct?

ggfan:
@dturner I think pretty much all is addressed for the code review, please have one more glance.
will wait for CI to finish: if no error, and no other things left, please help merge it.  

thank you so much! :-)

PR #104, opened 2018-06-07
Expose header directory to users via CMakeLists.txt

ggfan:
by PUBLIC for include directory, it will be automatically included for users using add_subdirectory(); it would also copy to the release directory if there is configures ( I think ).

a little convenience for users: it seems older versions were using public, latest one changed it to private


ggfan:
@dturner @philburk may you also take a look at this one? It should not have negative impact.  Thanks

dturner:
Thanks @ggfan!

PR #105, opened 2018-06-08
add const modifier to oboe::AudioStream::getFramesPerBurst()

ggfan:
It is more of a question: most of "get" functions are const, this getFramePerBurst() is not. Is this intentional? after adding "const", it still works.

the benefit is that: user does not have to do something like:
  const_cast<AudioStream*>(stream)->getFramesPerBurst() anymore

thanks

### DIFF
--- a/include/oboe/AudioStream.h
+++ b/include/oboe/AudioStream.h
@@ -142,7 +142,7 @@ public:
      *
      * @return burst size
      */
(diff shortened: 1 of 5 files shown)


ggfan:
may you please help merge it in, I do not have merge access right. Thanks!

PR #107, opened 2018-06-11
Adding travis ci to the repo

ggfan:
Adding CI to the repo:

- use samples directory ci every check-in
- only checking build to be successful

The final one should look like this one:
- [repo](https://github.com/ggfan/testOboe)
- [ci-page](https://travis-ci.org/ggfan/testOboe)

If you are interested, let me know. In order for it to work, additional steps:

- enable travis-ci on your repo (only repo owner could do it )
- merge in this pull-request

only then could think of how to do run-time for the built samples ( emulator or firebase ci with devices )

Kindly share your thoughts: if you like it, let's the fun begin

### DIFF
--- a/.ci_tools/build_samples.sh
+++ b/.ci_tools/build_samples.sh
@@ -1,0 +1,44 @@
+# Configurations:
+#  temp file name to hold build result
+BUILD_RESULT_FILE=build_result.txt
+
+# Repo root directory
+REPO_ROOT_DIR=.
+
+
+declare projects=(
+    samples
+)
+
+for d in "${projects[@]}"; do
+    pushd ${REPO_ROOT_DIR}/${d} >/dev/null
+    TERM=dumb ./gradlew  -q clean assembleDebug
+    popd >/dev/null
+done
+
+
+# Check the apks that all get built fine
+# TBD: add echo after it is merged
+#        samples/echo/build/outputs/apk/debug/echo-debug.apk
+declare apks=(
+    samples/hello-oboe/build/outputs/apk/debug/hello-oboe-debug.apk
+    samples/MegaDrone/build/outputs/apk/debug/MegaDrone-debug.apk
+    samples/RhythmGame/build/outputs/apk/debug/RhythmGame-debug.apk
+)
+
+rm -fr ${BUILD_RESULT_FILE}
+for apk in "${apks[@]}"; do
+  if [ ! -f ${REPO_ROOT_DIR}/${apk} ]; then
+    export SAMPLE_CI_RESULT=1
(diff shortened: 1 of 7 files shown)


ggfan:
+ @tjohns, @philburk, @dturner

mnaganov:
I'm not familiar with Travis so I can't tell if it's a good tool or not. The idea of having continuous integration makes sense.

Gerry, could you please confirm—is it for testing changes to Oboe against the sample apps?

ggfan:
any change to the repo will trigger travis-ci; travis-ci will build samples. Since sample build automatically builds oboe-lib ( oboe is a static lib at current time), so oboe and samples are all get built. If there is any interface change in Oboe, samples that use that "changed interface functions/data structures" will need to be changed, otherwise build will fail immediately.

we could also add the ULT into this CI thing if needed (I saw there is a ULT directory).

dturner:
SGTM. Gerry - OK to merge?

ggfan:
Hey @dturner, thanks!  yeah, please help to merge it.

samples/build.gradle has some conflict, but it is really about the a version, and sequence between google() and jcenter(), google() has to be in the first due to an issue on jcenter().

After merge, 1 thing on each side:
- github repo: turn on Travis CI  service for the repo [just turn it on, no need to do anything else, I believe ]
- log in to travis-ci page, find the oboe repo ( it should show up if you are owner of the repo ), turn on CI for it

that should make it work. Thanks!

dturner:
@ggfan - I followed your instructions but I just see "No builds for this repository" on the Travis CI page: https://travis-ci.org/google/oboe

Am I missing something? Does the setup take a while?

dturner:
@ggfan thanks for suggesting this. I actually ran into a failed build just now: https://travis-ci.org/google/oboe/jobs/391818696

However, there wasn't any stack trace or any reason why the build failed (it was my code not compiling). Is there any way of getting more info out of Travis?

ggfan:
the failure seems to be: 0.01s$ source .ci_tools/build_samples.sh
.ci_tools/build_samples.sh: line 15: ./gradlew: Permission denied

gradlew file does not have execution right for everyone, that needs to be added in repo.

For build, yeah, remove TERM=dumb will show lot of the logs.  ( that is inside build_sample.sh ).  probably we could enable it.  One thing is that: travis has a 4MB limit for log, as long as we do not break that limit, we are fine to lift the restriction.

ggfan:
add PR to add x for all for gradlew ( chmod a+x gradlew )
https://github.com/google/oboe/pull/111
CI passes for that PR.

Issue #106, opened 2018-06-11
move docs and PNG files into a /docs/ folder


philburk:
The goal is to have few files at the top level so the README moves up and appears above the fold.

dturner:
Not sure if the README is "above the fold" but it's as high as I can get it without moving/deleting more files. Fixed in #110

PR #110, opened 2018-06-13
Moving docs and images into subfolder

dturner:
Groups the documentation together, rather than being lumped into the root dir. Also requires less scrolling to see the README.


dturner:
Fixes #106

PR #126, opened 2018-06-22
fix link to GettingStarted guide

atsushieno:


atsushieno:
I believe I have entered everything I need for agreement but it's not updating the status. I don't mind if you go ahead and apply the changes by yourselves :-)

dturner:
Thank you for this fix!

Issue #119, opened 2018-06-20
add channel count const or enum into Oboe


ggfan:
this originated from code review:
"These constants should potentially be moved to an enum inside Oboe e.g. oboe::ChannelCount::Stereo"

PR #127, opened 2018-06-22
Adding Stereo and Mono as enum to the interface

ggfan:
this is trying to address https://github.com/google/oboe/issues/119. Please kindly review it.


samples/MegaDrone/src/main/cpp/MonoToStereo.h
```diff
@@ -36,8 +35,8 @@ class MonoToStereo : public RenderableAudio<T> {
         // e.g. 123 => 112233
         for (int i = numFrames - 1; i >= 0; --i) {
 
-            audioData[i * kStereoChannelCount] = audioData[i];
-            audioData[i * kStereoChannelCount + 1] = audioData[i];
+            audioData[i * oboe::ChannelCount::Stereo] = audioData[i];
```
philburk: This used to be an int. Now it is an enum. Doing math on enums is a bit sketchy.

philburk: Nice CL otherwise. Thanks.

include/oboe/Definitions.h
```diff
@@ -312,6 +312,23 @@ namespace oboe {
          Allocate = AAUDIO_SESSION_ID_ALLOCATE,
     };
 
+    enum ChannelCount {
```
dturner: Since channel count is always an int, and we want to safely cast back to an int in calculations I would suggest specifying the underlying type: 

`enum ChannelCount : int32_t {`

dturner:
@ggfan If you can update the underlying type I am happy to merge.

samples/MegaDrone/src/main/cpp/MonoToStereo.h
ggfan: ok, let me change the underlying to int32_t type

include/oboe/Definitions.h
ggfan: this is changed now to the request format. please take a look again.  Thanks

ggfan:
please take a look: changed according to the code review.  Thanks

ggfan:
Phil's requests are already included: merge it for now, thanks!

PR #130, opened 2018-06-26
opensles: generate channel masks safely by OS version

### LINKED ISSUE #117: SL_ANDROID_MAKE_INDEXED_CHANNEL_MASK called but not supported before N
Need to do something else if version < N.

https://github.com/google/oboe/blob/master/src/opensles/AudioInputStreamOpenSLES.cpp#L51

philburk:
Indexed channel masks were not supported before N.

Fixes #117

### DIFF
--- a/src/opensles/AudioInputStreamOpenSLES.cpp
+++ b/src/opensles/AudioInputStreamOpenSLES.cpp
@@ -57,15 +57,19 @@ AudioInputStreamOpenSLES::AudioInputStreamOpenSLES(const AudioStreamBuilder &bui
 AudioInputStreamOpenSLES::~AudioInputStreamOpenSLES() {
 }
 
-int AudioInputStreamOpenSLES::chanCountToChanMask(int channelCount) {
-    // from internal sles_channel_in_mask_from_count(chanCount);
+// Calculate masks specific to INPUT streams.
+SLuint32 AudioInputStreamOpenSLES::channelCountToChannelMask(int channelCount) {
+    // Derived from internal sles_channel_in_mask_from_count(chanCount);
+    // in "frameworks/wilhelm/src/android/channels.cpp".
+    // Yes, it seems strange to use SPEAKER constants to describe inputs.
+    // But that is how OpenSL ES does it internally.
     switch (channelCount) {
         case 1:
             return SL_SPEAKER_FRONT_LEFT;
         case 2:
             return SL_SPEAKER_FRONT_LEFT | SL_SPEAKER_FRONT_RIGHT;
         default:
-            return chanCountToChanMaskDefault(channelCount);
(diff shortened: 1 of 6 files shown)


src/opensles/AudioInputStreamOpenSLES.cpp
```diff
@@ -57,24 +57,15 @@ AudioInputStreamOpenSLES::AudioInputStreamOpenSLES(const AudioStreamBuilder &bui
 AudioInputStreamOpenSLES::~AudioInputStreamOpenSLES() {
 }
 
-#define AUDIO_CHANNEL_COUNT_MAX         30u
-#define SL_ANDROID_UNKNOWN_CHANNELMASK  0
-
 int AudioInputStreamOpenSLES::chanCountToChanMask(int channelCount) {
```
dturner: Why is this method in this *input stream* class when it refers to outputs? It feels like it should be a static/utility method elsewhere

src/opensles/AudioOutputStreamOpenSLES.cpp
```diff
@@ -46,10 +46,10 @@ constexpr int SL_ANDROID_SPEAKER_5DOT1 = (SL_ANDROID_SPEAKER_QUAD
 constexpr int SL_ANDROID_SPEAKER_7DOT1 = (SL_ANDROID_SPEAKER_5DOT1 | SL_SPEAKER_SIDE_LEFT
         | SL_SPEAKER_SIDE_RIGHT);
 
-int AudioOutputStreamOpenSLES::chanCountToChanMask(int chanCount) {
```
dturner: If you're going to update the parameter name it might be worth updating the method name to `channelCountToChannelMask` as well

src/opensles/AudioStreamOpenSLES.cpp
```diff
@@ -49,6 +49,29 @@ AudioStreamOpenSLES::~AudioStreamOpenSLES() {
     delete[] mCallbackBuffer;
 }
 
+#define AUDIO_CHANNEL_COUNT_MAX          30u
```
dturner: Convention is to use `constexpr` and `kAudioChannelCountMax`

philburk: The compiler rejected that. I'll try again.

src/opensles/AudioOutputStreamOpenSLES.cpp
philburk: Good. Done.

src/opensles/AudioStreamOpenSLES.cpp
philburk: Done

dturner:
LGTM

PR #132, opened 2018-06-26
opensles: make Input pause unimplemented

### LINKED ISSUE #88: pause() for INPUT on OpenSL ES should return UNIMPLEMENTED
This is the documented behavior for AAudio and Oboe should follow AAudio API.

We want to minimize the differences between the OPENSL and the AAudio implementations of Oboe.

philburk:
This is to match the behavior of AAudio.

Fixes #88

### DIFF
--- a/src/opensles/AudioInputStreamOpenSLES.cpp
+++ b/src/opensles/AudioInputStreamOpenSLES.cpp
@@ -230,12 +230,14 @@ Result AudioInputStreamOpenSLES::requestStart() {
 
 
 Result AudioInputStreamOpenSLES::requestPause() {
-    LOGD("AudioInputStreamOpenSLES::requestPause() is unavailable for input streams");
-    return Result::ErrorUnimplemented;
+    LOGW("AudioInputStreamOpenSLES::requestPause() is intentionally not implemented for input "
+         "streams");
+    return Result::ErrorUnimplemented; // Matches AAudio behavior.
 }
 
 Result AudioInputStreamOpenSLES::requestFlush() {
-    LOGW("AudioInputStreamOpenSLES::%s() UNIMPLEMENTED", __func__);
+    LOGW("AudioInputStreamOpenSLES::requestFlush() is intentionally not implemented for input "
+         "streams");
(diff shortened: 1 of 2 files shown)


philburk:
Oboe should behave the same for OpenSL ES and for AAudio.
Otherwise apps could behave unexpectedly.
Pause() and Flush() are not implemented for AAudio.
So they should also not be implemented for Oboe on OpenSL ES.

The alternative would be to keep the PAUSE behavior for OpenSL ES and modify Oboe pause() for AAudio to call AAudioStream_requestStop().

dturner:
From meeting today: Alternative option is to call `stop` on AAudio stream when `pause` is called

dturner:
Just trying to close this one out. What should we do here? Is there ever any case where pausing an input stream makes sense?

philburk:
Pausing input does not really make sense to me but it probably does to developers. So we should probably just call stop().

dturner:
@philburk I think the problem with stopping when pause is called is that you would expect the state to go into `Paused` when it's actually in `Stopped`. 

Let's keep things simple. Just return `Error::Unavailable` when attempting to pause an input stream for both APIs. I don't see this causing an issue for developers (famous last words!).

src/opensles/AudioInputStreamOpenSLES.cpp
```diff
@@ -230,24 +230,13 @@ Result AudioInputStreamOpenSLES::requestStart() {
...
-    Result result = setRecordState(SL_RECORDSTATE_PAUSED);
-    if (result == Result::OK) {
-        mPositionMillis.reset32(); // OpenSL ES resets its millisecond position when paused.
-        setState(StreamState::Paused);
-    } else {
-        setState(initialState);
-    }
-    return result;
+    LOGD("AudioInputStreamOpenSLES::requestPause() is unavailable for input streams");
+    return Result::ErrorUnavailable;
```
philburk: Flush() and AAudio input pause return Unimplemented.

dturner: Fixed

dturner:
Tidy up log messages

PR #140, opened 2018-06-28
Refactoring AudioStream::getBufferSizeInFrames to return a ResultWithValue

dturner:
I started updating the `AudioStream::get*` methods to return a `ResultWithValue` based on the assumption that we need to handle the case when the stream is null. 

I really don't like this change because of the extra work and noise which is creates when calling the API. 

In particular, take a look at code changes in [`PlayAudioEngine.cpp`](https://github.com/google/oboe/pull/140/commits/e513142b9009be9435f406c0b6459023dcada7a6#diff-0cab96b97d755d2a27561725775b00e0). Getting and setting the buffer size becomes an arduous job of checking the result of every call. 

The thing which sticks in my mind is that if the stream were null you wouldn't be able to call `pStream->getBufferSizeInFrames()` anyway, so it's pointless to return an `Result::ErrorNull`.

Having tried this API change I would advocate for a solution where we (somehow) guarantee (either in code or in documentation) that the AudioStream is never null. Thoughts?

### DIFF
--- a/include/oboe/AudioStreamBase.h
+++ b/include/oboe/AudioStreamBase.h
@@ -20,6 +20,7 @@
 #include <memory>
 #include "oboe/AudioStreamCallback.h"
 #include "oboe/Definitions.h"
+#include "oboe/ResultWithValue.h"
 
 namespace oboe {
 
@@ -74,15 +75,17 @@ public:
      *
      * @return buffer size or a negative error.
      */
-    virtual int32_t getBufferSizeInFrames() const {
+    virtual ResultWithValue<int32_t> getBufferSizeInFrames() const {
         // By default assume the effective size is the same as capacity.
-        return getBufferCapacityInFrames();
+        return ResultWithValue<int32_t>(getBufferCapacityInFrames());
     }
 
     /**
-     * @return capacityInFrames or OBOE_UNSPECIFIED
+     * @return capacityInFrames or kUnspecified
      */
-    virtual int32_t getBufferCapacityInFrames() const { return mBufferCapacityInFrames; }
+    virtual int32_t getBufferCapacityInFrames() const {
+        return mBufferCapacityInFrames;
+    }
 
     SharingMode getSharingMode() const { return mSharingMode; }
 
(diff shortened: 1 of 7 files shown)


philburk:
I also do not like the extra work for the caller.

> The thing which sticks in my mind is that if the stream were null you wouldn't be able to call
> pStream->getBufferSizeInFrames() anyway, so it's pointless to return an Result::ErrorNull.

No. pStream is not null. If it was you would just crash. The problem is that the AAudioStream pointer inside pStream might be null because another thread might have called pStream->close().
Also OpenSL ES might return UNIMPLEMENTED. So Oboe getters can fail.

The use case for this is a thread that updates the UI by calling getFramesRead(). Another thread can close the stream causing the UI task to crash. I hit this problem in my test app. It is a race condition.

We can modify the AAudio API in Q so that the stream resources can be released without deleting the stream object. But we would still have to deal with O and P.

One solution might be to say that the Oboe getter return values are undefined when the stream is closed. So it would be up to the app not to do that. We could just return zero. Or we could cache the last valid value and just return it after the stream is closed.  Then the UI frame counter might stop changing but that is better than crashing or showing negative frames. Hmm. That might be good.

philburk:
> Having tried this API change I would advocate for a solution where we (somehow) guarantee
> (either in code or in documentation) that the AudioStream is never null. Thoughts?

If the stream is deleted by close() then we have to null the AAudioStream pointer.
And we have to call close() at some point to release the hardware resources associated with the stream.
Otherwise we have a resource leak.

dturner:
> OpenSL ES might return UNIMPLEMENTED

Right, there's 2 separate issues here:

- calling methods which are not implemented on the underlying API like `getTimestamp` 
- calling methods when the underlying stream is null 

I think the first isn't a problem because the affected methods already return a `Result` of some sort. [Spreadsheet AudioStream methods](https://docs.google.com/spreadsheets/d/1VdLpxCuKCxwrdK1-XVfZRVSDH4dTOSLh6jyLjxgjn9k/edit?usp=sharing), and the client can call `isLatencyDetectionSupported` to check first if they want. 

The null stream is the bigger issue

> One solution might be to say that the Oboe getter return values are undefined when the stream is closed. So it would be up to the app not to do that. We could just return zero. Or we could cache the last valid value and just return it after the stream is closed.

Why not both :) We cache the getter return values *and* say that the values are undefined after `requestClose` has been called.

dturner:
Phil and I have come up with a better set of changes to handle this.

PR #146, opened 2018-06-30
Updated all sdks, and also updated to use AndroidX as support library.

AndroidDeveloperLB:


dturner:
Thanks for doing this. Since AndroidX is still in alpha and API 28 is still in developer preview I'm going to hold off doing this merge until both are stable.

AndroidDeveloperLB:
But you wrote me to do this...
"Feel free to submit a pull request :)"
:(

AndroidDeveloperLB:
BTW, Google keeps telling me that Android P DP3 is final API...

AndroidDeveloperLB:
And I think AndroidX is the same as support library, currently, no?

dturner:
Sorry for the confusion. I hadn't realised at the time of writing that the change would introduce non-stable dependencies. I'm currently checking with the various teams on the likelihood of API change.

dturner:
OK, I'm reliably informed that both APIs are locked down and won't change. 

Also, I have verified the samples still run successfully on Pixel XL API 27 emulator. 

Happy to merge - good to keep things current. Thanks for submitting this.

PR #161, opened 2018-07-17
oboe: remove dependencies on P constants

philburk:
This will allow Oboe to compile with old NDKs.

### DIFF
--- a/include/oboe/Definitions.h
+++ b/include/oboe/Definitions.h
@@ -155,11 +155,11 @@ namespace oboe {
         AAudio
     };
 
-// Help compile with old NDK versions.
+// Hard code constants so they can be compiled with versions of the NDK before P.
 #if __ANDROID_API_LEVEL__ >= __ANDROID_API_P__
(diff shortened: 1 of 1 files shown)


include/oboe/Definitions.h
```diff
@@ -155,6 +155,13 @@ namespace oboe {
         AAudio
     };
 
+// Help compile with old NDK versions.
```
dturner: Whilst I understand the reasoning behind this approach it's not particularly readable. My preference would be to have a separate `AAudioDefinitions.h` file which is a copy of the AAudio defines from later NDK versions.

philburk: I'm not clear what you mean. Would we include a different file based on API_LEVEL?

Maybe just define the enums twice
#if API_LEVEL
    enum { GAMES = 4, /* AAUDIO_GAMES */ }
#else 
    enum { GAMES = AAUDIO_GAMES, }
#endif 

It's verbose but readable.

PR #162, opened 2018-07-18
Allow setting of default stream values

dturner:
I've created a new class `DefaultStreamValues` in `Definitions.h` which has public static members for each stream property which we need a default value for.  

The values are initialized in `AudioStreamBuilder.cpp`. 

The client can set their own device-dependent values like this: 

    oboe::DefaultStreamValues::SampleRate = 96000;
    oboe::DefaultStreamValues::FramesPerBurst = 240;
    oboe::DefaultStreamValues::ChannelCount = 8;

These are read when the `AudioStream` is opened if an explicit value hasn't been specified by the client.


src/common/AudioStreamBuilder.cpp
```diff
@@ -26,6 +26,19 @@
 
 namespace oboe {
 
+/**
+ * The following default values are used when oboe does not have any better way of determining the optimal values
+ * for an audio stream. The most common scenarios for this are:
+ *
+ * - Creating a stream on API < 26 (OpenSLES) but has not supplied the optimal sample
```
philburk: The word "has" sounds strange. What has?

tests/testStreamOpen.cpp
```diff
@@ -0,0 +1,73 @@
...
+    AudioStreamBuilder mBuilder;
+    AudioStream *mStream = nullptr;
+
+};
+
+TEST_F(StreamOpen, ForOpenSLESDefaultSampleRateIsUsed){
+
+    DefaultStreamValues::SampleRate = 44100;
+    openStream();
+    if (mStream->getAudioApi() == AudioApi::OpenSLES){
```
philburk: Should request OpenSL ES when you open.

tests/testStreamOpen.cpp
```diff
@@ -0,0 +1,73 @@
...
+        ASSERT_EQ(mStream->getSampleRate(), 44100);
+    }
+    closeStream();
+}
+
+TEST_F(StreamOpen, ForOpenSLESDefaultFramesPerBurstIsUsed){
+
+    DefaultStreamValues::FramesPerBurst = 128;
+    openStream();
```
philburk: request OpenSL ES API

tests/testStreamOpen.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+TEST_F(StreamOpen, ForOpenSLESDefaultChannelCountIsUsed){
+
+    DefaultStreamValues::ChannelCount = 1;
+    openStream();
+    if (mStream->getAudioApi() == AudioApi::OpenSLES){
+        ASSERT_EQ(mStream->getChannelCount(), 1);
+    }
+    closeStream();
+}
```
philburk: add newline

src/common/AudioStreamBuilder.cpp
dturner: Text updated.

tests/testStreamOpen.cpp
dturner: Of course! That simplifies each test nicely.

tests/testStreamOpen.cpp
dturner: done

tests/testStreamOpen.cpp
dturner: done

tests/testStreamOpen.cpp
```diff
@@ -25,8 +25,11 @@ class StreamOpen : public ::testing::Test {
 protected:
 
     bool openStream(){
+        mBuilder.setAudioApi(AudioApi::OpenSLES);
```
philburk: If the whole file uses OpenSL ES then the file name should reflect that. Otherwise just do it in specific tests.

Or pass ApiIndex in openStream call.

dturner:
Docs also need updating

tests/testStreamOpen.cpp
dturner: > Or pass ApiIndex in openStream call.

Done

dturner:
@philburk I think that's all the feedback addressed. OK to merge? Would be good to get this one in as it gets us closer to "optimal latency everywhere"

ggfan:
LGTM

PR #164, opened 2018-07-19
Adding -Ofast for sample build.

ggfan:


samples/RhythmGame/CMakeLists.txt
```diff
@@ -42,4 +42,10 @@ set (OBOE_DIR ../..)
...
 # Specify the path to the Oboe header files.
-include_directories (${OBOE_DIR}/include)
\ No newline at end of file
+include_directories (${OBOE_DIR}/include)
+
+# Enable optimization for release build; it could be enabled for debug build
+# if performance is of interest.
+target_compile_options(native-lib
+    PRIVATE
+        -std=c++14 -Wall -Werror $<$<CONFIG:Release>:-Ofast>)
```
philburk: please add newline

samples/MegaDrone/CMakeLists.txt
```diff
@@ -30,4 +30,3 @@ include_directories (${OBOE_DIR}/include)
 # Append compiler optimization flags to both the debug and release build variants
 set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -Ofast")
 set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -Ofast")
```
philburk: Should we change these to the new technique?

samples/LiveEffect/src/main/cpp/CMakeLists.txt
```diff
@@ -39,6 +39,9 @@ target_link_libraries(liveEffect
         atomic
         log)
 
+# Enable optimization for release build; it could be enabled for debug build
+# if performance is of interest.
 target_compile_options(liveEffect
     PRIVATE
```
philburk: Why is this PRIVATE? Don's suggestion was PUBLIC.
Also "CONFIG:RELEASE" instead of "CONFIG:Release". Is it case sensitive?
Don?

https://github.com/android-ndk/ndk/issues/699#issuecomment-404925125

dturner: I don't think it matters whether it's `PUBLIC` or `PRIVATE` since this is used to define the scope of dependencies ([more here](https://stackoverflow.com/questions/26037954/cmake-target-link-libraries-interface-dependencies)). Probably best to prefer `PRIVATE`.

samples/MegaDrone/CMakeLists.txt
dturner: +1

samples/RhythmGame/CMakeLists.txt
ggfan: fixed, thanks!

samples/LiveEffect/src/main/cpp/CMakeLists.txt
ggfan: It is case insensitive. Changed everything to be upper case for consistence.

ggfan:
please have one more look: did add it into the oboe lib for the same, please review that part, thanks!

Issue #169, opened 2018-07-20
Change C style casts to C++ style


philburk:
Requested here:  https://buganizer.corp.google.com/issues/71872740#comment19

dturner:
Buganizer updated.

PR #173, opened 2018-07-25
Convert c-type cast to C++ cast

ggfan:
this should get rid of issue https://github.com/google/oboe/issues/169. PTAL

### DIFF
--- a/src/common/Utilities.cpp
+++ b/src/common/Utilities.cpp
@@ -36,12 +36,12 @@ void convertFloatToPcm16(const float *source, int16_t *destination, int32_t numS
         float fval = source[i];
         fval += 1.0; // to avoid discontinuity at 0.0 caused by truncation
         fval *= 32768.0f;
-        auto sample = (int32_t) fval;
+        auto sample = static_cast<int32_t>(fval);
         // clip to 16-bit range
         if (sample < 0) sample = 0;
         else if (sample > 0x0FFFF) sample = 0x0FFFF;
         sample -= 32768; // center at zero
-        destination[i] = (int16_t) sample;
+        destination[i] = static_cast<int16_t>(sample);
     }
 }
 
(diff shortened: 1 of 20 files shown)


src/common/Utilities.cpp
```diff
@@ -36,12 +36,12 @@ void convertFloatToPcm16(const float *source, int16_t *destination, int32_t numS
         float fval = source[i];
         fval += 1.0; // to avoid discontinuity at 0.0 caused by truncation
         fval *= 32768.0f;
-        auto sample = (int32_t) fval;
+        auto sample = static_cast<int32_t>(fval + 0.5f);
```
philburk: This changes the behavior of the code from floor to round! Please remove the "+ 0.5f". 
Maybe it should match the similar code in:
"frameworks/av/media/libaaudio/src/utility/AAudioUtilities.cpp"
If so then, let's do that in a separate Pull Request.

src/opensles/AudioStreamOpenSLES.cpp
```diff
@@ -76,7 +76,7 @@ SLuint32 AudioStreamOpenSLES::channelCountToChannelMaskDefault(int channelCount)
 
 static bool s_isLittleEndian() {
     static uint32_t value = 1;
-    return *((uint8_t *) &value) == 1; // Does address point to LSB?
+    return (*reinterpret_cast<uint8_t *>(&value) == 1);
```
philburk: Please restore the comment.

src/common/Utilities.cpp
ggfan: yeah, it is: so please check whether this helps or not ( rounding up or cut-off ). Let me restore it now

ggfan: done reverting, thanks

src/opensles/AudioStreamOpenSLES.cpp
ggfan: done, thanks! ( In old days, comment on the "return..." statement caused compile errors, I was under that impression.)  pointer is a special kind of value, once value is LE, pointer should be LE already, correct?

philburk: The pointer is not little endian. It points to a value that is either big or little endian. This is a trick for querying endianness.

philburk:
Thanks for doing this. But when making multiple changes in one Pull Request, please do **not** use "commit --amend".  Just keep adding commits to the same branch. It makes it very hard to see changes between patch sets.

Issue #172, opened 2018-07-24
[LiveEffect] Logcat warns that input stream is not low latency when it is


dturner:
Steps to repro: 

- Run LiveEffect sample on Nexus 9
- Hit `Start`
- Open logcat

Expected result

- Logcat shows that the input stream is created as a low latency stream and no warnings are shown

Actual result

    com.google.sample.oboe.liveeffect I/AudioRecord: AUDIO_INPUT_FLAG_FAST successful; frameCount 0
    com.google.sample.oboe.liveeffect D/OboeAudio: FifoProcessor: capacityInFrames = 3072, bytesPerFrame = 4
    com.google.sample.oboe.liveeffect W/AUDIO-APP: Stream is NOT low latency.Check your requested format, sample rate and channel count

Note the line: `Stream is NOT low latency.` which should not appear. 

This could indicate that the stream's performance mode is not being set correctly in Oboe

ggfan:
taking a look

philburk:
Ask rago@ about this. It may be because when Effects are turned on you may
not get a fast track. Maybe you have low latency input but not output or
vice versa.

ggfan:
Ok, will do.  I like to see it: Nexus 9 does not support O and later OSes, so this got to be OpenSL ES + Android N, only my guess.  The bug report missed the critical info about OS version, I did not see it on Pixel phone + P...

dturner:
My Nexus 9 is running Android 7.0 NRD91R

ggfan:
OS diff contributes to the different behaviour:
- NRD91R:  nyc_release branch,        dated 2016-12-09, Android 7.0, API 24,  saw the error
- N2G47S: nrc-mr2-release branch, dated 2017-04-03, Android 7.1.2, API 25, NOT seeing the error.

philburk:
There is no guarantee that you will get LowLatency mode when you ask for it. 

I think the problem here is that we were given a FAST track internally but there is no way for Oboe to determine that because it is < N_MR1. So it should just set performanceMode to None, which it does. This is actually WAI.

Note that we should query the final performance mode using GetConfiguration(). But that is not supported for < N_MR1. See https://github.com/google/oboe/issues/179

philburk:
How do you know the input stream is actually Low Latency?

Could this bug be related to #179 ?

PR #177, opened 2018-07-27
Leave AudioStreamBase::mPerformanceMode unchanged on old platforms

ggfan:
this fix issue https://github.com/google/oboe/issues/172. 

Please carefully review the SMALL change, the behavior is: if the feature is NOT supported, mPerformanceMode is not changed, which means whatever user asks, Oboe would pretend working at that mode. 

The facts:
```
on Android N:                                   InputStream    OutputStream
SL_IID_ANDROIDCONFIGURATION/Interface                  Y           N
SL_ANDROID_KEY_PERFORMANCE_MODE/Key                    N           N
```

the code change makes the first *column* to behave the same as the second *column* above.

I believe that this was added by me when checking into liveEffect sample that broke the protocol :-(.


philburk:
> if the feature is NOT supported, mPerformanceMode is not changed, which means
> whatever user asks, Oboe would pretend working at that mode.

We do NOT want that behavior. If the user requests a non-default performance mode, and it doesn't work for any reason, then the final performance mode should be None. 

Also this does not fix #172 because this change only affects devices running < N and that device was running N.

Issue #175, opened 2018-07-27
Logging should be off by default


dturner:
By default Oboe logs a lot of messages to logcat. This is fine for development but may not be desirable in a live environment. 

Oboe should have logging switched off by default. It should be enabled by adding: 

    target_compile_definitions(oboe PRIVATE "$<$<CONFIG:DEBUG>OBOE_ENABLE_LOGGING=1")

to `CMakeLists.txt`

ggfan:
tentatively adding target_compile_options(oboe PUBLIC -DOBOE_ENABLE_LOGGING=1): PUBLIC means that the flag is enabled for the compiling module itself PLUS all depending modules.

dturner:
Fixed in https://github.com/google/oboe/pull/180

dturner:
Fixed in https://github.com/google/oboe/commit/66783b1dd4c6b11e9981fee537e2944bcd7bc825

PR #178, opened 2018-07-27
Adding -Ofast for sample build.

ggfan:
re-open this one as I deleted the branch  for https://github.com/google/oboe/pull/164 accidently

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,12 +30,7 @@ target_include_directories(oboe
  # Compile Flags:
  #     Enable -Werror when building debug config
  #     Enable -Ofast
-target_compile_options(oboe PRIVATE -std=c++11
-                            PRIVATE -Wall
-                            PRIVATE "$<$<CONFIG:DEBUG>:-Werror>"
-                            PRIVATE -Ofast)
-
-# Enable logging when debugging
-# target_compile_options(oboe PUBLIC -DOBOE_ENABLE_LOGGING=1)
+target_compile_options(oboe
+    PRIVATE -std=c++11 -Wall "$<$<CONFIG:DEBUG>:-Werror>" -Ofast)
 
 target_link_libraries(oboe PRIVATE log OpenSLES)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,11 +29,11 @@ target_include_directories(oboe
 
  # Compile Flags:
  #     Enable -Werror when building debug config
- #     Enable -Ofast for release config (could be enabled for debug too)
+ #     Enable -Ofast
 target_compile_options(oboe PRIVATE -std=c++11
                             PRIVATE -Wall
                             PRIVATE "$<$<CONFIG:DEBUG>:-Werror>"
-                            PRIVATE "$<$<CONFIG:RELEASE>:-Ofast>")
+                            PRIVATE -Ofast)
 
 # Enable logging when debugging
 # target_compile_options(oboe PUBLIC -DOBOE_ENABLE_LOGGING=1)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,7 +27,12 @@ target_include_directories(oboe
                            PRIVATE src 
                            PUBLIC include)
 
+ # Compile Flags:
+ #     Enable -Werror when building debug config
+ #     Enable -Ofast for release config (could be enabled for debug too)
 target_compile_options(oboe PRIVATE -std=c++11
                             PRIVATE -Wall
-                            PRIVATE "$<$<CONFIG:DEBUG>:-Werror>") # Only include -Werror when building debug config
+                            PRIVATE "$<$<CONFIG:DEBUG>:-Werror>"
+                            PRIVATE "$<$<CONFIG:RELEASE>:-Ofast>")
+
 target_link_libraries(oboe PRIVATE log OpenSLES)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,4 +35,7 @@ target_compile_options(oboe PRIVATE -std=c++11
                             PRIVATE "$<$<CONFIG:DEBUG>:-Werror>"
                             PRIVATE "$<$<CONFIG:RELEASE>:-Ofast>")
 
+# Enable logging when debugging
+# target_compile_options(oboe PUBLIC -DOBOE_ENABLE_LOGGING=1)
+
 target_link_libraries(oboe PRIVATE log OpenSLES)
--- a/samples/LiveEffect/src/main/cpp/CMakeLists.txt
+++ b/samples/LiveEffect/src/main/cpp/CMakeLists.txt
@@ -42,5 +42,5 @@ target_link_libraries(liveEffect
 # Enable optimization for release build; it could be enabled for debug build
 # if performance is of interest.
 target_compile_options(liveEffect
-    PRIVATE
-        -std=c++14 -Wall -Werror $<$<CONFIG:RELEASE>:-Ofast>)
+    PRIVATE -std=c++14 -Wall -Werror $<$<CONFIG:RELEASE>:-Ofast>)
+
--- a/samples/LiveEffect/src/main/cpp/CMakeLists.txt
+++ b/samples/LiveEffect/src/main/cpp/CMakeLists.txt
@@ -39,8 +39,7 @@ target_link_libraries(liveEffect
         atomic
         log)
 
-# Enable optimization for release build; it could be enabled for debug build
-# if performance is of interest.
-target_compile_options(liveEffect
-    PRIVATE -std=c++14 -Wall -Werror $<$<CONFIG:RELEASE>:-Ofast>)
+# Enable optimization flags: if having problems with source level debugging,
+# disable -Ofast ( and debug ), re-enable it after done debugging.
+target_compile_options(liveEffect PRIVATE -Wall -Werror -Ofast)
 
--- a/samples/LiveEffect/src/main/cpp/CMakeLists.txt
+++ b/samples/LiveEffect/src/main/cpp/CMakeLists.txt
@@ -41,5 +41,5 @@ target_link_libraries(liveEffect
 
 # Enable optimization flags: if having problems with source level debugging,
 # disable -Ofast ( and debug ), re-enable it after done debugging.
-target_compile_options(liveEffect PRIVATE -Wall -Werror -Ofast)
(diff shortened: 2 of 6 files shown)


ggfan:
added flag OBOE_ENABLE_LOGGING for https://github.com/google/oboe/issues/175: @dturner  please really review it.  Thanks!

samples/debug-utils/logging_macros.h
```diff
@@ -18,7 +18,7 @@
 #define __SAMPLE_ANDROID_DEBUG_H__
 #include <android/log.h>
 
-#if 1
+#if OBOE_ENABLE_LOGGING
```
philburk: Why is this change here. It conflicts with an older fix.

samples/MegaDrone/CMakeLists.txt
```diff
@@ -27,7 +27,8 @@ target_link_libraries( megadrone log oboe )
...
 
-# Append compiler optimization flags to both the debug and release build variants
-set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -Ofast")
-set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -Ofast")
-
+# Enable optimization for release build; it could be enabled for debug build
+# if performance is of interest.
+target_compile_options(megadrone
+    PRIVATE
+        -std=c++14 -Wall $<$<CONFIG:RELEASE>:-Ofast>)
```
dturner: Why add `-std=c++14`? I don't think the app needs any C++14 features in which case it makes sense to leave this out.

CMakeLists.txt
```diff
@@ -27,8 +27,12 @@ target_include_directories(oboe
...
 
-target_compile_options(oboe PRIVATE -std=c++11
+ # Compile Flags:
+ #     Enable -Werror when building debug config
+ #     Enable -Ofast for release config (could be enabled for debug too)
+ target_compile_options(oboe PRIVATE -std=c++11
                             PRIVATE -Wall
-                            PRIVATE "$<$<CONFIG:DEBUG>:-Werror>") # Only include -Werror when building debug config
+                            PRIVATE "$<$<CONFIG:DEBUG>:-Werror>"
+                            PRIVATE "$<$<CONFIG:RELEASE>:-Ofast>")
```
dturner: Oboe is currently built from source when a dependent project (e.g. LiveEffect) is built. Is the build configuration (release or debug) used when building Oboe?

samples/LiveEffect/src/main/cpp/CMakeLists.txt
```diff
@@ -39,6 +39,8 @@ target_link_libraries(liveEffect
         atomic
         log)
 
+# Enable optimization for release build; it could be enabled for debug build
+# if performance is of interest.
 target_compile_options(liveEffect
     PRIVATE
-        -Wall -Werror)
+        -std=c++14 -Wall -Werror $<$<CONFIG:RELEASE>:-Ofast>)
```
dturner: Why add -std=c++14? I don't think the app needs any C++14 features in which case it makes sense to leave this out.

samples/MegaDrone/CMakeLists.txt
dturner: Also, for this particular example `-Ofast` should be enabled for both debug and release because it's a performance related sample and we don't want first-time developers thinking there's a bug in the code because of audio glitches in the default (debug) configuration.

ggfan: I thought we have email said that the standard is to use C++14 as a standard configuration, and enable -Ofast for release.

ggfan: [clang -g doc](https://clang.llvm.org/docs/CommandGuide/clang.html#code-generation-options) points out that -g  works best with -o0.  Implying -ofast enabled, source level debugging may have trouble.  I think to enable it for release; for debugging version, leave it out and leave users to add it.  thanks

samples/debug-utils/logging_macros.h
ggfan: this is for this: https://github.com/google/oboe/issues/175. 

add a flag to guard it, then could use compile flag to control it is on or off

samples/MegaDrone/CMakeLists.txt
dturner: I agree that `-O0` should be used in normal circumstances but for the MegaDrone sample it will almost certainly glitch and hence should be left as `-Ofast` for all variants.

CMakeLists.txt
```diff
@@ -27,7 +27,15 @@ target_include_directories(oboe
...
                            PUBLIC include)
 
+ # Compile Flags:
+ #     Enable -Werror when building debug config
+ #     Enable -Ofast
 target_compile_options(oboe PRIVATE -std=c++11
                             PRIVATE -Wall
-                            PRIVATE "$<$<CONFIG:DEBUG>:-Werror>") # Only include -Werror when building debug config
+                            PRIVATE "$<$<CONFIG:DEBUG>:-Werror>"
+                            PRIVATE -Ofast)
```
dturner: Shouldn't this be release=`-Ofast`, debug=`-O0`? 
Also why the multiple `PRIVATE` keywords? Couldn't these all go on one line?

CMakeLists.txt
```diff
@@ -27,7 +27,15 @@ target_include_directories(oboe
...
+
+# Enable logging when debugging
+# target_compile_options(oboe PUBLIC -DOBOE_ENABLE_LOGGING=1)
```
dturner: Set `OBOE_ENABLE_LOGGING=1` for debug configuration and not for release

samples/LiveEffect/src/main/cpp/CMakeLists.txt
```diff
@@ -39,6 +39,7 @@ target_link_libraries(liveEffect
         atomic
         log)
 
-target_compile_options(liveEffect
-    PRIVATE
-        -Wall -Werror)
+# Enable optimization flags: if having problems with source level debugging,
+# disable -Ofast ( and debug ), re-enable it after done debugging.
+target_compile_options(liveEffect PRIVATE -Wall -Werror -Ofast)
```
dturner: Again only -Ofast for release config

samples/MegaDrone/CMakeLists.txt
```diff
@@ -27,7 +27,6 @@ target_link_libraries( megadrone log oboe )
 # Make the Oboe public headers available to our app
 include_directories (${OBOE_DIR}/include)
 
-# Append compiler optimization flags to both the debug and release build variants
-set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -Ofast")
-set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -Ofast")
-
+# Enable optimization flags: if having problems with source level debugging,
+# disable -Ofast ( and debug ), re-enable it after done debugging.
+target_compile_options(megadrone PRIVATE -Wall -Werror -Ofast)
```
dturner: -Ofast for release only

samples/RhythmGame/CMakeLists.txt
```diff
@@ -42,4 +42,11 @@ set (OBOE_DIR ../..)
...
 
 # Specify the path to the Oboe header files.
-include_directories (${OBOE_DIR}/include)
\ No newline at end of file
+include_directories (${OBOE_DIR}/include)
+
+# Enable optimization flags: if having problems with source level debugging,
+# disable -Ofast ( and debug ), re-enable after done debugging.
+target_compile_options(native-lib
+    PRIVATE -std=c++14
```
dturner: Multiple PRIVATE statements

samples/hello-oboe/CMakeLists.txt
```diff
@@ -53,3 +53,7 @@ add_library(hello-oboe SHARED
 
 # Specify the libraries needed for hello-oboe
 target_link_libraries(hello-oboe android log oboe)
+
+# Enable optimization flags: if having problems with source level debugging,
+# disable -Ofast ( and debug ), re-enable after done debugging.
+target_compile_options(hello-oboe PRIVATE -std=c++14 -Wall -Werror -Ofast)
```
dturner: I think the only reason for using C++14 is that the sample is using `std::make_unique` which I believe it isn't.

samples/MegaDrone/CMakeLists.txt
ggfan: done with next checkin, thx

samples/LiveEffect/src/main/cpp/CMakeLists.txt
ggfan: done

CMakeLists.txt
ggfan: add -Ofast for debug, leave default for debug build. Done

CMakeLists.txt
ggfan: restore its original for this one.

samples/LiveEffect/src/main/cpp/CMakeLists.txt
ggfan: done

samples/MegaDrone/CMakeLists.txt
ggfan: first review, you said:
"I agree that -O0 should be used in normal circumstances but for the MegaDrone sample it will almost certainly glitch and hence should be left as -Ofast for all variants."
the same file but conflicting comment from you, which one you are going?

samples/RhythmGame/CMakeLists.txt
ggfan: done

samples/hello-oboe/CMakeLists.txt
ggfan: done

ggfan:
I think the code review requests have been addressed with this PR, please review again.

left -Ofast for release and debug builds for Oboe library as performance has higher priority than source code level debugging, assuming most users would trace to app code, not oboe lib source code. Please double check, it is all just assumption.

Thanks!

samples/MegaDrone/CMakeLists.txt
dturner: My apologies Gerry, this comment was meant for the other samples. Here's what the optimisation flag should be for each sample: 

hello-oboe: debug -O0, release -Ofast
LiveEffect: debug -O0, release -Ofast
RhythmGame: debug -O0, release -Ofast
MegaDrone: all -Ofast

Issue #65, opened 2018-02-26
Do not call requestPause() on AAudio if already paused


philburk:
To avoid state transition problems in AAudio MMAP code. In MR1 this could leave you in a PAUSING state.

See spreadsheet for other transitions to avoid.
Make version dependent.

dturner:
@philburk Do you have repro steps or a bug id for this? I [wrote a test](https://github.com/google/oboe/commit/71b6d41f2d82a1eccbbbccbde2d1f0d2c2242fa5) which I hoped would reproduce this bug and ran it on an emulator running Oreo MR1 (API 27) but it passed.

dturner:
From Phil: This must be tested on an MMAP stream.

philburk:
I found that I had already implemented a fix. It does not cover all state transitions.
For example requestPause() when Stopped will still transition to Pausing instead of Paused.

https://github.com/google/oboe/pull/191

REPRODUCE bug with NativeOboe Tester on Marlin on OC_MR1
click TEST OUTPUT
click OPEN
click START
click PAUSE
note that it displays "state = 6"
click PAUSE

EXPECT
state = 6

ACTUAL
state = 5

PR #191, opened 2018-08-15
aaudio: don't pause if already paused

philburk:
Avoid state transition problems in O and O_MR1.

Bug: #65


src/aaudio/AudioStreamAAudio.cpp
```diff
@@ -269,6 +269,15 @@ Result AudioStreamAAudio::requestStart() {
     std::lock_guard<std::mutex> lock(mLock);
     AAudioStream *stream = mAAudioStream.load();
     if (stream != nullptr) {
+        // Avoid state machine errors in O_MR1.
+#if defined(__ANDROID_API_O_MR1__)
```
philburk: Remove these because we require >= O_MR1 NDK.

PR #194, opened 2018-08-22
Fix #192 - Cannot open recording stream on Moto G

### LINKED ISSUE #192: Cannot open recording stream on Motorola Moto G 2nd Gen
Steps to repro: 

- Run the LiveEffect sample
- Hit Start

Expected outcome: 

- Audio from the microphone is heard through the speaker or headphones

Actual outcome: 

- No audio is heard from the speaker or headphones

The following logcat output shows that the recording stream fails to be created: 

> 07-21 18:51:02.063 16034-16034/com.google.sample.oboe.liveeffect E/libOpenSLES: Cannot create AudioRecorder: data sink must be in PCM format
> 07-21 18:51:02.063 16034-16034/com.google.sample.oboe.liveeffect E/libOpenSLES: Cannot create AudioRecorder: invalid source or sink
> 07-21 18:51:02.063 16034-16034/com.google.sample.oboe.liveeffect W/libOpenSLES: Leaving Engine::CreateAudioRecorder (SL_RESULT_PARAMETER_INVALID)
> 07-21 18:51:02.063 16034-16034/com.google.sample.oboe.liveeffect E/OboeAudio: createAudioRecorder() result:SL_RESULT_PARAMETER_INVALID
> 07-21 18:51:02.063 16034-16034/com.google.sample.oboe.liveeffect E/AUDIO-APP: Failed to create recording stream. Error: ErrorInternal
> 07-21 18:51:02.063 16034-16034/com.google.sample.oboe.liveeffect E/AUDIO-APP: Failed to create recording (0x0) and/or playback (0xb8aef600) stream

dturner:


src/opensles/AudioInputStreamOpenSLES.cpp
```diff
@@ -124,11 +126,29 @@ Result AudioInputStreamOpenSLES::open() {
...
                                                                         &audioSink);
+
+    // If the result is "invalid parameter" and we're using the extended data format it could be
+    // that the underlying hardware doesn't support it
+    // (e.g. Moto G: https://github.com/google/oboe/issues/192)
+    // Try again using the old format
+    if (result == SL_RESULT_PARAMETER_INVALID && usingExtendedDataFormat){
+
+        LOGW("Unable to open recording stream using PCM_EX format, trying again using PCM...");
+        audioSink.pFormat = &format_pcm;
```
philburk: Should we set usingExtendedDataFormat=false here? For line 150 below.

dturner: Doh! Thanks Phil - this is exactly why I ask you to review :)

dturner: Fixed now

src/opensles/AudioInputStreamOpenSLES.cpp
```diff
@@ -124,11 +126,30 @@ Result AudioInputStreamOpenSLES::open() {
...
+    }
+
     if (SL_RESULT_SUCCESS != result) {
         LOGE("createAudioRecorder() result:%s", getSLErrStr(result));
         goto error;
     }
 
+    // If the stream is not using the extended data format it must have a format of I16 since no
+    // other format is supported, regardless of what format was requested.
+    if (!usingExtendedDataFormat) mFormat = AudioFormat::I16;
```
philburk: In AAudio, if you request Format=UNSPECIFIED then you will get either I16 or FLOAT.
If you request FLOAT then you will get a FLOAT stream or the stream will not open.
You do not get an I16 stream. Same for channels.

So we should probably do something like:
    if (formatResuested == UNSPECIFIED) mFormat = I16
    else return ERROR;

dturner: > or the stream will not open

That's interesting. So it's not like PerformanceMode where if you request LowLatency but don't get it the stream still opens but with PerformanceMode=None.

In that case, like you say, we should make the behaviour the same for OpenSL ES streams. Which is (correct me if I'm wrong)...

For playback streams
==

**API < 21 (only I16 supported)**
Unspecified - Stream opens with format set to I16
Float - Stream does not open and returns relevant error (should we add `ErrorUnsupportedAudioFormat` so it's clear the reason the stream didn't open?)
I16 - Stream opens with format set to I16

**API 21+ (I16 and float supported)**
Unspecified - Stream opens with format set to Float
Float - Stream opens with format set to Float
I16 - Stream opens with format set to I16

For recording streams:
==
**API < 23 (only I16 supported)**
Unspecified - Stream opens with format set to I16
Float - Stream does not open and returns relevant error
I16 - Stream opens with format set to I16

**API 23+ (I16 and float supported)**
Unspecified - Stream opens with format set to Float
Float - Stream opens with format set to Float
I16 - Stream opens with format set to I16

dturner:
Since this behaviour is now defined (see comment above) I have refactored `Audio*StreamOpenSLES` and added tests. New tests passing on API 18, 21, 23 and 28.

dturner:
@philburk  Any chance of a +1 here?

src/opensles/AudioInputStreamOpenSLES.cpp
philburk: Right. If a caller request a specific format then they should either get that format or it should fail.

PR #196, opened 2018-08-24
builder: add OBOE_ENABLE_AAUDIO preprocessor flag

philburk:
Can be used to disable AAudio API for testing of OpenSL ES.


dturner:
Out of interest why is this necessary when you can use `builder.setAudioAPI`?

philburk:
Because there are a whole bunch of unit tests that just use the default API, which on my phone is AAudio. By setting this to 0 I was able to run all the unit tests on OpenSL ES.

PR #198, opened 2018-08-25
opensles: fix PerformanceMode setting

### LINKED ISSUE #179: Open stream on OpenSL ES should query performance mode
After opening the stream, we should call GetConfiguration(SL_ANDROID_KEY_PERFORMANCE_MODE) to determine the actual final performance mode.

The problem can be seen by using Oboe Tester on Marlin.
Select "Test Output"
Select "OpenSL ES" API.
Select sample rate 44100
Open a Stream:

EXPECT
See "NONE" for performance mode.

ACTUAL
See "LOW_LATENCY", which is not possible for 44100 Hz.

Note in logcat
07-30 04:01:12.778  2993  2993 W AudioTrack: AUDIO_OUTPUT_FLAG_FAST denied by server; frameCount 0 -> 1772

### LINKED ISSUE #197: LowLatency was not getting set for OpenSL ES
The GetInterface() call for output returns SL_RESULT_FEATURE_UNSUPPORTED
so the PERFORMANCEMODE is never set.

The reason seems to be because SL_IID_ANDROIDCONFIGURATION was not listed in EngineOpenSLES.cpp
Code is:
      const SLInterfaceID ids[] = {SL_IID_BUFFERQUEUE};
Should be:
    const SLInterfaceID ids[] = {SL_IID_BUFFERQUEUE, SL_IID_ANDROIDCONFIGURATION};

philburk:
The LowLatency request was not getting set correctly because
GetInterface was failing.

Also the final PerformanceMode was not being queried after the stream was created.
So it was set to whatever the caller requested instead of the actual value.

Fixes #197
Fixes #179

### DIFF
--- a/tests/testStreamOpen.cpp
+++ b/tests/testStreamOpen.cpp
@@ -69,3 +69,23 @@ TEST_F(StreamOpen, ForOpenSLESDefaultChannelCountIsUsed){
     ASSERT_EQ(mStream->getChannelCount(), 1);
     closeStream();
 }
+
+TEST_F(StreamOpen, OutputForOpenSLESPerformanceModeShouldBeNone){
+    // We will not get a LowLatency stream if we request 16000 Hz.
+    mBuilder.setSampleRate(16000);
+    mBuilder.setPerformanceMode(PerformanceMode::LowLatency);
+    mBuilder.setDirection(Direction::Output);
+    openStream(AudioApi::OpenSLES);
+    ASSERT_EQ((int)mStream->getPerformanceMode(), (int)PerformanceMode::None);
+    closeStream();
+}
+
+TEST_F(StreamOpen, InputForOpenSLESPerformanceModeShouldBeNone){
+    // We will not get a LowLatency stream if we request 16000 Hz.
+    mBuilder.setSampleRate(16000);
+    mBuilder.setPerformanceMode(PerformanceMode::LowLatency);
+    mBuilder.setDirection(Direction::Input);
+    openStream(AudioApi::OpenSLES);
+    ASSERT_EQ((int)mStream->getPerformanceMode(), (int)PerformanceMode::None);
+    closeStream();
+}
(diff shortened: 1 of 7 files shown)


tests/testStreamOpen.cpp
```diff
@@ -69,3 +69,23 @@ TEST_F(StreamOpen, ForOpenSLESDefaultChannelCountIsUsed){
...
+
+TEST_F(StreamOpen, InputForOpenSLESPerformanceModeShouldBeNone){
+    // We will not get a LowLatency stream if we request 16000 Hz.
+    mBuilder.setSampleRate(16000);
+    mBuilder.setPerformanceMode(PerformanceMode::LowLatency);
+    mBuilder.setDirection(Direction::Input);
+    openStream(AudioApi::OpenSLES);
+    ASSERT_EQ((int)mStream->getPerformanceMode(), (int)PerformanceMode::None);
+    closeStream();
+}
```
dturner: Would it be possible to add a complimentary test which runs on N MR1 and above (25+) which attempts to create an OpenSLES stream using PerformanceMode=LowLatency and then tests that it actually gets it?

The difficulty here is that Oboe's default sample rate would need to specified - and is only obtainable through Java.

dturner:
See one comment about adding a test for N MR1 for low latency streams which might not be possible. Happy to merge as long as all unit tests are passing on API 24 (where LowLatency property cannot be set) and API 25 (where it can). Can use the emulator for both these.

dturner:
Oh one other thing, it's probably worth running tests on API 16 as well just to make sure that the creation of the `SL_IID_ANDROIDCONFIGURATION` interface doesn't fail on the oldest API level we support.

tests/testStreamOpen.cpp
philburk: Unfortunately, there is never any API guarantee that we will get a FAST track even under ideal circumstances, particularly on older devices. So testing for getting Low Latency is a big problem.

philburk:
I tested on API24 on a physical Nexus 6P. All PASSed.
Testing on API16 would be difficult. How do you root the emulator?

philburk:
Unfortunately the API never promises a FAST track. There could be any number of reasons you won't get one. For example you may be connected to Bluetooth. SO we cannot fail a test for not getting a FAST track.

PR #211, opened 2018-09-11
Adding docs to all public objects

dturner:


dturner:
Sorry for the size of this commit. It contains all the generated doxygen html as well. 

Only the oboe source files actually need reviewing. They can be seen from here onwards by clicking on "show diff" 

https://github.com/google/oboe/pull/211/files#diff-7c0f779b5fcc10367fd7629464efc5f9

include/oboe/AudioStream.h
```diff
@@ -216,28 +300,43 @@ class AudioStream : public AudioStreamBase {
...
+     * @return true if this stream is implemented using the AAudio API.
      */
     bool usesAAudio() const {
         return getAudioApi() == AudioApi::AAudio;
     }
 
     /**
-     * Do not use this for production. This is only for debugging.
-     * If you need to call this method then something is wrong.
+     * Only for debugging. Do use in production.
```
dturner: DO NOT use...

include/oboe/ResultWithValue.h
```diff
@@ -23,21 +23,61 @@
...
+ * if it did, a value which was obtained during the operation.
+ *
+ * For example, when reading from a stream the caller needs to know the result of the read operation
+ * and, if it was successful, how many frames were read. Note that ResultWithValue can be evaluated
+ * as a boolean so it's simple to check whether the result is OK.
+ *
+ * <code>
+ * ResultWithValue<int32_t> resultOfRead = myStream.read(&buffer, numFrames, timeoutNanoseconds);
+ *
+ * if (resultOfRead){
```
philburk: space before {

include/oboe/ResultWithValue.h
```diff
@@ -23,21 +23,61 @@
...
+ *
+ * It has been designed for cases where the caller needs to know whether an operation succeeded and,
+ * if it did, a value which was obtained during the operation.
+ *
+ * For example, when reading from a stream the caller needs to know the result of the read operation
+ * and, if it was successful, how many frames were read. Note that ResultWithValue can be evaluated
+ * as a boolean so it's simple to check whether the result is OK.
+ *
+ * <code>
+ * ResultWithValue<int32_t> resultOfRead = myStream.read(&buffer, numFrames, timeoutNanoseconds);
```
philburk: Are we recommending:   auto result

include/oboe/AudioStream.h
```diff
@@ -277,22 +389,46 @@ class AudioStream : public AudioStreamBase {
...
+     * Used to set the format of the underlying stream
+     */
     virtual void setNativeFormat(AudioFormat format) {
         mNativeFormat = format;
     }
 
-    // TODO: make private
-    // These do not change after open.
+    /**
+     * Do not set directly, use `setNativeFormat`
```
philburk: Do not document mNativeFormat because we plan to make it private.

include/oboe/AudioStream.h
```diff
@@ -192,6 +266,15 @@ class AudioStream : public AudioStreamBase {
         return ResultWithValue<double>(Result::ErrorUnimplemented);
     }
 
+    /**
+     * Get the timestamp that the frame at `framePosition` was presented to the
+     * audio hardware.
```
philburk: Should this be the time when the sound in air is produced?
Here is a definition:
https://developer.android.com/reference/android/media/AudioTimestamp

include/oboe/AudioStream.h
```diff
@@ -151,18 +199,44 @@ class AudioStream : public AudioStreamBase {
...
+     * and the sample format. For example, a 2 channel floating point stream will have
+     * 2 * 4 = 8 bytes per frame.
+     *
+     * @return number of bytes in each audio frame.
+     */
     int32_t getBytesPerFrame() const { return mChannelCount * getBytesPerSample(); }
 
+    /**
+     * Get the number of bytes per sample. This is calculated using the sample format. For example,
+     * stream using 16-bit integer samples will have 2 bytes per sample.
```
philburk: streams
OR
a stream

include/oboe/AudioStream.h
```diff
@@ -151,18 +199,44 @@ class AudioStream : public AudioStreamBase {
      */
     virtual int32_t getFramesPerBurst() = 0;
 
+    /**
+     * Indicates whether the audio stream is playing.
+     *
+     * THIS MAY BE REMOVED IN FUTURE. https://github.com/google/oboe/issues/208
```
philburk: We should not document this at all except for deprecation.

include/oboe/AudioStream.h
```diff
@@ -28,6 +28,12 @@
 
 namespace oboe {
 
+/**
+ * The default number of nanoseconds to wait for when performing state change operations on the
```
philburk: Do we want this?

include/oboe/ResultWithValue.h
dturner: Fixed.

include/oboe/ResultWithValue.h
dturner: It's up to the caller to decide whether `auto` is appropriate. In this case I'm explicitly using `ResultWithValue<int32_t>` since the reader may not know that `AudioStream::read` returns that type.

include/oboe/AudioStream.h
dturner: I based the description on: https://developer.android.com/ndk/reference/group/audio#aaudiostream_gettimestamp

so we probably need to update those docs since "presented" isn't clearly defined (I always thought it was the time an audio frame was presented to, or received from, the codec) 

Description updated in accordance with `AudioTimestamp` docs

include/oboe/AudioStream.h
dturner: Fixed

include/oboe/AudioStream.h
dturner: Fixed

include/oboe/AudioStream.h
dturner: I think it's good to have a sensible default value for timeouts. Did you want to remove it?

Issue #171, opened 2018-07-23
Incorporate tricks from Chrome into Oboe, clear buffers on stop


philburk:
Clear buffers in stop()
https://cs.chromium.org/chromium/src/media/audio/android/opensles_output.cc?l=43

dturner:
https://cs.chromium.org/chromium/src/media/audio/android/opensles_output.cc?l=182

philburk:
Nice!
> What is the "vivo" blacklist?
Moved to separate bug.
https://github.com/google/oboe/issues/201

PR #223, opened 2018-09-19
Clear queue buffers when the stream is stopped

dturner:
Referencing #171


philburk:
Nice. Thanks for implementing requestFlush().

PR #226, opened 2018-09-22
check usingFifo() for getFramesRead/Written()

### LINKED ISSUE #153: Audio*StreamOpenSLES should check usingFifo() in getFramesRead/Written
If the stream is in blocking read/write mode then getFramesRead() and getFramesWritten() should return the number of frames related to the Oboe FIFO, not the internal mixer.

if (usingFifo()) return AudioStreamBuffered::getFramesRead();
else ...

philburk:
For better frame counting in OpenSL ES.
When a FIFO was being used, getFramesRead() was returning
the server position. It should return the FIFO position.

Fixes #153

### DIFF
--- a/include/oboe/AudioStream.h
+++ b/include/oboe/AudioStream.h
@@ -356,26 +356,6 @@ public:
 protected:
 
     /**
-     * Increment the frames written to this stream
-     *
-     * @param frames number of frames to increment by
-     * @return total frames which have been written
-     */
-    virtual int64_t incrementFramesWritten(int32_t frames) {
-        return mFramesWritten += frames;
-    }
-
-    /**
-     * Increment the frames which have been read from this stream
-     *
-     * @param frames number of frames to increment by
-     * @return total frames which have been read
-     */
-    virtual int64_t incrementFramesRead(int32_t frames) {
-        return mFramesRead += frames;
-    }
-
-    /**
      * Wait for a transition from one state to another.
      * @return OK if the endingState was observed, or ErrorUnexpectedState
      *   if any state that was not the startingState or endingState was observed
@@ -431,6 +411,7 @@ protected:
 
 private:
     int                  mPreviousScheduler = -1;
+
 };
 
 } // namespace oboe
--- a/include/oboe/AudioStream.h
+++ b/include/oboe/AudioStream.h
@@ -66,7 +66,7 @@ public:
     /**
      * Close the stream and deallocate any resources from the open() call.
      */
-    virtual Result close() = 0;
+    virtual Result close();
 
     /**
      * Start the stream. This will block until the stream has been started, an error occurs
@@ -229,7 +229,7 @@ public:
      *
      * @return the number of frames written so far
      */
-    virtual int64_t getFramesWritten() { return mFramesWritten; }
+    virtual int64_t getFramesWritten();
 
     /**
      * The number of audio frames read from the stream.
@@ -237,7 +237,7 @@ public:
      *
      * @return the number of frames read so far
      */
-    virtual int64_t getFramesRead() { return mFramesRead; }
+    virtual int64_t getFramesRead();
 
     /**
      * Calculate the latency of a stream based on getTimestamp().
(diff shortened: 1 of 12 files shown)


src/aaudio/AudioStreamAAudio.cpp
```diff
@@ -221,14 +221,17 @@ Result AudioStreamAAudio::open() {
...
-Result AudioStreamAAudio::close()
-{
+Result AudioStreamAAudio::close() {
     // The main reason we have this mutex if to prevent a collision between a call
     // by the application to stop a stream at the same time that an onError callback
     // is being executed because of a disconnect. The close will delete the stream,
     // which could otherwise cause the requestStop() to crash.
     std::lock_guard<std::mutex> lock(mLock);
 
+    // Update frame counter variables to avoid retrograde motion.
```
dturner: You're updating something by calling `get` methods which feels weird. Could you use a setter instead or even `incrementFramesRead(0)` at a push?

src/opensles/AudioStreamBuffered.cpp
```diff
@@ -40,19 +42,20 @@ void AudioStreamBuffered::allocateFifo() {
...
         }
-        mFifoBuffer.reset(new FifoBuffer(getBytesPerFrame(), capacity));
+        // TODO consider using std::make_unique if we require c++14
+        mFifoBuffer.reset( new FifoBuffer(getBytesPerFrame(), capacity));
     }
 }
 
 int64_t AudioStreamBuffered::getFramesWritten() {
-    if (usingFIFO()) {
+    if (mFifoBuffer) {
```
dturner: Would prefer explicit check `mFifoBuffer != nullptr` to make it clear that `mFifoBuffer` isn't a bool.

dturner: Ah right, I see later on you changed `mFifoBuffer` to a `unique_ptr` which has a bool operator. As you were :)

src/opensles/AudioStreamOpenSLES.cpp
```diff
@@ -216,6 +216,10 @@ Result AudioStreamOpenSLES::close() {
     if (mState == StreamState::Closed){
         return Result::ErrorClosed;
     } else {
+        // Update frame counter variables so they can be read after close.
+        getFramesWritten();
```
dturner: Again would prefer not to update something using get methods.

src/opensles/AudioStreamBuffered.cpp
philburk: I'm trying to be more C++ish. ;-)

src/aaudio/AudioStreamAAudio.cpp
philburk: Right. That does feel wierd. I will add an update method.

include/oboe/AudioStream.h
```diff
@@ -395,6 +395,16 @@ class AudioStream : public AudioStreamBase {
 
     AudioFormat mNativeFormat = AudioFormat::Invalid;
 
+    /**
+     * Update mFramesWritten.
```
dturner: Does this need to be a public method? It feels like this should only ever be called by the stream class (or subclass) to update its own frame counter from the underlying stream, which isn't something a client should be concerned with. If you agree, suggest changing to `protected`.

include/oboe/AudioStream.h
```diff
@@ -395,6 +395,16 @@ class AudioStream : public AudioStreamBase {
...
+     */
+    virtual void updateFramesWritten() = 0;
+
+    /**
+     * Update mFramesRead.
+     */
+    virtual void updateFramesRead() = 0;
```
dturner: As above. Also might be worth moving the 2 member variables below into a protected section as well. The more we can reduce the API surface the less likely we will be to break it.

include/oboe/AudioStream.h
philburk: It is already "protected:".

include/oboe/AudioStream.h
philburk: It is already "protected:".

PR #234, opened 2018-10-03
fix readNow buffer pointer

philburk:
Was clearing beginning of buffer instead of end of buffer.
Also document read() and readNow().
This affects OpenSL ES blocking writes when un underflow occured.

Fixes: 233
Fixes: 154

### DIFF
--- a/src/fifo/FifoBuffer.h
+++ b/src/fifo/FifoBuffer.h
@@ -64,7 +64,7 @@ public:
      * @param framesToRead number of frames requested
      * @return number of frames actually read
      */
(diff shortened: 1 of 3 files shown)


src/fifo/FifoBuffer.h
```diff
@@ -50,6 +56,14 @@ class FifoBuffer {
 
     uint32_t getBufferCapacityInFrames() const;
 
+    /**
+     * Calls read(). If all of the frames cannot be read then the remainder of the buffer
+     * is set to zero and the underruncount is incremented.
+     *
+     * @param destination
```
philburk: buffer, not destination

Issue #32, opened 2018-01-01
Clarification on thread safety


svenoaks:
It seems a little confusing to me from the guide what is permissible as far as thread safety. 

> Note: When a stream uses a callback function, it's safe to read/write from the callback thread while also closing the stream from the thread in which it is running

Does this mean I can be writing data to the audioData void* in the callback while simultaneously calling `requestPause()`, `requestStop()`, `close()` in some other function on another thread?
`

philburk:
Thanks for asking. The Thread Safety section is pretty confusing. Some of it is inaccurate. Oboe does use some mutexes for non-time-critical functions like start()/stop().

There are some things you CAN do.

- You can call AudioStream::stop(), requestStop() , pause() or requestPause() from an application thread while a callback is running in a system thread.  That will stop the callback thread.

- You can call AudioStream::close() from an application thread while a callback is running in a system thread. It will stop the callback before closing or deleting any stream resources.

There are some things you CANNOT do.

- Never call AudioStream::stop(), requestStop() , pause(), requestPause() or close() from a callback method. Return oboe::DataCallbackResult::Stop if you want the callback to stop.

- Never call  AudioStream::read() or AudioStream::write() on a stream from that stream's own callback method. You should use the audioData array instead. You can read() or write() another stream from a callback as long as that stream does not have an active callback.  We recommend that if you have data flowing between multiple streams then only one of them should use a callback.

In general you should only modify the stream from one application thread. From the callback methods try to only use get*() methods.

Please ask more questions. When we have an explanation that makes sense then we will update the docs.

svenoaks:
Thanks, what about `getBufferSizeInFrames()` calling which could be concurrent with the `LatencyTuner->tune()` in the callback?

philburk:
You can safely call **get**BufferSizeInFrames() from as many threads as you want. You may be sampling a changing value. So only use the information for non-critical purposes, such as a fullness display.

Only call **set**BufferSizeInFrames() from one thread. So if latency tuning is active in the callback, do not call it from the application.

svenoaks:
Thanks. As far as I understand, `getBufferSizeInFrames()` is an ever increasing value, then should become stable when it's no longer possible to glitch the audio. I'm saving this value in SharedPreferences for use in next session so it won't have to retune, exposing user to audio glitches again.  Is there a recommended setting for bufferSizeInFrames to prevent glitching to start with, I'm finding it often shoots up to 4 or more times framesPerBurst.

[...]

PR #236, opened 2018-10-05
docs: add restrictions to AudioStreamCallback

philburk:
For example, do not stop() or write() a stream from the callback.

Some of these docs were borrowed from AAudio.

Issue #32


philburk:
Ping for review.

PR #274, opened 2018-11-16
Adding stricter compile error flags to make Oboe compatible with JUCE

dturner:
Currently Oboe doesn't work with the JUCE DemoRunner project because it uses [stricter compile flags](https://github.com/WeAreROLI/JUCE/blob/master/examples/DemoRunner/DemoRunner.jucer#L96) causing errors when building Oboe.

This change adds some stricter compile flags and fixes the code which subsequently breaks - mainly where we  were comparing `int32_t` to `uint32_t`.

### DIFF
--- a/include/oboe/AudioStreamBase.h
+++ b/include/oboe/AudioStreamBase.h
@@ -76,7 +76,7 @@ public:
      *
      * @return buffer size
      */
-    virtual int32_t getBufferSizeInFrames() { return mBufferSizeInFrames; };
+    virtual int32_t getBufferSizeInFrames() { return mBufferSizeInFrames; }
 
     /**
      * @return capacityInFrames or kUnspecified
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,7 +26,7 @@ add_library(oboe STATIC ${oboe_sources})
 
 # Specify directories which the compiler should look for headers
 target_include_directories(oboe
-                           PRIVATE src 
+                           PRIVATE src
                            PUBLIC include)
 
 # Compile Flags:
@@ -37,6 +37,8 @@ target_compile_options(oboe
     -std=c++11
     -Wall
     -Wextra-semi
+	-Wshadow
+	-Wshadow-field
     -Ofast
     "$<$<CONFIG:DEBUG>:-Werror>")
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,7 +33,12 @@ target_include_directories(oboe
 #     Enable -Werror when building debug config
 #     Enable -Ofast
 target_compile_options(oboe
-    PRIVATE -std=c++11 -Wall "$<$<CONFIG:DEBUG>:-Werror>" -Ofast)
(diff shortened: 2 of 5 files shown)


include/oboe/AudioStream.h
```diff
@@ -167,7 +167,7 @@ class AudioStream : public AudioStreamBase {
     * @return the resulting buffer size in frames (obtained using value()) or an error (obtained
     * using error())
     */
-    virtual ResultWithValue<int32_t> setBufferSizeInFrames(int32_t requestedFrames) {
+    virtual ResultWithValue<uint32_t> setBufferSizeInFrames(uint32_t requestedFrames) {
```
philburk: AAudio uses signed int32_t throughout for frame counts. This was deliberate. It is Google style and prevents lots of annoying bugs.

include/oboe/AudioStreamBase.h
```diff
@@ -76,12 +76,12 @@ class AudioStreamBase {
      *
      * @return buffer size
      */
-    virtual int32_t getBufferSizeInFrames() { return mBufferSizeInFrames; };
+    virtual uint32_t getBufferSizeInFrames() { return mBufferSizeInFrames; }
```
philburk: please leave these as int32_t

include/oboe/AudioStream.h
philburk: > I thought the reason why negative ints were used in AAudio was because they
> allowed values (positive ints) and error codes (negative ints) to be returned
> by a single type (int32_t).

No. The main reason is because there are lots of bugs that can occur when using unsigned ints and this avoids them. If you subtract two small numbers then you expect a small number. Not a huge wraparound result. Unsigned ints should only be used for non-numbers like bit masks, or if you really need the full unsigned range.

For read/write counters, for example, we want to subtract them and sometimes get a negative number if there is an underflow. Unsigned wrapping would be a mess.

> does it ever make sense for requestedFrames to be negative?

No. But it also does not make sense to have a buffer size of 1,073,741,823 frames. We don't have to use the full range of the data type.

dturner:
@philburk I've rolled back all the int->uint conversions and just made a few minor adjustments to tidy the code up - happy to merge?

src/fifo/FifoBuffer.cpp
```diff
@@ -112,7 +112,7 @@ int32_t FifoBuffer::read(void *buffer, int32_t numFrames) {
         memcpy(destination, source, numBytes);
     } else {
         // just read in one shot
-        numBytes = convertFramesToBytes(framesToRead);
+        size_t numBytes = convertFramesToBytes(framesToRead);
```
philburk: This conflicts with some bounds checking that was recently added.
https://github.com/google/oboe/blob/master/src/fifo/FifoBuffer.cpp#L120

dturner: I've rebased master into this branch which picks up the bounds checking changes, and resolves conflicts with master. Any chance of an approval?

Issue #168, opened 2018-07-20
Implement P usage and input presets for OpenSL ES


philburk:
Use SL_ANDROID_KEY_STREAM_TYPE based on USAGE and CONTENT_TYPE.
Need to figure out the optimal mapping.

Use SL_ANDROID_KEY_RECORDING_PRESET based on INPUT_PRESET.

philburk:
Dup of https://github.com/google/oboe/issues/78

philburk:
Reopening this because this bug is for OpenSL ES issue, which still needs work. The AAudio side is done.

dturner:
Steps to test: 

- Clone https://github.com/pbodilis/oboe
- Open `hello-oboe` project (in Android Studio 3.4 Canary 8)
- In `PlayAudioEngine::setupPlaybackStreamParameters` set the stream usage to each of the following `Media`, `VoiceCommunication`, `Alarm`, `Notification` and `Assistant`
- For each usage run `hello-oboe`, change audio API to OpenSL ES. 
- Open volume settings and verify that changing the volume for that stream type changes the volume of the app's audio output 

Note the mappings from usage to the volume sliders in the settings UI (in Android Pie) are as follows: 

- Media -> Media volume
- VoiceCommunication -> Call volume
- Alarm -> Alarm volume
- Notification -> Ring volume
- Assistant -> Ring volume

It'd be better if the above tests were run from the command line - @philburk any ideas how we could do this? Pretty sure it's not possible at present.

philburk:
Fixed in #283

PR #283, opened 2018-11-26
[code] use the value stored in the builder's Usage in AudioOutputStreamOpenSLES

pbodilis:
I guess this partly solves issue #168.

I'm not sure about the ContentType. Would that be for AudioInputStreamOpenSLES?

### DIFF
--- a/src/opensles/AudioOutputStreamOpenSLES.cpp
+++ b/src/opensles/AudioOutputStreamOpenSLES.cpp
@@ -52,7 +52,7 @@ static SLuint32 OpenSLES_convertOutputUsage(Usage oboeUsage) {
             openslStream = SL_ANDROID_STREAM_SYSTEM;
             break;
         case Usage::Game:
-            openslStream = SL_ANDROID_RECORDING_PRESET_VOICE_COMMUNICATION;
+            openslStream = SL_ANDROID_STREAM_MEDIA;
             break;
         case Usage::Assistant:
         default:
@@ -201,8 +201,6 @@ Result AudioOutputStreamOpenSLES::open() {
             goto error;
         }
     }
-
-
 
     result = (*mObjectInterface)->Realize(mObjectInterface, SL_BOOLEAN_FALSE);
     if (SL_RESULT_SUCCESS != result) {
--- a/src/opensles/AudioOutputStreamOpenSLES.cpp
+++ b/src/opensles/AudioOutputStreamOpenSLES.cpp
@@ -28,6 +28,40 @@
 
 using namespace oboe;
 
+static SLuint32 OpenSLES_convertOutputUsage(Usage oboeUsage) {
+    SLuint32 openslStream = SL_ANDROID_STREAM_MEDIA;
+    switch(oboeUsage) {
+        case Usage::Media:
+            openslStream = SL_ANDROID_STREAM_MEDIA;
+            break;
+        case Usage::VoiceCommunication:
+        case Usage::VoiceCommunicationSignalling:
+            openslStream = SL_ANDROID_STREAM_VOICE;
+            break;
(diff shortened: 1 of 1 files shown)


pbodilis:
I signed it!

src/opensles/AudioOutputStreamOpenSLES.cpp
```diff
@@ -28,6 +28,40 @@
...
+        case Usage::NotificationEvent:
+            openslStream = SL_ANDROID_STREAM_NOTIFICATION;
+            break;
+        case Usage::AssistanceAccessibility:
+        case Usage::AssistanceNavigationGuidance:
+        case Usage::AssistanceSonification:
+            openslStream = SL_ANDROID_STREAM_SYSTEM;
+            break;
+        case Usage::Game:
+            openslStream = SL_ANDROID_RECORDING_PRESET_VOICE_COMMUNICATION;
```
philburk: SL_ANDROID_RECORDING_PRESET_VOICE_COMMUNICATION is not a streamType.
No ideal match is available.
Maybe should be SL_ANDROID_STREAM_MEDIA.

src/opensles/AudioOutputStreamOpenSLES.cpp
```diff
@@ -157,8 +191,19 @@ Result AudioOutputStreamOpenSLES::open() {
...
+        result = (*configItf)->SetConfiguration(configItf,
+                                                SL_ANDROID_KEY_STREAM_TYPE,
+                                                &presetValue,
+                                                sizeof(SLuint32));
+        if (SL_RESULT_SUCCESS != result) {
+            goto error;
+        }
     }
 
+
```
philburk: extra space

src/opensles/AudioOutputStreamOpenSLES.cpp
pbodilis: Oh, right, my bad. Agreed on SL_ANDROID_STREAM_MEDIA.

pbodilis:
I don't know about ContentType... I felt Usage was more in line. If ContentType were to be used, how would you prioritize one above the other? Or would you combine both of them? So far, I'm ok with ignoring the parameter as it is right now.

dturner:
For me: run a test using dumpsys to verify that the stream types are set correctly

pbodilis:
> For me: run a test using dumpsys to verify that the stream types are set correctly

Which part of dumpsys shows this?
My modification enabled me to change the right volume stream, so I guess it works fine.

pbodilis:
I'm not sure my changes in my PR have been taken into account, am I missing something?

dturner:
Sorry, that previous comment was literally for me. I need to run a dumpsys to verify that your changes operate as intended (which I'm sure they do), then I'll merge.

dturner:
Steps to test: 

- Clone https://github.com/pbodilis/oboe
- Open `hello-oboe` project (in Android Studio 3.4 Canary 8)
- In `PlayAudioEngine::setupPlaybackStreamParameters` set the stream usage to each of the following `Media`, `VoiceCommunication`, `Alarm`, `Notification` and `Assistant`
- For each usage run `hello-oboe`, change audio API to OpenSL ES. 
- Open volume settings and verify that changing the volume for that stream type changes the volume of the app's audio output 

Note the mappings from usage to the volume sliders in the settings UI (in Android Pie) are as follows: 

- Media -> Media volume
- VoiceCommunication -> Call volume
- Alarm -> Alarm volume
- Notification -> Ring volume
- Assistant -> Ring volume

It'd be better if the above tests were run from the command line - @philburk any ideas how we could do this? Pretty sure it's not possible at present.

dturner:
@pbodilis Thanks very much for doing this 👍

PR #290, opened 2018-11-29
Add LoadStabilizer class

dturner:
Adding `LoadStabilizer` class which is a wrapper for `AudioStreamCallback`. It executes the client's callback method, then keeps performing CPU operations for the remaining 80% of the callback in order to keep the CPU frequency high.

### DIFF
--- a/src/common/StabilizedCallback.cpp
+++ b/src/common/StabilizedCallback.cpp
@@ -18,7 +18,7 @@
 #include "common/AudioClock.h"
 #include "common/Trace.h"
 
-constexpr int32_t kLoadGenerationStepSizeNanos = 1000;
+constexpr int32_t kLoadGenerationStepSizeNanos = 20000;
 constexpr float kPercentageOfCallbackToUse = 0.8;
 
 using namespace oboe;
@@ -91,7 +91,7 @@ void StabilizedCallback::generateLoad(int64_t durationNanos) {
     // the CPU for a fixed amount of time (specified by kLoadGenerationStepSizeNanos).
     // After each step the opsPerStep value is re-calculated based on the actual time taken to
     // execute those operations.
-    int opsPerStep = (int)(mOpsPerNano * kLoadGenerationStepSizeNanos);
+    auto opsPerStep = (int)(mOpsPerNano * kLoadGenerationStepSizeNanos);
     int64_t stepDurationNanos = 0;
     int64_t previousTimeNanos = 0;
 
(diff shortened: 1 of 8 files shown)


src/common/StabilizedCallback.cpp
```diff
@@ -0,0 +1,108 @@
...
+    int64_t previousTimeNanos = 0;
+
+    while (currentTimeNanos <= deadlineTimeNanos){
+
+        for (int i = 0; i < opsPerStep; i++) cpu_relax();
+
+        previousTimeNanos = currentTimeNanos;
+        currentTimeNanos = AudioClock::getNanoseconds();
+        stepDurationNanos = currentTimeNanos - previousTimeNanos;
+        mOpsPerNano = (int)(opsPerStep / stepDurationNanos);
```
philburk: You may want to low pass filter this. Also don't (int) before setting mOpsPerNano.

static const kFilterCoefficient = 0.1;
double measuredOpsPerNano = opsPerStep / stepDurationNanos;
mOpsPerNano = (kFilterCoefficient * measuredOpsPerNano) + ((1.0 -kFilterCoefficient) * mOpsPerNano);

src/common/StabilizedCallback.cpp
```diff
@@ -0,0 +1,108 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "oboe/StabilizedCallback.h"
+#include "common/AudioClock.h"
+#include "common/Trace.h"
+
+constexpr int32_t kLoadGenerationStepSizeNanos = 1000;
```
philburk: 1000 is 1 microsecond. That seems pretty short. Maybe try 20 micros.

src/common/StabilizedCallback.cpp
dturner: Nice! TIL: an exponential moving average can be used as a low pass filter. Tested and working great with the MegaDrone sample

src/common/StabilizedCallback.cpp
dturner: Done

dturner:
@philburk Thanks for your review. I've made those changes. Just one thing, in my code here: 

https://github.com/google/oboe/blob/stabilizedcallback/src/common/StabilizedCallback.cpp#L81

I'm using a field `mFrameCount` to keep track of the frames written/read, however, the `AudioStream` can provide this information using `getFrames[Written|Read]()` - would it be more efficient to check the stream direction and use the relevant `getFrames` method to obtain this value?

philburk:
I think using your own mFrameCount is fine. The getFramesRead/Written() has some version specific issues and is not well supported on OpenSL ES.

PR #306, opened 2018-12-07
Fix the memory leak in RhythmGame. Fixes #295

### LINKED ISSUE #295: RhythmGame leaks sound assets
SoundRecording::loadFromAssets opens an AAset but does not close it.

https://github.com/google/oboe/blob/master/samples/RhythmGame/src/main/cpp/audio/SoundRecording.cpp

The SoundRecording uses the buffer that seems to be owned by the Asset.
This may be OK for the RhythmGame because the assets are opened once and then used for the duration of the app. But an app that needs to load and unload assets might inherit a memory leak if they use the code without modification.

dturner:
As per discussion earlier, I've decoupled the source from the player. Feedback welcome.


samples/RhythmGame/src/main/cpp/audio/AAssetDataSource.cpp
```diff
@@ -0,0 +1,52 @@
...
+ * limitations under the License.
+ */
+
+
+#include <utils/logging.h>
+#include "AAssetDataSource.h"
+
+
+AAssetDataSource* AAssetDataSource::newFromAssetManager(AAssetManager *assetManager,
+														const char *filename,
```
philburk: Indented with tabs!

samples/RhythmGame/src/main/cpp/audio/AAssetDataSource.cpp
```diff
@@ -0,0 +1,52 @@
...
+														const int32_t channelCount) {
+
+	// Load the backing track
+	AAsset* asset = AAssetManager_open(assetManager, filename, AASSET_MODE_BUFFER);
```
philburk: This whole file seems to have tabs. Four spaces are preferred.

samples/RhythmGame/src/main/cpp/audio/AAssetDataSource.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+	if (asset == nullptr){
+		LOGE("Failed to open track, filename %s", filename);
+		return nullptr;
+	}
+
+	// Get the length of the track (we assume it is stereo 48kHz)
+	off_t trackLength = AAsset_getLength(asset);
```
philburk: I think this is in bytes. Might avoid some bugs by calling this trackSizeBytes.

samples/RhythmGame/src/main/cpp/audio/AAssetDataSource.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+	if (audioBuffer == nullptr){
+		LOGE("Could not get buffer for track");
+		return nullptr;
+	}
+
+	// There are 4 bytes per frame because
+	// each sample is 2 bytes and
+	// it's a stereo recording which has 2 samples per frame.
+	int32_t numFrames = static_cast<int32_t>(trackLength / 4);
```
philburk: Change 4 to sizeof(int16_t) * channelCount

samples/RhythmGame/src/main/cpp/audio/Player.cpp
```diff
@@ -0,0 +1,48 @@
...
+        for (int i = 0; i < numFrames; ++i) {
+            for (int j = 0; j < mChannelCount; ++j) {
+                targetData[(i*mChannelCount)+j] = data[(mReadFrameIndex*mChannelCount)+j];
+            }
+
+            // Increment and handle wraparound
+            if (++mReadFrameIndex >= totalFrames) mReadFrameIndex = 0;
+        }
+
+    } else {
```
philburk: Should silence any frames that were not rendered, either the whole or partial buffer.

samples/RhythmGame/src/main/cpp/audio/Player.h
```diff
@@ -27,29 +27,36 @@
...
-class SoundRecording : public RenderableAudio{
+class Player : public RenderableAudio{
 
 public:
-    SoundRecording(const int16_t *sourceData, int32_t numFrames)
-            : mData(sourceData)
-            , mTotalFrames(numFrames)
+    /**
+     * Construct a new Player from the given DataSource. Note that the Player does *not* take
+     * ownership of the source so it's up to the caller to delete the source after it's no longer
```
philburk: Document that Multiple players can share the same DataSource

samples/RhythmGame/src/main/cpp/audio/AAssetDataSource.cpp
dturner: That's weird. In both Android Studio and CLion I have my editor set to 4 spaces. Fixed now.

samples/RhythmGame/src/main/cpp/Game.cpp
```diff
@@ -76,6 +90,18 @@ void Game::start() {
     }
 }
 
+void Game::stop(){
+
+    if (mAudioStream != nullptr){
+        mAudioStream->close();
+    }
+
+    delete mClap;
```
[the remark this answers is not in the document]
dturner: Good tip. Refactored to use `std::unique_ptr`.

samples/RhythmGame/src/main/cpp/audio/AAssetDataSource.cpp
dturner: Fixed

samples/RhythmGame/src/main/cpp/audio/AAssetDataSource.cpp
dturner: Done

samples/RhythmGame/src/main/cpp/audio/Player.cpp
dturner: Done

samples/RhythmGame/src/main/cpp/audio/Player.h
dturner: Done

samples/RhythmGame/src/main/cpp/Game.cpp
```diff
@@ -94,10 +94,9 @@ void Game::stop(){
 
     if (mAudioStream != nullptr){
         mAudioStream->close();
+        mAudioStream = nullptr;
```
philburk: Add delete mAudioStream;

This doc probably needs updating!
https://github.com/google/oboe/blob/master/docs/FullGuide.md#closing-an-audio-stream

samples/RhythmGame/src/main/cpp/audio/Player.cpp
```diff
@@ -19,30 +19,41 @@
 
 void Player::renderAudio(int16_t *targetData, int32_t numFrames){
 
+    const int32_t channelCount = mSource.getChannelCount();
```
philburk: Do we guarantee somehow that the AAudio stream channel count matches the channel count of the source files?

dturner: We don't. However, this is documented here: https://github.com/google/oboe/tree/master/samples/RhythmGame#caveats-and-limitations

samples/RhythmGame/src/main/cpp/Game.cpp
dturner: There must be a better way than this. Maybe for Oboe v2.0 we should change `openStream` to return a smart pointer where `stream->close()` is called when the stream is destroyed.

dturner: Anyway - done :)

Issue #61, opened 2018-02-14
AudioStreamOpenSLES::processBufferCallback returns SL_RESULT_INTERNAL_ERROR


philburk:
It does this in order to stop the stream if the callback requests a stop, in other words if the return from the callback is != DataCallbackResult::Continue.

Don suggested we could use the SL player interface to set the play state to stopped. E.g.

https://github.com/googlesamples/android-ndk/blob/master/audio-echo/app/src/main/cpp/audio_player.cpp#L235

Additionally, we could set the stream state to stopped.

We need to test and verify that it is safe to stop an OpenSL ES stream from a callback.

philburk:
Fixed by #323

Issue #277, opened 2018-11-21
should oboe::DataCallbackResult::Stop also stop the stream?


SchmadenSchmuki:
Hello
I would like to know, what to expect when I call oboe::DataCallbackResult::Stop in the onAudioReady-method?

In my recordingcode I use it as returnvalue when I have no memory left to store additional data.
The result is that onAudioReady still gets called continuously (and returns oboe::DataCallbackResult::Stop) until the stream is stopped somewhere else.

I can replace oboe::DataCallbackResult::Stop with oboeStream->stop() but I was wondering whether oboe::DataCallbackResult::Stop does something important and should also be called?

philburk:
Thanks for reporting this. It should stop the callbacks from being called. 

I cannot reproduce this bug.
What OS version were you using?
What device?

Can you insert a few lines from the logcat around the time the Stop is returned?

Are you seeing any logs like this:
    AAudioServiceStreamBase: writeUpMessageQueue(): Queue full. Did client die? Shared

> I can replace oboe::DataCallbackResult::Stop with oboeStream->stop()

No. Do not call oboeStream->stop() from the callback. The stop() has to be called from another thread. If I can reproduce the bug then I may change Oboe so that it stops the stream from another thread if this problem occurs.

SchmadenSchmuki:
I use a Nexus 5X which runs Android 8.1

logs from starting the stream are as following:

    2018-11-21 19:05:59.434 2862-9874/? I/EventLogSendingHelper: Sending log events.
    2018-11-21 19:06:01.846 10602-10602/jogi.programme.android.soundtests D/AAudio: AAudioStream_requestStart(0x7147eac500) called --------------
    2018-11-21 19:06:01.851 10602-10602/jogi.programme.android.soundtests D/AAudio: AAudioStream_requestStart(0x7147eac500) returned 0 ---------
    2018-11-21 19:06:01.851 10602-10614/jogi.programme.android.soundtests D/AudioStreamLegacy: onAudioDeviceUpdate() deviceId 8
    2018-11-21 19:06:01.852 562-10657/? I/msm8974_platform: platform_check_capture_backend_cfg:txbecf: afe: Codec backend needs to be updated. new bit width: 0 new sample rate: 0 new channel: 2
    2018-11-21 19:06:01.853 562-10657/? D/sound_trigger_platform: platform_stdev_check_and_update_concurrency: concurrency active 0, tx 1, rx 0, concurrency session_allowed 0
[...]

PR #323, opened 2018-12-22
oboe: stop stream if callback requests it

philburk:
Stop in a separate thread for AAudio before Q.
Call stop directly for OpenSL ES.

Possible fix for #277

### DIFF
--- a/src/opensles/AudioInputStreamOpenSLES.cpp
+++ b/src/opensles/AudioInputStreamOpenSLES.cpp
@@ -206,18 +206,24 @@ Result AudioInputStreamOpenSLES::open() {
 }
 
 Result AudioInputStreamOpenSLES::close() {
-
+    LOGD("AudioInputStreamOpenSLES::%s()", __func__);
+    mLock.lock();
+    Result result = Result::OK;
     if (mState == StreamState::Closed){
-        return Result::ErrorClosed;
+        result = Result::ErrorClosed;
     } else {
+        mLock.unlock(); // avoid recursive lock
         requestStop();
+        mLock.lock();
+        // invalidate any interfaces
         mRecordInterface = NULL;
-        return AudioStreamOpenSLES::close();
+        result = AudioStreamOpenSLES::close();
     }
+    mLock.unlock(); // avoid recursive lock
+    return result;
 }
 
 Result AudioInputStreamOpenSLES::setRecordState(SLuint32 newState) {
-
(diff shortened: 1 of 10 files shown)


src/opensles/AudioInputStreamOpenSLES.cpp
```diff
@@ -234,41 +240,64 @@ Result AudioInputStreamOpenSLES::setRecordState(SLuint32 newState) {
...
+            mLock.unlock();
+            return Result::OK;
+        case StreamState::Closed:
+            mLock.unlock();
+            return Result::ErrorClosed;
+        default:
+            break;
+    }
 
+    setDataCallbackEnabled(true);
```
dturner: As previous comment. Not sure why the stream should always be using callbacks.

philburk: Enabled just means it is not disabled. It does not mean it will be used. But that is confusing. I will clarify or modify the code.

src/opensles/AudioStreamOpenSLES.cpp
```diff
@@ -259,7 +257,21 @@ SLresult AudioStreamOpenSLES::processBufferCallback(SLAndroidSimpleBufferQueueIt
...
+    } else if (result == DataCallbackResult::Stop) {
+        LOGD("Oboe callback returned Stop");
+        stopStream = true;
+    } else {
+        LOGW("Oboe callback returned unexpected value = %d", result);
+        stopStream = true;
+    }
+    if (stopStream) {
+        requestStop();
+        // launchStopThread();
```
dturner: Commented out code

src/opensles/AudioInputStreamOpenSLES.cpp
philburk: I cleaned up the enable and also refactored requestStart for OpenSL ES.

philburk:
Fixes #61

dturner:
Sorry, your question slipped through the net...

You should still be able to use `stream->write()` just don't set the callback or set it to nullptr. 

You can stop the stream at any time by calling `requestStop()`, and you can check the stream state (e.g. to see whether it's been disconnected) by calling `getState()`. 

Please open a new issue referencing this one if you're still experiencing problems.

dturner:
Additionally, you can check the result of `stream->write()`, if it's `Result::ErrorDisconnected` then you know the stream has been disconnected

PR #336, opened 2019-01-02
Refactor hello-oboe sample to use floats in signal path

dturner:
Using only floats in the signal path means not having to maintain 2 code paths - one for 16 bit ints and one for floats. 

I've only refactored the `hello-oboe` sample so far. I'm most interested in feedback on the way the conversion to 16-bit is handled (see comments).


samples/hello-oboe/src/main/cpp/PlayAudioEngine.cpp
```diff
@@ -80,10 +84,22 @@ void PlayAudioEngine::createPlaybackStream() {
 
         // TODO: Implement Oboe_convertStreamToText
         // PrintAudioStreamInfo(mPlayStream);
-        prepareOscillators();
+        if (mPlayStream->getFormat() == oboe::AudioFormat::I16){
```
dturner: After the stream is opened check whether it's 16-bit. 

If it is then create a buffer of floats which we'll render our data into. The buffer size is determined by `stream->getBufferCapacityInFrames` since this is the maximum number of frames we could feasibly be requested to render inside `onAudioReady`

samples/hello-oboe/src/main/cpp/PlayAudioEngine.cpp
```diff
@@ -178,27 +187,18 @@ PlayAudioEngine::onAudioReady(oboe::AudioStream *audioStream, void *audioData, i
     Trace::beginSection("numFrames %d, Underruns %d, buffer size %d",
                         numFrames, underrunCountResult.value(), bufferSize);
 
+    bool is16BitFormat = (audioStream->getFormat() == oboe::AudioFormat::I16);
```
dturner: If the stream is 16-bit then render into the buffer, then use `convertFloatToPCM16` on it

dturner:
My issue with this approach is that it feels messy to leave the conversion logic scattered about through `PlayAudioEngine`. I'm thinking maybe a shared "FloatToPCM16Converter" class which handles the allocation of the float buffer and the conversion.

samples/hello-oboe/src/main/cpp/PlayAudioEngine.cpp
```diff
@@ -80,10 +84,22 @@ void PlayAudioEngine::createPlaybackStream() {
 
         // TODO: Implement Oboe_convertStreamToText
         // PrintAudioStreamInfo(mPlayStream);
-        prepareOscillators();
+        if (mPlayStream->getFormat() == oboe::AudioFormat::I16){
+            // create a buffer of floats which we can render our audio data into
+            int conversionBufferSamples = mPlayStream->getBufferCapacityInFrames() * channelCount;
+            LOGD("Stream format is 16-bit integers, creating a temporary buffer of %d samples"
+                 " for float->int16 conversion", conversionBufferSamples);
+            mConversionBuffer = new float[conversionBufferSamples];
```
philburk: Could use unique_ptr<float[]>

samples/hello-oboe/src/main/cpp/PlayAudioEngine.cpp
```diff
@@ -178,27 +187,18 @@ PlayAudioEngine::onAudioReady(oboe::AudioStream *audioStream, void *audioData, i
...
-                mOscillators[i].render(static_cast<int16_t *>(audioData) + i, channelCount, numFrames);
-            }
-        } else {
-            memset(static_cast<uint8_t *>(audioData), 0,
-                   sizeof(int16_t) * channelCount * numFrames);
-        }
+    // If the stream is 16-bit render into a float buffer then convert that buffer to 16-bit ints
+    float *outputBuffer = (is16BitFormat) ? mConversionBuffer : static_cast<float *>(audioData);
+
+    mSoundGenerator->renderAudio(outputBuffer, numFrames);
```
philburk: What if numFrames is greater than the size of the mConversionBuffer?
I guess that won't happen in a callback.

samples/shared/Mixer.h
```diff
@@ -17,19 +17,19 @@
...
-class Mixer : public RenderableAudio<T> {
+class Mixer : public IRenderableAudio {
 
 public:
-    void renderAudio(T *audioData, int32_t numFrames) {
+    void renderAudio(float *audioData, int32_t numFrames) {
 
         // Zero out the incoming container array
-        memset(audioData, 0, sizeof(T) * numFrames);
+        memset(audioData, 0, sizeof(float) * numFrames);
```
philburk: Assumes a frame has only one channel.

samples/hello-oboe/src/main/cpp/PlayAudioEngine.cpp
dturner: Thanks, done.

samples/hello-oboe/src/main/cpp/PlayAudioEngine.cpp
dturner: That's why I used `stream->getBufferCapacityInFrames` to determine the size of `mConversionBuffer` - could `numFrames` ever be bigger than that?

samples/shared/Mixer.h
dturner: I've changed the class name to `MixerMono` and updated the docs

PR #348, opened 2019-01-07
Set default buffer size to two bursts for AAudio low latency streams

dturner:


tests/testStreamOpen.cpp
```diff
@@ -198,3 +198,14 @@ TEST_F(StreamOpen, PlaybackFormatI16ReturnsI16){
...
     closeStream();
 }
+
+TEST_F(StreamOpen, LowLatencyStreamHasBufferSizeOfTwoBursts){
+
+    if (mBuilder.isAAudioRecommended()){
+        mBuilder.setDirection(Direction::Output);
+        mBuilder.setPerformanceMode(PerformanceMode::LowLatency);
+        openStream();
+        ASSERT_EQ(mStream->getBufferSizeInFrames(), mStream->getFramesPerBurst() * 2);
```
philburk: setBufferSizeInFrames() does not work for OpenSL ES callbacks. But we set the buffer size to two bursts by default. So this may happen to work.

tests/testStreamOpen.cpp
```diff
@@ -198,3 +198,14 @@ TEST_F(StreamOpen, PlaybackFormatI16ReturnsI16){
     ASSERT_EQ(mStream->getFormat(), AudioFormat::I16);
     closeStream();
 }
+
+TEST_F(StreamOpen, LowLatencyStreamHasBufferSizeOfTwoBursts){
```
philburk: space before {

src/common/AudioStreamBuilder.cpp
```diff
@@ -88,6 +90,18 @@ Result AudioStreamBuilder::openStream(AudioStream **streamPP) {
     }
     Result result = streamP->open(); // TODO review API
     if (result == Result::OK) {
+
+        // Use a reasonable default buffer size for low latency streams.
+        if (streamP->getPerformanceMode() == PerformanceMode::LowLatency){
```
philburk: space before {

tests/testStreamOpen.cpp
dturner: The reason I haven't merged this is because I couldn't get the test to pass for OpenSL ES. Need to spend more time investigating.

PR #365, opened 2019-01-30
oboe: use shared_ptr to manage the AAudio stream pointer

philburk:
Use an owner class to contain the raw 'C' pointer.

### DIFF
--- a/src/aaudio/AudioStreamAAudio.cpp
+++ b/src/aaudio/AudioStreamAAudio.cpp
@@ -89,7 +89,11 @@ bool AudioStreamAAudio::isSupported() {
 }
 
 Result AudioStreamAAudio::open() {
+    std::lock_guard<std::mutex> lock(mLock); // prevent use before open finishes
+
     Result result = Result::OK;
+    AAudioStream *stream = nullptr;
+
 
     if (mAAudioStream != nullptr) {
         return Result::ErrorInvalidState;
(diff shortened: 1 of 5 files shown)


src/aaudio/AudioStreamAAudio.h
```diff
@@ -98,24 +98,53 @@ class AudioStreamAAudio : public AudioStream {
...
+
 private:
 
-    bool                 isMMapUsed();
+    /** Define a class that wraps a 'C' style AAudioStream pointer.
+     * We can then use a std:shared_ptr to manage the ownership of that pointer.
+     */
+    class   AAudioStreamOwner {
+    public:
+
```
mnaganov: I would suggest making this class non-copyable by deleting copy constructor and copy assignment.

src/aaudio/AudioStreamAAudio.h
```diff
@@ -98,24 +98,53 @@ class AudioStreamAAudio : public AudioStream {
...
+        aaudio_result_t getOpenResult() {
+            return mOpenResult;
+        }
+
+        AAudioStream *getStream() {
+            return mAAudioStream.load();
+        }
+
+    private:
+        aaudio_result_t mOpenResult = AAUDIO_OK;
```
mnaganov: I think it would be better not to have 'mOpenResult' here. Instead, provide a static factory method which returns a shared_ptr&lt;AAudioStreamOwner&gt;, and fills out the result as an out parameter. Make AAudioStreamOwner's constructor private.

src/aaudio/AudioStreamAAudio.h
```diff
@@ -98,24 +98,53 @@ class AudioStreamAAudio : public AudioStream {
...
-    void *getUnderlyingStream() const override {
-        return mAAudioStream.load();
-    }
+    bool usesMMap() override;
 
 protected:
     void updateFramesRead() override;
     void updateFramesWritten() override;
 
+    static AAudioLoader *mLibLoader;
```
mnaganov:
```suggestion
    static AAudioLoader *mLibLoader = nullptr;
```

src/aaudio/AudioStreamAAudio.cpp
```diff
@@ -162,39 +166,44 @@ Result AudioStreamAAudio::open() {
     mLibLoader->builder_setErrorCallback(aaudioBuilder, oboe_aaudio_error_callback_proc, this);
 
     // ============= OPEN THE STREAM ================
-    {
-        AAudioStream *stream = nullptr;
-        result = static_cast<Result>(mLibLoader->builder_openStream(aaudioBuilder, &stream));
-        mAAudioStream.store(stream);
-    }
+    mAAudioStream = std::make_shared<AAudioStreamOwner>(aaudioBuilder);
+    std::shared_ptr<AAudioStreamOwner> mine(mAAudioStream);
```
mnaganov: I don't think you have to create a second reference. The existence of mAAudioStream should prevent the stream from being deleted. Or am I missing something here?

src/aaudio/AudioStreamAAudio.cpp
```diff
@@ -345,9 +350,11 @@ Result AudioStreamAAudio::requestStop() {
 ResultWithValue<int32_t>   AudioStreamAAudio::write(const void *buffer,
                                      int32_t numFrames,
                                      int64_t timeoutNanoseconds) {
-    AAudioStream *stream = mAAudioStream.load();
-    if (stream != nullptr) {
-        int32_t result = mLibLoader->stream_write(mAAudioStream, buffer,
+    // Own the stream while in use.
+    std::shared_ptr<AAudioStreamOwner> mine(mAAudioStream);
```
mnaganov: Does it mean that an instance of AudioStreamAAudio can be destroyed while this method is running? That can lead to bad consequences since it can be destroyed before we even get to reading of 'mAAudioStream' here. I think, a safer solution is to manage AudioStreamAAudio by a shared pointer, and hold one from all the threads that need it.

philburk:
Thank you Kevin and Mikhail. Following our discussion. Kevin recommended using a std:shared_mutex for the lock. I will try that in an alternative CL.

philburk:
We decided to just document that apps should not close a stream while another thread is reading or writing.

PR #369, opened 2019-01-31
CI script: no need to install constraintLayout after ver 1.0.2

ggfan:


philburk:
Is this the fix for Travis?

PR #352, opened 2019-01-14
Second round of Chromium security review.

Dor1s:

PR #373, opened 2019-02-04
oboe: cleanup AAudio loader

philburk:
Use signatures and helper functions for loading
all AAudio functions.
Use RTLD_NOW in dlopen().

This partially addresses the Chrome security review in https://github.com/google/oboe/pull/352

### DIFF
--- a/src/aaudio/AAudioLoader.cpp
+++ b/src/aaudio/AAudioLoader.cpp
@@ -38,7 +38,7 @@ int AAudioLoader::open() {
     if (mLibHandle != nullptr) {
         return 0;
     }
-    mLibHandle = dlopen(LIB_AAUDIO_NAME, 0);
+    mLibHandle = dlopen(LIB_AAUDIO_NAME, RTLD_LAZY);
     if (mLibHandle == nullptr) {
         LOGI("AAudioLoader::open() could not find " LIB_AAUDIO_NAME);
         return -1; // TODO review return code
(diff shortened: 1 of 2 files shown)


Dor1s:
Hey Phil, thanks a lot for the changes! Is it fine if I take a look at this closer to the end of week?

src/aaudio/AAudioLoader.cpp
```diff
@@ -38,7 +38,7 @@ int AAudioLoader::open() {
     if (mLibHandle != nullptr) {
         return 0;
     }
-    mLibHandle = dlopen(LIB_AAUDIO_NAME, 0);
+    mLibHandle = dlopen(LIB_AAUDIO_NAME, RTLD_LAZY);
```
dturner: Investigate whether this could have a real time impact

Dor1s: Copying my reply from https://github.com/google/oboe/pull/352#discussion_r252294750:

> Sorry, I might've been mistaken, but I thought that `RTLD_LAZY` is `0`, so my point was to use explicit identified instead of a literal value.

Now I have checked and see that `RTLD_LOCAL` is for `0`, so I guess either of those is fine, depending on the performance.

philburk:
Good info about RTLD_LAZY here:
http://tldp.org/HOWTO/Program-Library-HOWTO/dl-libraries.html
"While you're debugging, you'll probably want to use RTLD_NOW; using RTLD_LAZY can create inscrutable errors if there are unresolved references. Using RTLD_NOW makes opening the library take slightly longer (but it speeds up lookups later); if this causes a user interface problem you can switch to RTLD_LAZY later."
Switching to RTLD_NOW.

PR #375, opened 2019-02-04
oboe: fix openSL ES read, write

### LINKED ISSUE #370: Cannot play audio with AudioStream::write without dataCallback
I cannot use AudioStream::write to play audio without data callback.

It seems to be related to this change: https://github.com/google/oboe/commit/b878e2f389007e66f21bec7a7aefbd916289d1d1
This change requests AudioStream to stop if callback is not set. After commenting the check, I am able to play audio because it falls back again to the default callback.

philburk:
Blocking IO was stopping immediately because of a recent change.
The callback was disabled but blocking IO on OpenSL ES
needs an internal callback to read/write the top level FIFO.

Fixes #370
See bug for repro steps.

### DIFF
--- a/src/opensles/AudioInputStreamOpenSLES.cpp
+++ b/src/opensles/AudioInputStreamOpenSLES.cpp
@@ -253,9 +253,10 @@ Result AudioInputStreamOpenSLES::requestStart() {
             break;
     }
 
-    if (mStreamCallback != nullptr) { // Was a callback requested?
-        setDataCallbackEnabled(true);
-    }
+    // We use a callback if the user requests one
+    // OR if we have an internal callback to fill the blocking IO buffer.
+    setDataCallbackEnabled(true);
+
     setState(StreamState::Starting);
     Result result = setRecordState(SL_RECORDSTATE_RECORDING);
     if (result == Result::OK) {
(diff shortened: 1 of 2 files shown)


src/opensles/AudioInputStreamOpenSLES.cpp
```diff
@@ -253,9 +253,10 @@ Result AudioInputStreamOpenSLES::requestStart() {
             break;
     }
 
-    if (mStreamCallback != nullptr) { // Was a callback requested?
-        setDataCallbackEnabled(true);
-    }
+    // We use a callback if the user requests one
+    // OR if we have an internal callback to fill the blocking IO buffer.
+    setDataCallbackEnabled(true);
```
vietanhto: I am not familiar with oboe code base however In my opinion, if we always set dataCallbackEnabled as true on starting stream then we better get rid of it. The state(mDataCallbackEnabled) is not going to change mid stream and will never be false. Therefore we should remove it to simplify.

philburk: It is not always set true. It is set conditionally by AAudio. This was added to workaround some old bugs in the framework. And it does change mid stream. So we still need it.

philburk:
This is a small change and it fixes a regression. So merging now.

PR #378, opened 2019-02-04
oboe_opensles: lock requestFlush()

philburk:
Add _l version to prevent recursive lock from requestStop().

### DIFF
--- a/src/opensles/AudioOutputStreamOpenSLES.cpp
+++ b/src/opensles/AudioOutputStreamOpenSLES.cpp
@@ -343,6 +343,11 @@ Result AudioOutputStreamOpenSLES::requestPause() {
  * Flush/clear the queue buffers
  */
 Result AudioOutputStreamOpenSLES::requestFlush() {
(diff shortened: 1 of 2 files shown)


src/opensles/AudioOutputStreamOpenSLES.cpp
```diff
@@ -343,6 +343,11 @@ Result AudioOutputStreamOpenSLES::requestPause() {
  * Flush/clear the queue buffers
  */
 Result AudioOutputStreamOpenSLES::requestFlush() {
+    std::lock_guard<std::mutex> lock(mLock);
+    return requestFlush_l();
+}
+
+Result AudioOutputStreamOpenSLES::requestFlush_l() {
```
dturner: Does the `_l` mean: this function should be called under a lock?

philburk: Yes, that is a convention in Android framework.

PR #395, opened 2019-03-04
Flowgraph cleanup

philburk:
Change Block to Buffer
Add comments and TODO
Use unique_ptr for float buffer.


dturner:
I could do with checking this code out more thoroughly but since this isn't a core part of the Oboe library and you're blocked on it I'm LGTMing now. I'd like to give it a proper test drive soon though and will no doubt have more feedback then.

PR #397, opened 2019-03-07
CMakeLists.txt: make it possible to build shared library

atsushieno:
In cmake manner, building shared library is easily doable by specifying
BUILD_SHARED_LIBS option (with `-D` command line argument),
but this explicit `STATIC` specification prevents that.

Since STATIC build is the default, it can be simply removed.


dturner:
Thanks for this PR!

PR #399, opened 2019-03-08
OboeTester - full duplex streams and echo test, refactor

philburk:
Refactor OboeTester and add an Echo test.
Support full duplex input+output callbacks.


philburk:
@dturner - please see the commits related to flowgraph

philburk:
This CL is getting too big. It is just a test and it has lots of new features. So I am just going to merge it.

PR #213, opened 2018-09-12
Version 1.0 changes

dturner:

PR #408, opened 2019-03-15
Remove isPlaying from AudioStream header

julioz:
Since https://github.com/google/oboe/pull/213/commits/6437f5aa224330fbdf77ecc161cc868be663a974, `isPlaying` is no longer part of Oboe's API.

Looks like we forgot to remove the function definition from the header file which made it a pure `virtual` function with no implementation that will error out when called.

### DIFF
--- a/include/oboe/AudioStream.h
+++ b/include/oboe/AudioStream.h
@@ -200,13 +200,6 @@ public:
     virtual int32_t getFramesPerBurst() = 0;
 
     /**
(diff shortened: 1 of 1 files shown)


dturner:
Thanks very much for this. I think the plan was to deprecate the method first, then remove it at a later date. I'll just confirm our deprecation policy with another dev before merging this in.

philburk:
This seems safe to remove. I think that if anyone actually called this method then their build would fail.

Issue #406, opened 2019-03-15
Crash when headset plugs in/out


pokey909:
The crash happens when the stream is in Paused state and a headset gets connected or disconnected.

Next time you call `play()` (the synchronous API) we get the crash.
And the reason seems to be that as soon as the stream starts again, it will call `onErrorInThread` on the stream since it is now in `DISCONNECTED` state and delegate to `onErrorAfterClose` functions which run in their own detached thread. 
We restart the stream in the callbacks but the stream will be closed. Concurrently, the `start()` function will run `waitForStateChange` which is not supposed to be called during or after `close()`. And here the crash happens since the underlying stream got deleted.

I can work around that problem by only using the `requestStart()` function since it doesn't call `waitForStateTransition `.

Are we not supposed to use the synchronous methods at all? I don't see a way to make that threadsafe unless we use locks to synchronize the detached thread from `oboe_aaudio_error_callback_proc` with the `waitForStateTransition ` method.

Stacktrace below:
```
2019-03-15 14:17:28.570 3794-3823/com.soundcloud.flipper I/FlipperNative: 2019-03-15 13:17:28.570181 (tid:4152329364) play - play() API invoked.
2019-03-15 14:17:28.570 3794-3823/com.soundcloud.flipper I/FlipperNative: 2019-03-15 13:17:28.570310 (tid:4152329364) post - Posting Play ...
2019-03-15 14:17:28.570 3794-3823/com.soundcloud.flipper D/FlipperNative: 2019-03-15 13:17:28.570533 (tid:3598301552) operator() - Processing Play event.
2019-03-15 14:17:28.570 3794-3823/com.soundcloud.flipper W/FlipperNative: 2019-03-15 13:17:28.570649 (tid:3598301552) playing_action - PLAY ACTION
2019-03-15 14:17:28.571 3794-3823/com.soundcloud.flipper I/FlipperNative: 2019-03-15 13:17:28.570739 (tid:3598301552) on_entry - Entering state: Playing
2019-03-15 14:17:28.571 3794-3823/com.soundcloud.flipper D/FlipperNative: 2019-03-15 13:17:28.571215 (tid:3598301552) start - OBOE START
2019-03-15 14:17:28.571 3794-3827/com.soundcloud.flipper D/AAudio: AAudioStream_requestStart(0xece7fd00) called --------------
[...]

PR #409, opened 2019-03-16
synchronize with close() to avoid crashes when headset plugs in/out

pokey909:
Fixes the issue i created recently https://github.com/google/oboe/issues/406
But also uses locks inside `waitForStateTransition`. 
I did not dig into the code very deeply, but I can't see a reason not to use locks here. `close()` already uses a lock for the very same reason.
Does that make sense?


pokey909:
I signed it!

dturner:
Thanks for this pull request. We have reviewed it and decided to fix in a different way, see #406.

PR #411, opened 2019-03-16
Fix missing CMake install targets

NewProggie:
This commit adds missing CMake install targets so that a consuming
(Android) project can integrate Oboe using ExternalProject_Add().

Additionally, some cleanup in the CMake config file is done. Namely:

 1. Add `project()`, which adds name, version information and some convenient variables
 1. Sort files in alphabetical order
 1. Distinguish between build and install include folder inclusion
 1. Set C++ standard and extensions via proper CMAKE properties
 1. Remove unnecessary quotes
 1. Refer to bin/include/lib folder structure using standard GNU install directories


ggfan:
Thank you: Great! LGTM

dturner:
Gerry to test resulting binaries.

NewProggie:
> Question for installation: The default install location on linux is /usr/local/ set in CMAKE_INSTALL_PREFIX, probably that is not what want to install. When building for Android, local host installation (usr/local ) is not used as this is for cross compilation, correct? Should CMAKE_INSTALL_PREFIX be set to somewhere else instead (${CMAKE_CURRENT_SOURCE_DIR/install})?

Currently (that is without the changes in this PR) there simply is no install target for Oboe. Hence, the
documentation states that one should simply add oboe as another sub-project (using `add_subdirectory()`), which clearly violates any rules for good software dependency management.

With the changes in this PR, oboe is instructed to put its build artifacts in a dedicated lib/bin/lib64 folder, relative to the install prefix path provided by the user when the install target is executed. For instance: `set(CMAKE_INSTALL_PREFIX ${CMAKE_CURRENT_BINARY_DIR}/installed)`.

Please note that my changes in this PR are non-intrusive as it doesn't interfere with the current documented way of adding oboe to an Android project. More on this in the next section.

> "CMake install" need to be called to actually execute install target; but there is no support from gradle to pass this command (cmake install) when calling cmake, how to use this install target?

The preferred way to add external dependencies using CMake is `ExternalProject_Add()` [1]. For instance, adding oboe would look as follows:

```cmake
ExternalProject_Add(oboe
  URL ${oboe_path}
  URL_MD5 ${oboe_md5}
  INSTALL_DIR ${YourProject_INSTALL_PREFIX}/oboe
  UPDATE_COMMAND ""
  CMAKE_ARGS -Wno-dev ${YourProject_DEFAULT_ARGS}
    -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR>
  BUILD_BYPRODUCTS <INSTALL_DIR>/lib/${oboe_name})
```
where `${YourProject_DEFAULT_ARGS}` is populated with project or Android specific settings, which are as well listed here: https://developer.android.com/ndk/guides/cmake.html.
I'd be happy to provide a fully working oboe example using this approach as well, if needed. 

[1] https://cmake.org/cmake/help/latest/module/ExternalProject.html

ggfan:
@NewProggie just ping.  I wonder whether we could separate the change into 2 PRs
-  remove install() part from this CL
- add install on [this project](https://github.com/googlesamples/android-ndk/tree/master/hello-libs/gen-libs/src/main/cpp): that one has more reaches and was exactly targeted for this. 

For the later, we would need to solve how "install" target could be invoked via gradle plugin ( I am not clear how could be done with current plugin ).

May you kindly think about this approach?

NewProggie:
> For the later, we would need to solve how "install" target could be invoked via gradle plugin ( I am not clear how could be done with current plugin ).
> 
> May you kindly think about this approach?

You see, there is really no need for Gradle to do just about anything. Furthermore, we shouldn't enforce the user of this library to install anything.

This PR just happens to add an optional install target to the CMake build configuration that Oboe is using. Currently, it is not possible to install this library, because a dedicated install target is missing from the build configuration. Thus, the build system configuration of this library is broken in my point of view.

Whether a user wants to install this library or not should not be decided here, I think. However, *if* a user wants to install this library, we should provide a logical install target in order for the user to be able to do so. This is the point of this PR. In any other case, nothing actually changes.

CMakeLists.txt
```diff
@@ -1,48 +1,65 @@
...
 
 add_library(oboe STATIC ${oboe_sources})
 
 # Specify directories which the compiler should look for headers
 target_include_directories(oboe
-                           PRIVATE src
-                           PUBLIC include)
+  PRIVATE src
+  PUBLIC
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
```
dturner: Nice, thanks.

CMakeLists.txt
```diff
@@ -1,48 +1,65 @@
...
+    $<INSTALL_INTERFACE:include>)
```
dturner: https://cmake.org/cmake/help/v3.0/command/target_include_directories.html

dturner: This needs thinking through fully about the install path

dturner:
@NewProggie Thanks very much for this PR. We've attempted to get this working, however, we only have a rudimentary knowledge of CMake. A working example would be really helpful if you had the time to put one together.

dturner:
Waiting on working example. This might help: https://stackoverflow.com/questions/15175318/cmake-how-to-build-external-projects-and-include-their-targets

NewProggie:
Right, sorry. Forgot about that one.
I've just pushed a working example using CMake's `ExternalProject_Add()` to properly add oboe as a thirdparty library to an (existing) Android project. The most interesting part is probably: samples/hello-cmake-oboe/cmake/External_Oboe.cmake.

samples/hello-cmake-oboe/cmake/External_Oboe.cmake
```diff
@@ -0,0 +1,55 @@
...
+set(oboe_version "1.1.1")
+set(oboe_url "https://github.com/google/oboe/archive/${oboe_version}.zip")
+set(oboe_md5 "85b7cefe6046f17f8c7cf62762cbb5c7")
+
+set(oboe_name ${CMAKE_STATIC_LIBRARY_PREFIX}oboe${CMAKE_STATIC_LIBRARY_SUFFIX})
+
+ExternalProject_Add(oboe
+  GIT_REPOSITORY https://github.com/NewProggie/oboe
+  GIT_TAG fix_cmake_install_targets
+  # URL ${oboe_url}
```
NewProggie: Note that, in order to make this work, you need to change GIT_REPOSITORY/GIT_TAG with URL and URL_MD5, as the current version on master is missing said install targets

dturner:
This is really great. Thanks. I have just cloned your repo and tested the hello-cmake-oboe project and it's working as it should on my machine.

I now need to go through and figure out which bits we can merge and what documentation needs to be updated. 

Do you have any opinions on distributing binaries rather than the current source distribution?

NewProggie:
> I now need to go through and figure out which bits we can merge and what documentation needs to be updated.

I'd probably update the Getting Started Guide to include just the External_Oboe.cmake part (besides any other native thirdparty library a user might have) instead of manually cloning the repository and putting it alongside the user's own sources
 
> Do you have any opinions on distributing binaries rather than the current source distribution?

There is no real rule of thumb, unfortunately. In the case of oboe, I'd argue that it is acceptably small enough to be distributed in source. In any other case, you'd need to put binary artifacts for each supported architecture on Android obivously.

philburk:
> Do you have any opinions on distributing binaries rather than the current source distribution?

One thing I like about the user getting Oboe source by default  is that they can find bugs in Oboe more easily. If this External_Oboe tricks makes it easier to integrate the source then that would be great!

NewProggie:
Waiting

dturner:
Fair comment ;-) 

I've spent a fair few hours looking at this today. My plan was to update the documentation to include instructions on how to include Oboe in your project using `ExternalProject_Add`. Unfortunately I ran into a few issues: 

1) When using this approach the indexer fails meaning autocomplete for oboe classes doesn't work.  
2) much slower builds - the build ended up cloning the oboe repo from github _each time_ I ran a build which increased build time dramatically
3) added CMake complexity. This approach requires ~30 extra lines of code in your project's `CMakeLists.txt`, some of them cryptic and fragile (particularly the `CopyLibToJniFolder` step). Your example `External_Oboe.cmake` has upward dependencies on its parent project (e.g. `${HelloCMakeOboe_INSTALL_PREFIX}/oboe/lib/${oboe_name}` and adds a significant amount of extra cognitive load to the end user. We cannot assume our users are experts in CMake - the config for a new project should be able to be copy/pasted whilst remaining robust and broadly understandable - at the moment there's just too much "magic". 

At the moment I'm struggling to see the benefit of using `ExternalProject_Add` - please convince me otherwise. 

That said, I'm happy to update our `CMakeLists.txt` to support this approach. I've created a branch for this here. https://github.com/google/oboe/tree/new_cmakelists

NewProggie:
Alright, so it's been a while, but all I was asking for was to add proper install targets to Oboe in the first place :-)

Using CMake's `ExternalProject_Add` is a simple, basic approach to deal with dependencies in a C++ project. However, I agree that this approach may not be appropriate for upstream projects such as Oboe. On the other hand, upstream projects should make it as easy as possible (or even enable in this case) to be consumed in a somewhat natural CMake'ey way.

That being said, I do use Oboe besides a few other libraries using the external project mechanism in an Android project, so I'd like to address the issues you were mentioning:

1. I did experience the same problem with the  indexer. However, that went away when invalidating caches / restart Android Studio

2. That shouldn't really be the case either, so I'm guessing something else is different in your build config. I remember that older versions of Android Studio had that issue (3.1 maybe?), but that got fixed in later versions.

3. Again, I agree that this approach is not worthwhile for Oboe. Also, the build logic overhead may look daunting. However, that typically vanishes as more external dependencies get added to the project as the additional build code is less compared to a single dependency as shown here.

Your branch looks fine. I have a few minor suggestions though.

We can safely remove the quotes here. CMake is _stringly_ typed, hence the quotes are unnecessary:
```cmake
# Enable logging for debug builds
target_compile_definitions(oboe PUBLIC $<$<CONFIG:DEBUG>:OBOE_ENABLE_LOGGING=1>)
```

We need to tell _where_ CMake should install executables and static/shared libraries. Using `GNUInstallDirs` is not necessary, but chooses conventional install folders
```cmake
# TODO: Explain why are these necessary for the `ExternalProject_Add` approach
include(GNUInstallDirs)
install(TARGETS oboe
  RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
  LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
  ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
```

Alternatively, we could also write:
```cmake
install(TARGETS oboe
  RUNTIME DESTINATION bin
  LIBRARY DESTINATION lib
  ARCHIVE DESTINATION lib)
```

dturner:
Thanks for bearing with me on this and for the explanations. The delay has largely been due to my lack of experience with CMake - I'm getting there _slowly_. 

New pull request created here: https://github.com/google/oboe/pull/588

We'll need you to sign the CLA to accept your code.

dturner:
Fixed in https://github.com/google/oboe/pull/588. Thanks for helping to improve Oboe, your changes, feedback and patience are much appreciated 👍

PR #426, opened 2019-03-28
Lock waitForStateChange()

philburk:
Prevent race conditions with close() from another thread.

### DIFF
--- a/src/aaudio/AudioStreamAAudio.cpp
+++ b/src/aaudio/AudioStreamAAudio.cpp
@@ -20,6 +20,7 @@
 
 #include "aaudio/AAudioLoader.h"
 #include "aaudio/AudioStreamAAudio.h"
+#include "common/AudioClock.h"
 #include "common/OboeDebug.h"
 #include "oboe/Utilities.h"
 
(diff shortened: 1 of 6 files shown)


philburk:
This broke some tests including

04-04 11:42:19.186 15149 15174 D com.google.oboe.tests.unittestrunner.MainActivity: [ RUN      ] StreamClosedReturnValues.WaitForStateChangeReturnsOK
04-04 11:42:19.195 15149 15174 D com.google.oboe.tests.unittestrunner.MainActivity: /Volumes/android/github/oboe/tests/testStreamClosedMethods.cpp:313: Failure
04-04 11:42:19.195 15149 15174 D com.google.oboe.tests.unittestrunner.MainActivity: Expected equality of these values:

because the behavior changed when a timeout=0 was passed.
AAudio will return ERROR_TIMEOUT if the state is not changed from the passed in state.
Oboe used to return OK but with this patch it returned ErrorTime, like AAudio.

    Result r = mStream->waitForStateChange(StreamState::Open, &next, 0);

Should we change the Oboe API or keep the old behavior?

philburk:
Don and I discussed this and decided to change Oboe to match AAudio.

PR #438, opened 2019-04-04
Revert "Lock waitForStateChange()"

philburk:
Reverts google/oboe#426


philburk:
We decided that we do want the new behavior so I am reverting the revert.
I will update the tests as needed.

PR #453, opened 2019-04-11
Add <atomic> and <mutex> to AudioStream includes

### LINKED ISSUE #392: Missing atomic and mutex includes on AudioStream.h
When compiling with NDK r17 and Qt 5.10 I was confronted with missing includes for std::atomic and std::mutex. Adding the following to linux fixes the problem:
```
#include <atomic>
#include <mutex>
```

dturner:
`AudioStream.h` uses both `std::atomic` and `std::mutex`. It should have these headers included. Fixes #392

### DIFF
--- a/include/oboe/AudioStream.h
+++ b/include/oboe/AudioStream.h
@@ -19,6 +19,8 @@
 
 #include <cstdint>
 #include <ctime>
+#include <atomic>
+#include <mutex>
 #include "oboe/Definitions.h"
 #include "oboe/ResultWithValue.h"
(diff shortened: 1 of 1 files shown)


include/oboe/AudioStream.h
```diff
@@ -19,6 +19,8 @@
 
 #include <cstdint>
 #include <ctime>
+#include <atomic>
```
philburk: Nit: Move <atomic> above <cstdint>. Alphabetical order for grouped includes is ANdroid style.

PR #455, opened 2019-04-11
Fix waitForStateChange and related unit tests

philburk:


src/aaudio/AudioStreamAAudio.cpp
```diff
@@ -28,6 +28,12 @@
 #include <sys/system_properties.h>
 #endif
 
+#ifndef OBOE_FIX_FORCE_STARTING_TO_STARTED
```
dturner: OBOE_WA_<issueno>

PR #457, opened 2019-04-12
tests: sleep between close and open

### LINKED ISSUE #425: Flaky test: StreamStates.InputStreamLeftRunningShouldNotInterfereWithNextOpen
This test often fails when running on the emulator. Let's discuss during next meeting.

philburk:
This is to prevent a race condition on the emulator from
causing some tests to fail.

Fixes #425

### DIFF
--- a/tests/testStreamStates.cpp
+++ b/tests/testStreamStates.cpp
@@ -21,7 +21,7 @@ using namespace oboe;
 
 // Sleep between close and open to avoid a race condition inside Android Audio.
 // On a Pixel 2 emulator on a fast Linux host, the minimum value is around 16 msec.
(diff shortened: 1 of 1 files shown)


tests/testStreamStates.cpp
```diff
@@ -19,6 +19,10 @@
 
 using namespace oboe;
 
+// Sleep between close and open to avoid a race condition inside Android Audio.
+// On a Pixel 2 emulator on a fast Linux host, the minimum value is around 16 msec.
+#define OBOE_CLOSE_OPEN_SLEEP_MSEC  100
```
dturner: change to `constexpr`

PR #463, opened 2019-04-17
Add version string

dturner:


src/common/Version.cpp
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "oboe/Version.h"
+
+namespace oboe {
+	char kVersionText[] = "OboeVersion" OBOE_VERSION_TEXT;
```
philburk: Someone might come along later and optimize this.
Maybe add a comment that says it is coded this way so that the version string can be read from the app binary.

src/common/Version.cpp
```diff
@@ -0,0 +1,28 @@
...
+ * limitations under the License.
+ */
+#include "oboe/Version.h"
+
+namespace oboe {
+
+	// This variable enables the version information to be read from the resulting binary e.g.
+	// by running `objdump -s --section=.data <binary>`
+	// Please do not optimize or change in any way.
+	char kVersionText[] = "OboeVersion" OBOE_VERSION_TEXT;
```
mnaganov: I would try making this "const char" as it's being used in the function. In my prototype I had to make the array to be non-const because it wasn't used anywhere thus triggering a compiler warning.

src/common/Version.cpp
```diff
@@ -0,0 +1,28 @@
...
+
+	const char * getVersionText(){
+		return kVersionText;
```
mnaganov: 4 spaces?

dturner: good spot, thanks

src/common/Version.cpp
dturner: I tried that and string should appear in the `.rodata` section, however the output from `objdump -s --section=.rodata <binary>` was so verbose that I couldn't isolate the `OboeVersion` string visually. Don't suppose you have some magic grep/regexp incantation which will check for that string in the objdump output?

PR #487, opened 2019-05-09
486 - added state description to Stream Dump

atneya-nair:


apps/OboeTester/app/src/main/java/com/google/sample/oboe/manualtest/AudioStreamBase.java
```diff
@@ -81,6 +81,18 @@ String dump(int framesPerBurst) {
...
             return buffer.toString();
         }
+        /**
+         *Converts ints from JNI to human-readable stream state
+         */
+        private String parseState(int stateId) {
+            final String[] STATE_ARRAY = {"Uninit", "Unknown", "Open", "Starting", "Started",
+                    "Pausing", "Paused", "Flushing", "Flushed",
+                    "Stopping", "Stopped", "Closing", "Closed", "Disconnected"};
+            if (stateId < 0 || stateId > STATE_ARRAY.length) {
```
philburk: Should be ">="

apps/OboeTester/app/src/main/java/com/google/sample/oboe/manualtest/AudioStreamBase.java
```diff
@@ -81,6 +81,18 @@ String dump(int framesPerBurst) {
 
             return buffer.toString();
         }
+        /**
+         *Converts ints from JNI to human-readable stream state
```
philburk: Add space between "* C".
Change "JNI" to "Oboe index". JNI is just the interface between Java and Native.

apps/OboeTester/app/src/main/java/com/google/sample/oboe/manualtest/AudioStreamBase.java
```diff
@@ -81,6 +81,18 @@ String dump(int framesPerBurst) {
...
+         */
+        private String parseState(int stateId) {
+            final String[] STATE_ARRAY = {"Uninit", "Unknown", "Open", "Starting", "Started",
+                    "Pausing", "Paused", "Flushing", "Flushed",
+                    "Stopping", "Stopped", "Closing", "Closed", "Disconnected"};
```
philburk: I was originally thinking of shorter names. But since you removed the "state = " then these full names should fit. If we have to we can rearrange  some other text to make fit.

apps/OboeTester/app/src/main/java/com/google/sample/oboe/manualtest/AudioStreamBase.java
```diff
@@ -81,6 +81,18 @@ String dump(int framesPerBurst) {
 
             return buffer.toString();
         }
+        /**
+         *Converts ints from JNI to human-readable stream state
+         */
+        private String parseState(int stateId) {
```
philburk: We normally call these toString functions. How about "convertStateToString().

atneya-nair:
I updated with the modifications you suggested.

PR #497, opened 2019-05-13
oboe: move flowgraph from OboeTester to inside Oboe

philburk:
This will be used internally for doing data conversion within a stream.
This flowgraph API is experimental and is likely to change.

This also includes a bump up to C++14


philburk:
Fixes 494

PR #520, opened 2019-05-22
Code samples set stream default values. Fixes #513

### LINKED ISSUE #513: Samples should follow best practice for obtaining optimal latency
Check all samples to ensure they follow our guidance here: https://github.com/google/oboe/blob/master/docs/GettingStarted.md#obtaining-optimal-latency

dturner:


philburk:
I think you need to say "Fixes #513" in the comment, not the title for GitHub to parse it.

docs/GettingStarted.md
```diff
@@ -150,23 +150,17 @@ Here's a code sample showing how the default values for built-in devices can be
...
-	    native_setDefaultSampleRate(defaultSampleRate);
-	    native_setDefaultFramesPerBurst(defaultFramesPerBurst);
+	    native_setDefaultStreamValues(defaultSampleRate, defaultFramesPerBurst);
 	}
 
 *jni-bridge.cpp*
 
 	JNIEXPORT void JNICALL
-	Java_com_google_sample_oboe_hellooboe_MainActivity_native_1setDefaultSampleRate(JNIEnv *env,
+	Java_com_google_sample_oboe_hellooboe_MainActivity_native_1setDefaultStreamValues(JNIEnv *env,
```
philburk: Why is there a '1' in  "native_1set"?

samples/LiveEffect/src/main/java/com/google/sample/oboe/liveEffect/LiveEffectEngine.java
```diff
@@ -39,4 +36,17 @@
     static native void setRecordingDeviceId(int deviceId);
     static native void setPlaybackDeviceId(int deviceId);
     static native void delete();
-}
+    static native void native_setDefaultStreamValues(int defaultSampleRate, int defaultFramesPerBurst);
+
+    static void setDefaultStreamValues(Context context) {
```
philburk: Does this need @TargetApi(16) ?

docs/GettingStarted.md
dturner: It's a weird quirk of defining method names in JNI. If your method name in Java is `native_something` then your JNI declaration must have a "1" after the "_" so `native_1something`. Prefixing the method names with `native_` is good practice because you can easily identify native methods in your Java code.

samples/LiveEffect/src/main/java/com/google/sample/oboe/liveEffect/LiveEffectEngine.java
dturner: Since Oboe's minimum API level is 16 I don't think we need this.

PR #522, opened 2019-05-23
Refactor Live Effect to use Full Duplex Stream

atneya:
Refactoring Live Effect to use Full Duplex stream. Having a separate effect engine and duplex stream is a little redundant. But, the duplex stream class does not take ownership of stream creating and closure, so a second class is still necessary.

### DIFF
--- a/samples/LiveEffect/src/main/cpp/FullDuplexPass.h
+++ b/samples/LiveEffect/src/main/cpp/FullDuplexPass.h
@@ -20,11 +20,14 @@
 
 class FullDuplexPass : public FullDuplexStream {
 public:
-    const int BYTES_PER_FRAME = 4;
     virtual oboe::DataCallbackResult
     onBothStreamsReady(const void *inputData, int numInputFrames, void *outputData,
                        int numOutputFrames) {
-        memcpy(outputData, inputData, numInputFrames * BYTES_PER_FRAME);
+        size_t bytesPerFrame = this->getOutputStream()->getBytesPerFrame();
+        size_t bytesToWrite = numInputFrames * bytesPerFrame;
+        size_t bytesToZero = (numOutputFrames - numInputFrames) * bytesPerFrame > 0 ? : 0;
+        memcpy(outputData, inputData, bytesToWrite);
+        memset((u_char*) outputData + bytesToWrite, 0, bytesToZero);
         return oboe::DataCallbackResult::Continue;
     }
 };
--- a/samples/LiveEffect/src/main/cpp/FullDuplexPass.h
+++ b/samples/LiveEffect/src/main/cpp/FullDuplexPass.h
@@ -20,14 +20,11 @@
 
 class FullDuplexPass : public FullDuplexStream {
 public:
+    const int BYTES_PER_FRAME = 4;
     virtual oboe::DataCallbackResult
     onBothStreamsReady(const void *inputData, int numInputFrames, void *outputData,
                        int numOutputFrames) {
-        int16_t* intIn = (int16_t*) inputData;
-        int16_t* intOut = (int16_t*) outputData;
-        for(int i = 0; i < numOutputFrames; i++){
-            intOut[i] = intIn[i];
-        }
+        memcpy(outputData, inputData, numInputFrames * BYTES_PER_FRAME);
         return oboe::DataCallbackResult::Continue;
     }
 };
--- a/samples/LiveEffect/src/main/cpp/FullDuplexPass.h
+++ b/samples/LiveEffect/src/main/cpp/FullDuplexPass.h
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2018 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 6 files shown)


samples/LiveEffect/src/main/cpp/FullDuplexStream.h
```diff
@@ -0,0 +1,106 @@
...
+    void setMNumInputBurstsCushion(int32_t mNumInputBurstsCushion);
+
+private:
+
+    // TODO add getters and setters
+    static constexpr int32_t kNumCallbacksToDrain   = 20;
+    static constexpr int32_t kNumCallbacksToDiscard = 30;
+
+    // let input fill back up, usually 0 or 1
+    int32_t mNumInputBurstsCushion = 0;
```
philburk: Please experiment with setting to 1. That may reduce glitches.

samples/LiveEffect/src/main/cpp/FullDuplexStream.cpp
```diff
@@ -0,0 +1,117 @@
...
+    }
+    oboe::Result result = getInputStream()->requestStart();
+    if (result != oboe::Result::OK) {
+        return result;
+    }
+    return getOutputStream()->requestStart();
+}
+
+oboe::Result FullDuplexStream::stop() {
+    getOutputStream()->requestStop(); // TODO result?
```
philburk: Return any non-zero result.

samples/LiveEffect/src/main/cpp/LiveEffectEngine.cpp
```diff
@@ -61,112 +52,26 @@ bool LiveEffectEngine::setAudioApi(oboe::AudioApi api) {
 void LiveEffectEngine::setEffectOn(bool isOn) {
     if (isOn != mIsEffectOn) {
         mIsEffectOn = isOn;
-
         if (isOn) {
-            openAllStreams();
+            bindStreams();
```
philburk: Please use "open" instead of "bind" to be consistent with Oboe and AAudio APIs.

samples/LiveEffect/src/main/cpp/LiveEffectEngine.cpp
```diff
@@ -61,112 +52,26 @@ bool LiveEffectEngine::setAudioApi(oboe::AudioApi api) {
...
-            closeAllStreams();
+            mFullDuplexPass.stop();
+            closeStream(mRecordingStream);
+            closeStream(mPlayStream);
         }
     }
 }
-
-void LiveEffectEngine::openAllStreams() {
-    // Note: The order of stream creation is important. We create the playback
```
philburk: Please preserve this comment.

samples/LiveEffect/src/main/cpp/LiveEffectEngine.cpp
```diff
@@ -61,112 +52,26 @@ bool LiveEffectEngine::setAudioApi(oboe::AudioApi api) {
...
-             mRecordingStream, mPlayStream);
-        closeAllStreams();
-    }
-}
-
-/**
- * Stops and closes the playback and recording streams.
- */
-void LiveEffectEngine::closeAllStreams() {
-    /**
```
philburk: Please preserve this comment.

samples/LiveEffect/src/main/cpp/LiveEffectEngine.cpp
```diff
@@ -61,112 +52,26 @@ bool LiveEffectEngine::setAudioApi(oboe::AudioApi api) {
...
-
-        warnIfNotLowLatency(mPlayStream);
-
-    } else {
-        LOGE("Failed to create playback stream. Error: %s",
-             oboe::convertToText(result));
-    }
+void LiveEffectEngine::bindStreams() {
+    oboe::AudioStreamBuilder inBuilder, outBuilder;
+    setupRecordingStreamParameters(&inBuilder);
```
philburk: setupRecording probably needs to come after opening the output stream so it can match some properties.

samples/LiveEffect/src/main/cpp/FullDuplexPass.h
```diff
@@ -0,0 +1,31 @@
...
+#define SAMPLES_FULLDUPLEXPASS_H
+#include "FullDuplexStream.h"
+
+class FullDuplexPass : public FullDuplexStream {
+public:
+    const int BYTES_PER_FRAME = 4;
+    virtual oboe::DataCallbackResult
+    onBothStreamsReady(const void *inputData, int numInputFrames, void *outputData,
+                       int numOutputFrames) {
+        memcpy(outputData, inputData, numInputFrames * BYTES_PER_FRAME);
```
philburk: Please make this work for different channelCounts and frameCounts.

Zero out any part of outputData that is not set by the memcpy.

samples/LiveEffect/src/main/cpp/FullDuplexPass.h
```diff
@@ -20,11 +20,14 @@
...
 class FullDuplexPass : public FullDuplexStream {
 public:
-    const int BYTES_PER_FRAME = 4;
     virtual oboe::DataCallbackResult
     onBothStreamsReady(const void *inputData, int numInputFrames, void *outputData,
                        int numOutputFrames) {
-        memcpy(outputData, inputData, numInputFrames * BYTES_PER_FRAME);
+        size_t bytesPerFrame = this->getOutputStream()->getBytesPerFrame();
+        size_t bytesToWrite = numInputFrames * bytesPerFrame;
+        size_t bytesToZero = (numOutputFrames - numInputFrames) * bytesPerFrame > 0 ? : 0;
```
philburk: An empty case in the ternary operator! Interesting.

 Would bytesToZero be set to 
  ((numOutputFrames - numInputFrames) * bytesPerFrame)
OR
  ((numOutputFrames - numInputFrames) * bytesPerFrame > 0)
I suspect the latter.

An old fashioned if{} might be simpler.

samples/LiveEffect/src/main/cpp/FullDuplexPass.h
atneya: To confirm, is this issue resolved?

PR #523, opened 2019-05-23
Reinit. Audio Engine for MegaDrone on restart.

### LINKED ISSUE #521: [samples/MegaDrone] Crashes when onResume is called
MegaDrone sample crashes with the following error when onResume is called.
Test device: Samsung Galaxy S5 API 23 (MMB29M)

Steps to repro

- Run MegaDrone app
- Press power button (causing `onPause`) to be called (screen goes off)
- Press power button again (screen back on) and unlock

Expected result

- App is running as normal and can produce sound

Actual result

- App crashes with the following logcat output:

    05-22 18:46:55.797 19943-19943/com.example.oboe.megadrone A/libc: Fatal signal 11 (SIGSEGV), code 1, fault addr 0xcc6d8510 in tid 19943 (.oboe.megadrone)

Probably related to #511

atneya:
Fixes #521

### DIFF
--- a/samples/MegaDrone/src/main/cpp/Synth.h
+++ b/samples/MegaDrone/src/main/cpp/Synth.h
@@ -22,7 +22,6 @@
 #include "shared/Oscillator.h"
 #include "shared/Mixer.h"
 #include "shared/MonoToStereo.h"
-#include "../../../../../src/common/OboeDebug.h"
 
 constexpr int kNumOscillators = 100;
 constexpr float kOscBaseFrequency = 116.0;
@@ -33,19 +32,15 @@ constexpr float kOscAmplitude = 0.009;
 class Synth : public IRenderableAudio {
 public:
 
-    Synth(int32_t sampleRate, int32_t channelCount) {
+    Synth(int32_t sampleRate, int32_t channelCount):
+    mOutputStage(channelCount == oboe::ChannelCount::Stereo ?
+    static_cast<IRenderableAudio*>(&mConverter) : static_cast<IRenderableAudio*>(&mMixer)) {
 
         for (int i = 0; i < kNumOscillators; ++i) {
             mOscs[i].setSampleRate(sampleRate);
             mOscs[i].setFrequency(kOscBaseFrequency+(static_cast<float>(i)/kOscDivisor));
             mOscs[i].setAmplitude(kOscAmplitude);
             mMixer.addTrack(&mOscs[i]);
-        }
-
-        if (channelCount == oboe::ChannelCount::Stereo){
-            mOutputStage = &mConverter;
-        } else {
-            mOutputStage = &mMixer;
         }
     }
 
@@ -66,7 +61,7 @@ private:
     std::array<Oscillator, kNumOscillators> mOscs;
     Mixer mMixer;
     MonoToStereo mConverter = MonoToStereo(&mMixer);
-    IRenderableAudio *mOutputStage; // This will point to either the mixer or converter
+    IRenderableAudio * const mOutputStage; // This will point to either the mixer or converter
 };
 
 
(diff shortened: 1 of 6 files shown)


samples/MegaDrone/src/main/cpp/native-lib.cpp
```diff
@@ -41,23 +41,23 @@ extern "C"
...
 }
 
 extern "C"
 JNIEXPORT void JNICALL
 Java_com_example_oboe_megadrone_MainActivity_stopEngine(JNIEnv *env, jobject instance) {
 
-    engine.stop();
-
+    engine->stop();
+    //delete engine;
```
philburk: Where does this get deleted? Could we use a unique_ptr<>?

atneya-nair: This is what I noted in the issue, I was unsure where the engine was being deleted, but when I attempted to delete again, I hit mem access errors.

atneya:
Changed username, replying to recheck CLA

samples/MegaDrone/src/main/cpp/Synth.h
```diff
@@ -60,13 +59,14 @@ class Synth : public IRenderableAudio {
...
+    virtual ~Synth() {
+    }
 private:
-
     // Rendering objects
     std::array<Oscillator, kNumOscillators> mOscs;
     Mixer mMixer;
-    std::shared_ptr<IRenderableAudio> mOutputStage;
+    MonoToStereo mConverter = MonoToStereo(&mMixer);
+    IRenderableAudio *mOutputStage; // This will point to either the mixer or converter
```
philburk: Can this be declared const? The pointer will not change after the constructor.

IRenderableAudio * const mOutputStage; // const pointer to mutable object

samples/RhythmGame/src/main/cpp/Game.h
```diff
@@ -59,8 +59,8 @@ class Game : public AudioStreamCallback {
 private:
     AAssetManager& mAssetManager;
     AudioStream *mAudioStream { nullptr };
-    std::shared_ptr<Player> mClap;
-    std::shared_ptr<Player> mBackingTrack;
+    Player *mClap;
```
philburk: unique_ptr<> is another option

samples/shared/Mixer.h
```diff
@@ -45,15 +45,15 @@ class Mixer : public IRenderableAudio {
         }
     }
```
philburk: Add comment that ownership must be managed by the caller. Do not delete the render while in use.

samples/MegaDrone/src/main/cpp/Synth.h
```diff
@@ -22,6 +22,7 @@
 #include "shared/Oscillator.h"
 #include "shared/Mixer.h"
 #include "shared/MonoToStereo.h"
+#include "../../../../../src/common/OboeDebug.h"
```
philburk: Still needed?

atneya:
Using a const ptr is a little bit messy because of the need to initialize prior to the constructor. If it is clearer, we can use a non-const pointer and have it instantiated as before.

samples/MegaDrone/src/main/cpp/Synth.h
philburk: > Using a const ptr is a little bit messy because of the need to initialize prior to the constructor.
> If it is clearer, we can use a non-const pointer and have it instantiated as before.

I see what you mean. This is messy. Feel free to go back to the original.

atneya: Added in latest commit

dturner:
Many thanks for fixing this.

PR #530, opened 2019-05-31
Add sample rate and format conversion for optimal latency

philburk:
Add sample rate, format and channel count conversion to Oboe.

This allows various an app to request specific formats, sample rates or channel configurations and still get a low latency stream.


dturner:
From today's meeting. We should test: 

- On a variety of real devices (e.g. open a stream with unspecified sample rate, then open a stream with a different sample rate to the one we got, then check that resampling works)
- Test the resampler on low end device / legacy stream to ensure it's performant - no glitches

Diljeet:
Hello everyone, i would like to contribute a simple script for resampler
I am not an expert audio dsp programmer, so you must test it before using
I have a synthesizer and a dj app in playstore and i use this resampler algorithm in both my apps.
Normally sinc resampling is heavy duty task, but this algo is by far the quickest for me and it produces identical quality results.

Code: 4 point resampler using pre-computed coefficients
let "samples" be array of 4 samples
outSample be the resampled sample calculated from the 4 samples
coeffs be array of 4 coefficients

in every resampling, we calculate a point between 2 know samples and a percent is needed i.e. percent of the sample location between the two know samples.

so calculations are like this
```
float coeffs[4]; // coefficients
float samplePercent; // percent between the two know samples 0 to 1.0
float samples[4] = {0.5f, 1.0f, 0.25f, -0.5f}; // temporary samples array for example
float outSample;
void calculateCoeffs (float percent) {
    float xpow1 = percent;
    float xpow2 = xpow1 * xpow1;
    float xpow3 = xpow2 * xpow1;

    coeffs[0] = -0.5f * xpow3 + xpow2 - 0.5f * xpow1;
    coeffs[1] = 1.5f * xpow3 - 2.5f * xpow2 + 1.0f;
    coeffs[2] =-1.5f * xpow3 + 2.0f * xpow2 + 0.5f * xpow1;
    coeffs[3] = 0.5f * xpow3 - 0.5f * xpow2;
}
void main() { // main execution for temporary example
  float percent = 0.248f; // calculated during resampling in all algorithms
  calculateCoeffs(percent);
  outSample = samples[0] * coeffs[0] + samples[1] * coeffs[1] + samples[2] * coeffs[2] + samples[3] * coeffs[3];
// print the out sample
// use this algo with a full array, i resampled only a single sample
// i hope this is simple enough
}
```
Please try it once, i just want to help make oboe better.
Please notify me about the results.

include/oboe/AudioStreamBase.h
```diff
@@ -125,6 +125,27 @@ class AudioStreamBase {
      */
     SessionId getSessionId() const { return mSessionId; }
 
+    /**
+     * @return true if Oboe can convert channel counts to achieve optimal results.
+     */
+    bool isChannelConversionAllowed()const {
```
dturner: nit: space before `const`

philburk:
@Diljeet - thanks for the suggestion. That is code for a cubic spline interpolation.
https://www.maximalsound.com/mastering/interpolation%20methods.pdf

We were thinking of adding a polynomial interpolation as an option when upsampling. Unfortunately it does not provide the low pass filtering needed when downsampling.

Diljeet:
> 
> 
> @Diljeet - thanks for the suggestion. That is code for a cubic spline interpolation.
> https://www.maximalsound.com/mastering/interpolation%20methods.pdf
> 
> We were thinking of adding a polynomial interpolation as an option when upsampling. Unfortunately it does not provide the low pass filtering needed when downsampling.

Thank you sir, you are awesome (you are right, it is cubic interpolation).
I never thought that low pass filtering was that much important, yes i have read that downsampling is more tricky and needs low-pass filtering (but i don't know how to do that), once it will be implemented in oboe then i will study the code and will learn how to do it. Thanks for all the help, oboe and you guys are the best.

dturner:
Copying, and adding to, the TODOs in `AudioProcessorBase.h`: 

- Move the classes in `AudioProcessorBase.h` into separate files, this also avoids the need to forward declare `AudioPort` and `AudioFloatInputPort`
- Remove "Audio" prefix from class names. 
- After doing the above rename classes to make them more succinct. Some suggestions: `FloatBufferPort` -> `BufferPort`, `FloatOutputPort` -> `OutputPort`, `FloatInputPort` -> `InputPort`, `Filter`->`Processor`
- Review use of raw pointers for `connect()`. Maybe use smart pointers but need to avoid run-time deallocation in audio thread.

philburk:
I tested the resampler doing 44100 to 48000 Hz on An Asus Zen Go Phone running 8.1.
Even on the "Best" quality, it only took 2-3% of the CPU.

PR #531, opened 2019-05-31
Add support to easily use unique pointers for AudioStream

atneya:
Added a ManagedStream typedef which is an alias for a unique ptr of AudioStream with a custom destructor that closes the stream.

This allows for easy creation of a ManagedStream object, that can be modified with reset semantics, safely closing and deleting the old stream object.

Additionally, the ManagedStream is automatic, and when the enclosing class goes out of scope, it is safely deleted, preventing memory leaks and the necessity of overriding the enclosing class' destructor to close the stream.

### DIFF
--- a/include/oboe/AudioStream.h
+++ b/include/oboe/AudioStream.h
@@ -456,7 +456,10 @@ private:
     std::atomic<bool>    mDataCallbackEnabled{};
 
 };
-
+/**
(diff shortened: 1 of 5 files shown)


atneya:
Deleting this PR/branch in favor of issue-529_new

PR #532, opened 2019-06-03
[samples/MegaDrone] Move ownership of AudioEngine into Java

dturner:
Avoids having a raw pointer which requires manual deletion.

### DIFF
--- a/samples/MegaDrone/src/main/cpp/native-lib.cpp
+++ b/samples/MegaDrone/src/main/cpp/native-lib.cpp
@@ -20,7 +20,7 @@
 
 #include "AudioEngine.h"
 
-AudioEngine *engine;
+std::unique_ptr<AudioEngine> engine;
 
 std::vector<int> convertJavaArrayToVector(JNIEnv *env, jintArray intArray){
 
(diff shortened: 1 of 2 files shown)


atneya:
Since the unique pointer is not a member of a class, and it is a global, it will never go out of scope and won't be deleted.

dturner:
Good point. In my code the old `engine` object will be deleted when a new `engine` is created inside `startEngine` on this line: 

    engine = std::make_unique<AudioEngine>();

Question is: does it matter if there's always one `engine` object which doesn't get deleted and its memory is (presumably) cleaned up only when the app is destroyed?

atneya:
It generally is bad practice to use a smart pointer with a scoped lifetime outside of a specific scope. 

That said, you are correct that there will only be one AudioEngine at a time since we are using make unique, but the app holds resources it will delete onResume when it has gone out of focus.

dturner:
Best option is to pass pointer back to java - more research required, what is best practice?

dturner:
So I did some research into this. Here's what I found: 

Talk: [How to Manage Native C++ Memory in Android (Google I/O '17)](https://youtu.be/7_caITSjk1k)
 
In this talk Hans Boehm [mentions using an explicit close](https://youtu.be/7_caITSjk1k?t=846) when the C++ object is no longer required, with the caveat that knowing when to delete the C++ object is as difficult as knowing when our owning Java object will be garbage collected. 

In our case, we know that our MainActivity will exist for the lifetime of our app and that when `onStop` is called we should delete our audio engine. This is what I've implemented. 

If we had more a complicated object lifecycle we could use `PhantomReference` and reference queues to ensure our C++ objects were cleaned up.  

Another relevant talk: [The Ghost in the Virtual Machine: A Reference to References](https://www.youtube.com/watch?v=KTC0g14ImPc)

atneya:
Looks good, I think using close would be best practice for AudioStreams, since we don't create very many of them, and it follows traditional stream semantics.

PR #533, opened 2019-06-03
Adding ManagedStream type

atneya:
Adding easy support for uniqueptrs of AudioStreams via a ManagedStream type. This type has a custom deleter which closes the AudioStream to safely delete it. ManagedStream can be treated as a generic object, with AudioStreamBuilder safely repopulating a ManagedStream with the openManagedStream method. (This assumes close() is thread safe/safe to repeat).

Also refactoring Hello-Oboe to use this type. (This assumes onErrorAfterClose is safe).

### DIFF
--- a/include/oboe/AudioStream.h
+++ b/include/oboe/AudioStream.h
@@ -461,7 +461,7 @@ private:
  * This struct is a stateless functor which closes a audiostream prior to its deletion.
  * This means it can be used to safely delete a smart pointer referring to an open stream.
  */
-    struct streamDeleterFunctor {
+    struct StreamDeleterFunctor {
         void operator()(AudioStream  *audioStream) {
             if (audioStream) {
                 audioStream->close();
--- a/include/oboe/AudioStream.h
+++ b/include/oboe/AudioStream.h
@@ -457,6 +457,18 @@ private:
 
 };
 
+/**
+ * This struct is a stateless functor which closes a audiostream prior to its deletion.
+ * This means it can be used to safely delete a smart pointer referring to an open stream.
+ */
+    struct streamDeleterFunctor {
+        void operator()(AudioStream  *audioStream) {
+            if (audioStream) {
+                audioStream->close();
+            }
+            delete audioStream;
+        }
+    };
 } // namespace oboe
 
 #endif /* OBOE_STREAM_H_ */
--- a/include/oboe/AudioStreamBuilder.h
+++ b/include/oboe/AudioStreamBuilder.h
@@ -22,6 +22,9 @@
 
 namespace oboe {
 
+    // This depends on AudioStream, so we use forward declaration, it will close and delete the stream
+    struct streamDeleterFunctor;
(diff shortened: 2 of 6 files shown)


atneya:
We should add documentation that a reference to a ManagedStream should not be passed by reference between multiple threads (as typical of unique ptrs), as this makes deleting it unsafe.

include/oboe/AudioStream.h
```diff
@@ -457,6 +457,18 @@ class AudioStream : public AudioStreamBase {
 
 };
 
+/**
+ * This struct is a stateless functor which closes a audiostream prior to its deletion.
+ * This means it can be used to safely delete a smart pointer referring to an open stream.
+ */
+    struct streamDeleterFunctor {
```
philburk: Capitalize Stream

include/oboe/AudioStreamBuilder.h
```diff
@@ -324,6 +327,18 @@ class AudioStreamBuilder : public AudioStreamBase {
      */
     Result openStream(AudioStream **stream);
 
+    /**
+     * Create and open a ManagedStream object based on the current builder state.
+     *
+     * The caller must create a unique ptr, and pass by reference so it can be
+     * modified to point to an opened stream. The caller owns the unique ptr,
+     * and it will be automatically closed and cleaned up upon going out of scope.
```
philburk: "cleaned up upon" is hard to read.
Maybe "deleted when".

samples/hello-oboe/src/main/cpp/PlayAudioEngine.cpp
```diff
@@ -223,7 +206,6 @@ void PlayAudioEngine::onErrorAfterClose(oboe::AudioStream *oboeStream, oboe::Res
 
 void PlayAudioEngine::restartStream(oboe::AudioStreamBuilder *builder) {
     LOGI("Restarting stream");
-    closeOutputStream();
     createPlaybackStream(builder);
```
philburk: We still want to delete the old stream before we create the new stream because the new stream may need exclusive resources held by the old stream.

src/common/AudioStream.cpp
```diff
@@ -167,4 +167,11 @@ void AudioStream::launchStopThread() {
     t.detach();
 }
 
+Result AudioStreamBuilder::openManagedStream(oboe::ManagedStream &stream) {
```
philburk: Move to AudioStreamBuilder.cpp

atneya:
Also important to note, this relies on `close()` and `~AudioStream()` and methods which override it to not throw Exceptions and be generally safe.

samples/hello-oboe/src/main/cpp/PlayAudioEngine.cpp
atneya: This is taken care of in the `AudioStreamBuilder::openManagedStream` method.

atneya:
Addressed prev comments.

PR #534, opened 2019-06-03
oboe: prevent onError callback from causing an assert

### LINKED ISSUE #519: crash report on abort in AudioStreamAAudio::onErrorInThread
I have noticed a crash report on our App from google play store, that reports an abort in 
AudioStreamAAudio::onErrorInThread. ( Huawei HUAWEI Y9 2019 (HWJKM-H), Android 8.1)
due to the `assert(stream == mAAudioStream.load());`
I suppose this is due to some Thread timing or not locked issue ( stream already closed before..) ?
but isn't there a none crashing way to deal with this in the mean time ?

philburk:
There was a race that could be caused by multiple onErrorCallbacks.
There should only be one callback.
We now check to make sure that only the first callback is handled.

Fixes #519

### DIFF
--- a/include/oboe/AudioStream.h
+++ b/include/oboe/AudioStream.h
@@ -381,7 +381,9 @@ protected:
      * These were bugs in some versions of Android that caused multiple error callbacks.
      * Internal bug b/63087953
      *
-     * @return true if the error callback was already called, otherwise false
+     * Calling this sets an atomic<bool> true and returns the previous value.
+     *
+     * @return false on first call, true on subsequent calls
      */
     bool wasErrorCallbackCalled() {
         return mErrorCallbackCalled.exchange(true);
--- a/include/oboe/AudioStream.h
+++ b/include/oboe/AudioStream.h
@@ -374,6 +374,17 @@ public:
      */
     void launchStopThread();
 
+    /**
+     * This can be called to prevent handling more than one error callback from a stream.
+     * These were bugs in some versions of Android that caused multiple error callbacks.
+     * Internal bug b/63087953
+     *
+     * @return true if the error callback was already called, otherwise false
+     */
+    bool wasErrorCallbackCalled() {
+        return mErrorCallbackCalled.exchange(true);
+    }
+
 protected:
 
     /**
@@ -453,7 +464,8 @@ protected:
 private:
     int                  mPreviousScheduler = -1;
 
-    std::atomic<bool>    mDataCallbackEnabled{};
+    std::atomic<bool>    mDataCallbackEnabled{false};
+    std::atomic<bool>    mErrorCallbackCalled{false};
 
 };
 
--- a/include/oboe/AudioStream.h
+++ b/include/oboe/AudioStream.h
@@ -374,8 +374,10 @@ public:
      */
     void launchStopThread();
 
+protected:
+
     /**
-     * This can be called to prevent handling more than one error callback from a stream.
+     * This is used to detect more than one error callback from a stream.
      * These were bugs in some versions of Android that caused multiple error callbacks.
      * Internal bug b/63087953
      *
@@ -384,8 +386,6 @@ public:
     bool wasErrorCallbackCalled() {
         return mErrorCallbackCalled.exchange(true);
     }
-
-protected:
 
     /**
      * Wait for a transition from one state to another.
--- a/src/aaudio/AudioStreamAAudio.cpp
+++ b/src/aaudio/AudioStreamAAudio.cpp
@@ -71,28 +71,6 @@ static void oboe_aaudio_error_thread_proc(AudioStreamAAudio *oboeStream,
     LOGD("%s() - exiting <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<", __func__);
 }
 
-// 'C' wrapper for the error callback method.
-// Launch a thread to handle the error.
-// That other thread can safely stop, close and delete the stream.
-static void oboe_aaudio_error_callback_proc(
-        AAudioStream *stream,
-        void *userData,
-        aaudio_result_t error) {
-    AudioStreamAAudio *oboeStream = reinterpret_cast<AudioStreamAAudio*>(userData);
-    if (oboeStream == nullptr) {
-        LOGE("%s() oboeStream is NULL!", __func__);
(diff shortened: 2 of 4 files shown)


src/aaudio/AudioStreamAAudio.cpp
```diff
@@ -95,6 +90,28 @@ bool AudioStreamAAudio::isSupported() {
...
 
+// Static 'C' wrapper for the error callback method.
+// Launch a thread to handle the error.
+// That other thread can safely stop, close and delete the stream.
+void AudioStreamAAudio::internalErrorCallback(
+        AAudioStream *stream,
+        void *userData,
+        aaudio_result_t error) {
+    AudioStreamAAudio *oboeStream = reinterpret_cast<AudioStreamAAudio*>(userData);
+    if (oboeStream == nullptr) {
```
philburk: Remove NULL check.
Add comment about error callbacks not getting called after stopped and closed.

milos-pesic-sc:
The change looks perfect to me! Thanks for sorting it out quickly. Looking back at number of occurrences  of the crash in `onErrorInThread` we experienced so far - it is very unlikely that we have that many crashes caused by stream disconnected errors at the same time as we close the stream in preparation for opening a new one (i.e in response to user changing track). But double error reporting could explain it. I would also be interesting in measuring how many times these unexpected scenarios -  happen in the field. What do you think about the idea of exposing some kind of optional reporting/analytics callback on library level - where these unexpected scenarios could be reported to library consumers - so that we can log them as events to our remote tracking systems (i.e Firebase) and get more detailed information for debugging?

include/oboe/AudioStream.h
```diff
@@ -376,6 +376,17 @@ class AudioStream : public AudioStreamBase {
...
 protected:
 
+    /**
+     * This is used to detect more than one error callback from a stream.
+     * These were bugs in some versions of Android that caused multiple error callbacks.
+     * Internal bug b/63087953
+     *
+     * @return true if the error callback was already called, otherwise false
+     */
+    bool wasErrorCallbackCalled() {
```
atneya: Document that this method is also what is responsible for setting the atomic bool

src/aaudio/AudioStreamAAudio.cpp
```diff
@@ -54,26 +54,21 @@ static aaudio_data_callback_result_t oboe_aaudio_data_callback_proc(
...
-        AAudioStream *stream,
-        void *userData,
-        aaudio_result_t error) {
-
-    AudioStreamAAudio *oboeStream = reinterpret_cast<AudioStreamAAudio*>(userData);
-    if (oboeStream != NULL) {
-        // Handle error on a separate thread
-        std::thread t(oboe_aaudio_error_thread_proc, oboeStream, stream, static_cast<Result>(error));
-        t.detach();
+    oboeStream->close();
```
atneya: We should add more documentation on what can and cannot be queried both before and after the stream is closed.

atneya: For example not deleting stream in onErrorBeforeClose

src/aaudio/AudioStreamAAudio.cpp
```diff
@@ -54,26 +54,21 @@ static aaudio_data_callback_result_t oboe_aaudio_data_callback_proc(
     }
 }
 
+// This runs in its own thread.
```
atneya: Note that this method can only be called once

src/aaudio/AudioStreamAAudio.cpp
```diff
@@ -95,6 +90,28 @@ bool AudioStreamAAudio::isSupported() {
...
+        AAudioStream *stream,
+        void *userData,
+        aaudio_result_t error) {
+    AudioStreamAAudio *oboeStream = reinterpret_cast<AudioStreamAAudio*>(userData);
+    if (oboeStream == nullptr) {
+        LOGE("%s() oboeStream is NULL!", __func__);
+    } else if (oboeStream->wasErrorCallbackCalled()) { // block double error callbacks
+        LOGE("%s() multiple error callbacks called!", __func__);
+    } else if (stream != oboeStream->getUnderlyingStream()) {
+        LOGD("%s() stream already closed", __func__);
```
atneya: Can this ever happen (since the underlying stream is the only think that should make this callback)?

include/oboe/AudioStream.h
```diff
@@ -453,7 +464,8 @@ class AudioStream : public AudioStreamBase {
 private:
     int                  mPreviousScheduler = -1;
 
-    std::atomic<bool>    mDataCallbackEnabled{};
+    std::atomic<bool>    mDataCallbackEnabled{false};
+    std::atomic<bool>    mErrorCallbackCalled{false};
```
atneya: Do we need to make this atomic? The only time it is mutated is `internalErrorCallback` which is only called by the underlying stream. As such, there should not be any thread safety issues, and it could be replaced with a standard bool without incurring the performance penalty of an atomic operation in a callback method (which blocks close etc.)

philburk: A lot of Oboe is about trying to protect against bugs in AudioFlinger. It is possible that more than one thread from the stream could call the error callback. The old bug was a race condition that was crashing millions of devices. I want to be very certain there will be no race this time. Plus atomic bool is very fast and this error callback only happens one (or twice). So performance is not an issue.

philburk:
> What do you think about the idea of exposing some kind of optional reporting/analytics callback on library level?

I would be interested in that. Please file a separate Issue with more details so we can track it.

include/oboe/AudioStream.h
philburk: Done.

src/aaudio/AudioStreamAAudio.cpp
philburk: The docs for queries allowed after close should be another PR.
I warned users now not to delete in this method.

src/aaudio/AudioStreamAAudio.cpp
philburk: This was what caused the assert in the old code. If it happens again then this will help us debug the problem.

src/aaudio/AudioStreamAAudio.cpp
atneya: Opened issue #536

PR #544, opened 2019-06-11
Deprecating old getTimestamp method. Fixes #478

### LINKED ISSUE #478: One of the two getTimestamp()methods is not in the documentation
virtual Result getTimestamp(clockid_t clockId,
                                int64_t *framePosition,
                                int64_t *timeNanoseconds)

is in the doxygen docs, but I stumbled upon 

virtual ResultWithValue<FrameTimestamp> getTimestamp(clockid_t clockId)

in a code example and couldn't find it in the docs.

atneya:


philburk:
The "Fixes #478" needs to be in a comment, maybe a commit message, in order for GitHub to recognize it. Putting it in the title is good but not enough.

PR #553, opened 2019-06-17
tests: make sure OpenSL ES rejects bad formats

philburk:


tests/testStreamOpen.cpp
```diff
@@ -85,14 +85,39 @@ TEST_F(StreamOpen, OutputForOpenSLESPerformanceModeShouldBeNone){
...
 	ASSERT_EQ((int)mStream->getPerformanceMode(), (int)PerformanceMode::None);
     closeStream();
 }
+
+TEST_F(StreamOpen, ForOpenSlesIllegalFormatRejectedOutput) {
+    mBuilder.setAudioApi(AudioApi::OpenSLES);
+    mBuilder.setPerformanceMode(PerformanceMode::LowLatency);
+    mBuilder.setFormat(static_cast<AudioFormat>(666));
+    Result r = mBuilder.openStream(&mStream);
+    EXPECT_NE(r, Result::OK) << "Should not open stream " << convertToText(r);
```
dturner: Wouldn't it be better to check for `oboe::Result::ErrorInvalidFormat`?

tests/testStreamOpen.cpp
```diff
@@ -85,14 +85,39 @@ TEST_F(StreamOpen, OutputForOpenSLESPerformanceModeShouldBeNone){
...
+    }
+}
+
+TEST_F(StreamOpen, ForOpenSlesIllegalFormatRejectedInput) {
+    mBuilder.setAudioApi(AudioApi::OpenSLES);
+    mBuilder.setPerformanceMode(PerformanceMode::LowLatency);
+    mBuilder.setDirection(Direction::Input);
+    mBuilder.setFormat(static_cast<AudioFormat>(666));
+    Result r = mBuilder.openStream(&mStream);
+    EXPECT_NE(r, Result::OK) << "Should not open stream " << convertToText(r);
```
dturner: Wouldn't it be better to check for `oboe::Result::ErrorInvalidFormat`?

tests/testStreamOpen.cpp
philburk: It is currently returning ErrorInternal.  

I think we should add some more general parameter checking to AudioStreamBuilder. That will get called in the open() and return a more suitable error message.

tests/testStreamOpen.cpp
philburk: ditto

philburk:
The rebase and merge is not working because this is too old.
So I will resubmit this on a different branch.

PR #561, opened 2019-06-21
Only load API 28 methods on API 28 and above. Fixes #149

### LINKED ISSUE #149: Errors shown in logcat when creating an audio stream on API < 28
Steps to repro

- Run the megadrone sample on a device running API 27 or lower (I tested on Pixel XL emulator running API 27)
- Open logcat

Expected result

- No errors are shown when creating the audio stream

Actual result

The following errors are shown:

    07-02 16:51:51.247 22866-22866/com.example.oboe.megadrone E/OboeAudio: AAudioLoader could not find AAudioStream_getUsage
    AAudioLoader could not find AAudioStream_getContentType
    AAudioLoader could not find AAudioStream_getInputPreset
    07-02 16:51:51.248 22866-22866/com.example.oboe.megadrone E/OboeAudio: AAudioLoader could not find AAudioStream_getSessionId

Suggest that these errors are only shown if API == 28 and the dlsym operations fail.

dturner:


atneya:
It seems we already have nullchecks for the methods in `AudioStreamAAudio.cpp`: 
https://github.com/google/oboe/blob/44a5a49323a936b511dadada4b89d23c0e90aed3/src/aaudio/AudioStreamAAudio.cpp#L161

What currently happens if those methods are attempted to be loaded on API <28? Are they set to null regardless? If so this check may not be necessary.

philburk:
The existing code is safe as far as having all the right null checks. But it does print an error message that can be disturbing. Since we are trying to eliminate unnecessary error messages this change would be good.

atneya:
Makes sense.

PR #571, opened 2019-07-10
Mark Fifo getters as consts

tguilbert-google:
This addresses some of the comments in #352. It makes
FifoControllerBase's getters consts, and it uses default destructors.


tguilbert-google:
Refreshing CLA check

PR #573, opened 2019-07-11
Use uint32_t for frames and bytes in FifoBuffer

tguilbert-google:
FifoBuffer::convertFramesToBytes() uses signed types, and uses negative
values as a form of error checking.

This CL updates FifoBuffer to used uint32_t when representing frames and
bytes, and changes the error checking to rely on comparisons rather than
negative values.

### DIFF
--- a/src/fifo/FifoBuffer.cpp
+++ b/src/fifo/FifoBuffer.cpp
@@ -80,14 +80,14 @@ FifoBuffer::~FifoBuffer() {
 }
 
 
-int32_t FifoBuffer::convertFramesToBytes(int32_t frames) {
+uint32_t FifoBuffer::convertFramesToBytes(uint32_t frames) {
     return frames * mBytesPerFrame;
 }
 
 int32_t FifoBuffer::read(void *buffer, int32_t numFrames) {
     int32_t framesAvailable = mFifo->getFullFramesAvailable();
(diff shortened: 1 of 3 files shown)


src/fifo/FifoBuffer.cpp
```diff
@@ -99,28 +99,26 @@ int32_t FifoBuffer::read(void *buffer, int32_t numFrames) {
     uint8_t *destination = reinterpret_cast<uint8_t *>(buffer);
     uint8_t *source = &mStorage[convertFramesToBytes(readIndex)];
     if ((readIndex + framesToRead) > mFrameCapacity) {
-        // read in two parts, first part here is at the end of the mStorage buffer
-        uint32_t frames1 = mFrameCapacity - readIndex;
-        int32_t numBytes = convertFramesToBytes(frames1);
-        if (numBytes < 0) {
+        // First, read frames until the end of the mStorage buffer.
+        if(readIndex > mFrameCapacity) {
```
tguilbert-google: It seems like this should never happen, if we trust mFifo to be internally consistent. Should this check still be here?

tguilbert-google:
Thank you for the clarification!

tguilbert-google:
Abandoning these changes

PR #575, opened 2019-07-12
Add atomic increment to FifoControllers

tguilbert-google:
This CL adds an increment counter methods that allow for atomically
incrementing the read/write indexes, rather than geting the current
index, adding to it and setting it.

Also, getFullFramesAvailable() and getEmptyFramesAvailable() now clamp
negative values to 0.

### DIFF
--- a/src/fifo/FifoBuffer.cpp
+++ b/src/fifo/FifoBuffer.cpp
@@ -85,7 +85,7 @@ int32_t FifoBuffer::convertFramesToBytes(int32_t frames) {
 }
 
 int32_t FifoBuffer::read(void *buffer, int32_t numFrames) {
-    uint32_t framesAvailable = mFifo->getFullFramesAvailable();
+    int32_t framesAvailable = mFifo->getFullFramesAvailable();
     int32_t framesToRead = numFrames;
     // Is there enough data in the FIFO
     if (framesToRead > framesAvailable) {
@@ -129,7 +129,7 @@ int32_t FifoBuffer::read(void *buffer, int32_t numFrames) {
 }
 
 int32_t FifoBuffer::write(const void *buffer, int32_t framesToWrite) {
-    uint32_t framesAvailable = mFifo->getEmptyFramesAvailable();
+    int32_t framesAvailable = mFifo->getEmptyFramesAvailable();
     if (framesToWrite > framesAvailable) {
         framesToWrite = framesAvailable;
     }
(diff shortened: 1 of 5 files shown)


src/fifo/FifoController.h
```diff
@@ -32,19 +32,24 @@ class FifoController : public FifoControllerBase
     FifoController(uint32_t bufferSize, uint32_t threshold);
     virtual ~FifoController() = default;
 
-    // TODO review use atomics or memory barriers
```
tguilbert-google: From what I could tell, the current use of memory barriers is adequate to keep mWriteCounter and mReadCounter consistent. However, I don't know if there could be problems in FifoControllerBase::getFullFramesAvailable(), when we access both variables. I can't tell if there is a scenario where two threads would get a different number of frames, that could be prevented by a use of std::memory_order_seq_cst instead. From here (https://en.cppreference.com/w/cpp/atomic/memory_order#Sequentially-consistent_ordering) :
"Sequential ordering may be necessary for multiple producer-multiple consumer situations where all consumers must observe the actions of all producers occurring in the same order."

How many producers/consumers can FifoBuffer have?

tguilbert-google: Useful video:
https://youtu.be/ZQFzMfHIxng?t=3117

philburk: This FIFO has only one reader and one writer.

src/fifo/FifoBuffer.cpp
```diff
@@ -85,7 +85,7 @@ int32_t FifoBuffer::convertFramesToBytes(int32_t frames) {
 }
 
 int32_t FifoBuffer::read(void *buffer, int32_t numFrames) {
-    int32_t framesAvailable = mFifo->getFullFramesAvailable();
+    uint32_t framesAvailable = mFifo->getFullFramesAvailable();
```
philburk: Please do not convert these to unsigned. See comments on https://github.com/google/oboe/pull/573

src/fifo/FifoControllerBase.cpp
```diff
@@ -30,33 +30,34 @@ FifoControllerBase::FifoControllerBase(uint32_t totalFrames, uint32_t threshold)
 {
 }
 
-int32_t FifoControllerBase::getFullFramesAvailable() const {
-    return static_cast<int32_t>(getWriteCounter() - getReadCounter());
+uint32_t FifoControllerBase::getFullFramesAvailable() const {
```
philburk: This value can go negative if there is an underflow! Please do not convert it to signed.

PR #577, opened 2019-07-15
Small non-functional changes

tguilbert-google:
This CL addresses some of the simple changes suggested
in https://github.com/google/oboe/pull/352

### DIFF
--- a/include/oboe/AudioStream.h
+++ b/include/oboe/AudioStream.h
@@ -63,7 +63,7 @@ public:
      *
      * @return
      */
-    virtual Result open();
+    virtual Result open() = 0;
 
     /**
      * Close the stream and deallocate any resources from the open() call.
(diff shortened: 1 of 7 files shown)


src/opensles/AudioStreamBuffered.h
```diff
@@ -64,7 +64,7 @@ class AudioStreamBuffered : public AudioStream {
     // If there is no callback then we need a FIFO between the App and OpenSL ES.
     bool usingFIFO() const { return getCallback() == nullptr; }
 
-    virtual Result updateServiceFrameCounter() { return Result::OK; }
+    virtual Result updateServiceFrameCounter() = 0;
```
dturner: How did you know that this could be made a pure virtual? Do you have a specific compiler warning on?

tguilbert-google: I tried it, following a comment in #352. Building a project that uses Oboe still worked. I think making this pure virtual also makes AudioStreamOpenSLES pure virtual, but it is never instantiated, and only the Input/Ouput subclasses are.

include/oboe/AudioStream.h
```diff
@@ -63,7 +63,7 @@ class AudioStream : public AudioStreamBase {
      *
      * @return
      */
-    virtual Result open();
+    virtual Result open() = 0;
```
philburk: This broke the build. See https://github.com/google/oboe/issues/582

PR #578, opened 2019-07-15
Use nullptr and <cassert>

tguilbert-google:
This CL cleans up the code to uniformly use the same assert
headers, and to use nullptr instead of NULL whenever appropriate.

### DIFF
--- a/src/opensles/AudioStreamOpenSLES.cpp
+++ b/src/opensles/AudioStreamOpenSLES.cpp
@@ -13,7 +13,7 @@
  * limitations under the License.
  */
 #include <sys/types.h>
-#include <assert.h>
+#include <cassert>
 #include <android/log.h>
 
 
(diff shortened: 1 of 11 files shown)


src/opensles/AudioStreamOpenSLES.cpp
```diff
@@ -13,7 +13,7 @@
  * limitations under the License.
  */
 #include <sys/types.h>
-#include <cassert>
+#include <assert.h>
```
dturner: Just a question: since we're doing `NULL` -> `nullptr` which is C to C++, why are we then doing `<cassert>` to `<assert.h>` which is C++ to C header?

tguilbert-google: The idea was just to get all headers uniform, following a comment in #352. You are right, I will convert <assert.h> into <cassert> instead

tguilbert-google: *into <cassert>

PR #595, opened 2019-07-31
Update CI to download and install sdkmanager inside CI scope

ggfan:
There was a failure that CI failed due to sdkmanager is not found. This was mostly suspected to be Travis CI server update issue. It seems to be fixed on Travis side, but to keep ourselves from future glitches, I propose to let CI download sdkmanager and not depending on Travis-CI's copy.

### DIFF
--- a/.ci_tools/before_install.sh
+++ b/.ci_tools/before_install.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+# Handler for before_install step in CI
+# Functionalities:
+# 1.  install sdkmanager manually
+# 2.  TODO: install whole new SDK
+# 3.  TODO: point sdk location to our installed one ( not the travis's )
+# Side Effect:
+#     $CI_SDK_DIR is left behind on the disk: could not delete.
+
+set -e
+
+CI_SDK_DIR=$HOME/ci_sdk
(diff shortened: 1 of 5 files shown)


.ci_tools/before_install.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+# Handler for before_install step in CI
+# Functionalities:
+# 1.  install sdkmanager manually
+# 2.  TODO: install whole new SDK
+# 3.  TODO: point sdk location to our installed one ( not the travis's )
```
philburk: Does this mean that the build will not use the downloaded SDK?!

.ci_tools/before_install.sh
```diff
@@ -0,0 +1,22 @@
...
+# Side Effect:
+#     $CI_SDK_DIR is left behind on the disk: could not delete.
+
+set -e
+
+CI_SDK_DIR=$HOME/ci_sdk
+SDK_TOOL_FILENAME=sdk-tools-linux-4333796.zip
```
philburk: It looks like the script now has version dependencies. So we will have to update them when the SDK is updated.

PR #598, opened 2019-08-07
Commenting out parameters to use Wall compiler flag

atneya:
Allows clients to use stronger compiler warning flags  (like Wall).


philburk:
Good. In case others are curious, commenting out parameters is recommended by the Android C++ Style Guide as the right way to avoid "unused parameter" warnings.

Issue #606, opened 2019-08-15
Travis CI should build OboeTester


philburk:
The Travis build checks the sample code.
It should also try to build apps/OboeTester because it covers more of the Oboe API.

ggfan:
This is about the apps/OboeTester, correct?

PR #609, opened 2019-08-20
Add OboeTester to CI build

ggfan:
this is to address https://github.com/google/oboe/issues/606, tested on my local Linux, works fine. 
please take a look, thank you!

### DIFF
--- a/.ci_tools/build_samples.sh
+++ b/.ci_tools/build_samples.sh
@@ -8,6 +8,7 @@ REPO_ROOT_DIR=.
 
 declare projects=(
     samples
+    apps/OboeTester
 )
 
 for d in "${projects[@]}"; do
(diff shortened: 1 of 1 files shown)


.ci_tools/build_samples.sh
```diff
@@ -23,6 +24,7 @@ declare bundles=(
     samples/MegaDrone/build/outputs/bundle/debug/MegaDrone.aab
     samples/RhythmGame/build/outputs/bundle/ndkExtractorDebug/RhythmGame.aab
     samples/LiveEffect/build/outputs/bundle/debug/LiveEffect.aab
+    apps/OboeTester/app/build/outputs/bundle/debug/app.aab
```
philburk: Why is OboeTester using "app.aab" instead of OboeTester.aab?
I'm worried that if I change OboeTester build then I might break this.

philburk: Which file defines the bundle name?

ggfan: the directory "app" defined the apk name, by default ( in apps/OboeTester/app <-- this 'app')

Issue #615, opened 2019-08-27
crash in LiveEffect when create delete


zhongxuerui521:
Android Version(s): 8.0.0
Android Device(s): Honor7C
Oboe Version: 1.2.0

android::RefBase::decStrong(void const*) const 0x00000000c246cd3c
android::IPCThreadState::processPendingDerefs() 0x00000000eaf0aa96
android::IPCThreadState::joinThreadPool(bool) 0x00000000eaf0aad8
<unknown> 0x00000000eaf21168
android::Thread::_threadLoop(void*) 0x00000000c246f244
android::AndroidRuntime::javaThreadShell(void*) 0x00000000e988202e
__pthread_start(void*) 0x00000000e8493674
__start_thread 0x00000000e8465ccc

__dl___libc_format_buffer_va_list 0x00000000eb809c7c
__dl__ZL11fatal_errnoPKcz 0x00000000eb80a846
<unknown> 0x00000000eb881d04

LiveEffectEngine.create();
LiveEffectEngine.setEffectOn(true);
LiveEffectEngine.setEffectOn(false);
LiveEffectEngine.delete();
LiveEffectEngine.setEffectOn(true);
LiveEffectEngine.create();
then it's crash

philburk:
Thanks for reporting this. I got a crash when I run LiveEffect the first time after installing it. 

It crashes in the background and then leaves the microphone permission dialog showing. if I grant permission then it works the second time.

Is that the same problem you are seeing?

philburk:
@zhongxuerui521 - please try the fix in #617

philburk:
Assuming that #617 fixed this. Closing.

PR #617, opened 2019-08-29
LiveEffect: fix crash if audio permission needed

philburk:
stop() died because of null stream


philburk:
This fixes the crash that I get in LiveEffect whe the app has no mic permission.

https://github.com/google/oboe/issues/615

PR #638, opened 2019-09-20
OboeTester: add menu for InputPreset

philburk:
Also bump to code 22, name 1.5.14


philburk:
Bumps version to 1.5.15

PR #642, opened 2019-09-22
OboeTester: add Tolerance fader

philburk:
Adjust sensitivity of Glitch Test


apps/OboeTester/app/src/main/java/com/google/sample/oboe/manualtest/AnalyzerActivity.java
```diff
@@ -132,7 +132,7 @@ protected void resetConfiguration() {
     public void startAudio() {
         if (mBufferSizeView != null && mBufferSizeView.isEnabled()) {
             mBufferSizeView.updateBufferSize();
-            mBufferSizeView.setEnabled(false);
+   //         mBufferSizeView.setEnabled(false);
```
philburk: do this later

Issue #670, opened 2019-10-16
set initial size in LatencyTuner to 2*burst


philburk:
Based on comment in #662

philburk:
just need a 2*

https://github.com/google/oboe/blob/ec8de0feef6df0af28d8de1d85437e2b7620c92e/src/common/LatencyTuner.cpp#L97

PR #673, opened 2019-10-21
Double initial stream buffer size.

android-usb:
Issue #670

### DIFF
--- a/src/common/LatencyTuner.cpp
+++ b/src/common/LatencyTuner.cpp
@@ -94,7 +94,7 @@ void LatencyTuner::reset() {
     mState = State::Idle;
     mIdleCountDown = kIdleCount;
     // Set to minimal latency
-    mStream.setBufferSizeInFrames(2*mStream.getFramesPerBurst());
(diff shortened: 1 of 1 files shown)


src/common/LatencyTuner.cpp
```diff
@@ -94,7 +94,7 @@ void LatencyTuner::reset() {
     mState = State::Idle;
     mIdleCountDown = kIdleCount;
     // Set to minimal latency
-    mStream.setBufferSizeInFrames(mStream.getFramesPerBurst());
+    mStream.setBufferSizeInFrames(2*mStream.getFramesPerBurst());
```
philburk: Spaces around the " * ".

philburk:
> Issue #670

If you write:

    Fixes #670 

in a Pull Request comment then it will automatically close the Issue when merged.

src/common/LatencyTuner.cpp
android-usb: Done.

PR #706, opened 2019-11-20
Fix memory leak and add to namespace

jimmytobin2425:
GetIntArrayElements needs to be paired with ReleaseIntArrayElements. elements will be released once the unique_ptr goes out of scope.

### DIFF
--- a/samples/MegaDrone/src/main/cpp/native-lib.cpp
+++ b/samples/MegaDrone/src/main/cpp/native-lib.cpp
@@ -21,20 +21,19 @@
 #include "MegaDroneEngine.h"
 
 
+namespace {
+
 std::vector<int> convertJavaArrayToVector(JNIEnv *env, jintArray intArray){
 
-    std::vector<int> v;
-    jsize length = env->GetArrayLength(intArray);
+    jsize length;
+    std::unique_ptr<jint[]> elements =
+        CHECK_JNI(FATAL, env).GetIntArrayElements(intArray, &length);
 
-    if (length > 0) {
-        jboolean isCopy;
-        jint *elements = env->GetIntArrayElements(intArray, &isCopy);
-        for (int i = 0; i < length; i++) {
-            v.push_back(elements[i]);
-        }
-    }
-    return v;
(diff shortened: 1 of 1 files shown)


philburk:
When the unique_ptr or goes out of scope it will call delete on the jint array. Is that the full equivalent of calling the release function?

jimmytobin2425:
Hey Phil,

The GetIntArrayElements will allocate the array. When the unique_ptr goes out of scope it will delete the array. This is the same behavior as ReleaseIntArrayElements.

philburk:
I'm not convinced that using unique_ptr here is safe. I believe GetIntArrayElements() will increment the reference count. So the ReleaseIntArrayElements() is needed to decrement it. Also the data is not always copied so you might be freeing the original Java array! 

But there is a leak that needs to be fixed. Could you just add ReleaseIntArrayElements() to the original?

Also please sign the CLA above or register as a Googler on GitHub.

philburk:
This patch is failing in Travis because of:
  /home/travis/build/google/oboe/samples/MegaDrone/src/main/cpp/native-lib.cpp:30:19: error: use of undeclared identifier 'FATAL'
          CHECK_JNI(FATAL, env).GetIntArrayElements(intArray, &length);

samples/MegaDrone/src/main/cpp/native-lib.cpp
```diff
@@ -21,20 +21,19 @@
...
 
+namespace {
+
 std::vector<int> convertJavaArrayToVector(JNIEnv *env, jintArray intArray){
 
-    std::vector<int> v;
-    jsize length = env->GetArrayLength(intArray);
+    jsize length;
+    std::unique_ptr<jint[]> elements =
+        CHECK_JNI(FATAL, env).GetIntArrayElements(intArray, &length);
```
philburk: GetIntArrayElements() does not take a length pointer. It takes a pointer to a boolean that indicates whether a copy of the data was made.

Also we need to balance this with ReleaseIntArrayElements so that the Java resource is free up.

philburk:
See alternative CL at https://github.com/google/oboe/pull/714

Issue #659, opened 2019-10-04
OboeTester should average multiple round trip latency results


philburk:
Also need to update the Intent driven test.

Maybe average ten results from a button.

PR #709, opened 2019-11-22
OboeTester: add Average button for latency

philburk:
Run the latency test 10 times and report the average latency.


philburk:
Fixes #659

PR #733, opened 2019-12-18
Update NDKExtractor.cpp

gatzsche:
Rhythm Game Example - Last buffer is not read from m4a files

### DIFF
--- a/samples/RhythmGame/src/main/cpp/audio/NDKExtractor.cpp
+++ b/samples/RhythmGame/src/main/cpp/audio/NDKExtractor.cpp
@@ -154,7 +154,7 @@ int32_t NDKExtractor::decode(AAsset *asset, uint8_t *targetData, AudioProperties
                 if (info.flags & AMEDIACODEC_BUFFER_FLAG_END_OF_STREAM){
                     LOGD("Reached end of decoding stream");
                     isDecoding = false;
-                } else {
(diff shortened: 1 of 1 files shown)


gatzsche:
@googlebot I signed it!

samples/RhythmGame/src/main/cpp/audio/NDKExtractor.cpp
```diff
@@ -154,23 +154,22 @@ int32_t NDKExtractor::decode(AAsset *asset, uint8_t *targetData, AudioProperties
...
+                /*LOGV("Got output buffer index %d, buffer size: %d, info size: %d writing to pcm index %d",
+                     outputIndex,
+                     outputSize,
+                     info.size,
+                     m_writeIndex);*/
+
+                // copy the data out of the buffer
+                memcpy(targetData + bytesWritten, outputBuffer, info.size);
+                bytesWritten+=info.size;
+                AMediaCodec_releaseOutputBuffer(codec, outputIndex, false);
```
gatzsche: The following code needs to be executed also if end of stream has been reached. Otherwise last samples are not decoded.

dturner: Good spot - thanks.

Issue #654, opened 2019-10-02
Limit setBufferSize to reduce glitches


philburk:
In QuirksManager.cpp, based on device and version, prevent setBufferSize from going too low or too high.

philburk:
Provide flag to enable workaround so Oboetester can still test underlyng API.
Provide stream to QuirksManager.
Add method to clip bufferSize.

dturner:
Solution for now: For all Samsung models we will not allow the buffer size to be <2 bursts. This is because at 1 burst you may get glitches caused by timing issues, and you won't receive any increase in underrun count so you may not be aware.

paulorrfilho:
I think this is duplicated with #564

paulorrfilho:
Currently I am working on porting my iOS app, DrumKnee, to Android but the constant glitches are a deal breaker. Hopefully someone will pick this one up shortly.

philburk:
We have added a range limit to setBufferSize in the QuirksManager. This will not fix all glitches forever. But it will help. This task is finished so closing.

PR #734, opened 2019-12-19
Oboe: Add device specific Quirks for Samsung

philburk:
Use a minimum of 2 bursts for MMAP on Exynos devices.
Add OboeGlobals::setWorkaroundEnabled() for testing
so we can repro AAudio bugs using Oboe.

Note Oboe API change. OboeGlobals added.

Test: run OboeTester, look at MMAP output buffer size min
Test: Disable and enable workarounds from MainActivity.


include/oboe/Definitions.h
```diff
@@ -495,6 +495,25 @@ namespace oboe {
...
 
+    class OboeGlobals {
+    public:
+
+        static bool areWorkaroundsEnabled() {
+            return mWorkaroundsEnabled;
+        }
+
+        /**
+         * Test programs that want to reproduce bugs in AAudio or OpenSL ES
```
philburk: Not clear.

include/oboe/Definitions.h
```diff
@@ -495,6 +495,25 @@ namespace oboe {
...
+         * that have workarounds in Oboe, can disable them by setting this false.\\\\
```
philburk: Remove \\\\

philburk:
This is related to - Glitches on S8,S9  #564
and - Limit setBufferSize to reduce glitches #654

dturner:
Tested on Samsung Galaxy S9 SM-G960F Build number PPR1.180610.011.G960FXXU2CSA2 (Exynos 9810)

Setup: 
- Build and run OboeTester (from this quirksize branch)
- Connect headphones using 3.5mm headphone socket

Steps to test (workarounds OFF): 
- Tap TEST OUTPUT
- Tap OPEN then START

Outcome: 
- Sine wave with lots of audible artifacts (crackles, pops etc) is heard
- Buffer size listed as 96 frames

Steps to test (workarounds ON)
- Tap back
- Check "enable Oboe workarounds"
- Tap TEST OUTPUT
- Tap OPEN then START 

Outcome: 
- Less audible artifacts heard, but still very noticeable. I'd estimate at one artifact per second. These do not show up as xruns. 
- Buffer size is set to 192

Summary: Increasing the buffer size definitely reduces the number of glitches, but doesn't solve the problem entirely.

include/oboe/Definitions.h
```diff
@@ -495,6 +495,25 @@ namespace oboe {
         int64_t timestamp; // in nanoseconds
     };
 
+    class OboeGlobals {
```
dturner: I'm fine with these being set here. I wonder if the default stream values should be moved into this class as well.

dturner:
@dturner Take a look at the S9 properties to correctly identify it, then choose a buffer size which eliminates glitches.

dturner:
Do `adb shell getprop` then grep for vendor, should get model no etc. Send to Phil.

include/oboe/Definitions.h
dturner: Add getters/setters for the default stream values e.g. `setOpenSLESDefaultSampleRate` etc 
Add comment saying that direct access to `DefaultStreamValues` is deprecated.

dturner: One other thing. It might be better to name this `Globals` rather than `OboeGlobals` since this is already in the `oboe` namespace.

Nordhal:
Will it be included on the 1.3.3 then ? The version was bumped from the quirksize branch but not merged even at oboe 1.3.2

Issue #727, opened 2019-12-13
OboeTester needs a Load fader for debugging


philburk:
Some bugs only appear because of a collision with the data callback. if the callback is too short then the bug may not happen. Loading the callback can help.
Pull random number workload from cpu hints test.
Replace the Amplitude fader with a Load fader.

philburk:
For internal issues b/145027922 and b/141241120

PR #737, opened 2019-12-21
OboeTester: add Workload fader that will burn up the CPU

philburk:
This can be used to stress the system and cause glitches.
It can also be used to make race conditions more likely.


philburk:
This Fixes #727

PR #743, opened 2019-12-28
Add BeatScratch to list of apps using Oboe

JonLatane:


JonLatane:
@googlebot I signed it!

philburk:
Thanks for using Oboe and adding the link.

PR #745, opened 2019-12-30
Avoid global variables

katyo:
Using global variables in c++ gives the problems with integrating with other languages.
In particular the linking errors like that: https://github.com/rust-lang/rust/issues/36710.
So I think we can and should avoid global variables at all.

### DIFF
--- a/src/flowgraph/resampler/IntegerRatio.cpp
+++ b/src/flowgraph/resampler/IntegerRatio.cpp
@@ -14,14 +14,12 @@
  * limitations under the License.
  */
 
-#include <vector>
-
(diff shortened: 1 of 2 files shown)


katyo:
@googlebot I signed it!

philburk:
The kPrimes vector is not really global. It is a static member of IntegerRatio.
But it seems simpler to not use the vector so I will merge it.
Thanks for the patch.

katyo:
@philburk Of course, I mean that `kPrimes` actually was variable which constructed/destructed at runtime, despite the fact that it actually used as a constant.
Thanks for merge.

Issue #751, opened 2020-01-06
when I change the sampling rate with setSampleRate(), the numFrames parameter of onAudioReady() will not be reflected at the same rate.


erlee-hdtel:
Android Version(s): 8.0.0
Android Device(s): Samsung galaxy s7
Oboe Version: 1.3

**Short Description**
when I change the sampling rate with setSampleRate(), the numFrames parameter of onAudioReady() will not be reflected at the same rate.

**Steps To Reproduce**
When open oboe input stream, set the builder like followings.
```cpp
oboe::AudioStreamBuilder builder;
builder.setDirection(oboe::Direction::Input);
builder.setPerformanceMode(oboe::PerformanceMode::LowLatency);
builder.setSampleRate(8000);
builder.setFramesPerCallback(160);
```

**Expected behavior**
when I set sample rate as 8000, I want onAudioReady() 's numFrames to be 160.

**Actual behavior**
numFrames is always 960.

**Additional context**
when I modify configureBufferSizes() function in AudioStreamOpenSLES.cpp like following, It works.

```cpp
    // For high latency streams, use a larger buffer size.
    // Performance Mode support was added in N_MR1 (7.1)

    int32_t framesPerHighLatencyBurstWithResamplingRate = kFramesPerHighLatencyBurst;
    if (mSampleRate > 0 && mSampleRate < DefaultStreamValues::SampleRate) {
        framesPerHighLatencyBurstWithResamplingRate = kFramesPerHighLatencyBurst * mSampleRate / DefaultStreamValues::SampleRate;
    }

    if (getSdkVersion() >= __ANDROID_API_N_MR1__
        && mPerformanceMode != PerformanceMode::LowLatency
        && mFramesPerBurst < framesPerHighLatencyBurstWithResamplingRate) {
        // Find a multiple of framesPerBurst >= resampleFramesPerHighLatencyBurst.
        int32_t numBursts = (framesPerHighLatencyBurstWithResamplingRate + mFramesPerBurst - 1) / mFramesPerBurst;
        mFramesPerBurst *= numBursts;
        LOGD("AudioStreamOpenSLES:%s() NOT low latency, set mFramesPerBurst = %d",
             __func__, mFramesPerBurst);
    }
    mFramesPerCallback = mFramesPerBurst;
```
As you can see, I set the performance mode to PerformanceMode::LowLatency, but when check mPerformanceMode value, it isn't PerformanceMode::LowLatency.
so, I add upper code.

[...]

PR #760, opened 2020-01-15
fix 8KHz resampling frame buffer size for issue #751

erlee-hdtel:
when change the sampling rate with setSampleRate(8000), let the numFrames parameter of onAudioReady() be 20ms frame buffer size(160FPS).

### DIFF
--- a/src/opensles/AudioStreamOpenSLES.cpp
+++ b/src/opensles/AudioStreamOpenSLES.cpp
@@ -102,11 +102,17 @@ Result AudioStreamOpenSLES::configureBufferSizes() {
     }
     // For high latency streams, use a larger buffer size.
     // Performance Mode support was added in N_MR1 (7.1)
+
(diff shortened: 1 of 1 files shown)


erlee-hdtel:
@googlebot I signed it!

philburk:
GitHub looks for this kind of comment:
Fixes #751

philburk:
**Thank you** for reporting the bug and finding a fix. Based on your code, I am working now on a slightly different patch that calculates the buffer size based on a 20 msec target.

philburk:
@erlee-hdtel - Does https://github.com/google/oboe/pull/762 also fix the problem?

erlee-hdtel:
When I run with #762 , It works good. thank you.

philburk:
I merged #762 so I will close this one.
Thanks for reporting the problem and suggesting a fix.

PR #773, opened 2020-01-27
honor setFramesPerCallback()

### LINKED ISSUE #766: callback numFrames not matching value from setFramesPerCallback()
Android Version(s): Android 8.0
Android Device(s): Samsung Galaxy S7.
Oboe Version: 1.3.0

Using audioStream.setFramesPerCallback(256) I assume In audioReady callback numFrames = 256. And so it is. But only not for Samsung Galaxy S7, for this device 1024 was returned. 
So sound very glitch.

But if checkout 1.2-stable all is correct.

philburk:
Fixes #766

### DIFF
--- a/tests/testStreamOpen.cpp
+++ b/tests/testStreamOpen.cpp
@@ -20,6 +20,15 @@
 
 using namespace oboe;
 
+class NumFramesCallback : public AudioStreamCallback {
+public:
+    DataCallbackResult onAudioReady(AudioStream *oboeStream, void *audioData, int32_t numFrames) override {
+		framesPerCallback = numFrames;
+        return DataCallbackResult::Continue;
+    }
+    
+    std::atomic<int32_t> framesPerCallback{0};
+};
 
 class StreamOpen : public ::testing::Test {
 
@@ -28,6 +37,8 @@ protected:
     bool openStream(){
         Result r = mBuilder.openStream(&mStream);
         EXPECT_EQ(r, Result::OK) << "Failed to open stream " << convertToText(r);
+        EXPECT_EQ(0, openCount) << "Should start with a fresh object every time.";
+        openCount++;
         return (r == Result::OK);
     }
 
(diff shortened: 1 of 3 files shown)


tests/testStreamOpen.cpp
```diff
@@ -93,6 +108,80 @@ TEST_F(StreamOpen, InputForOpenSLESPerformanceModeShouldBeNone){
 	ASSERT_EQ((int)mStream->getPerformanceMode(), (int)PerformanceMode::None);
     closeStream();
 }
+
+// Make sure the callback is called with the requested FramesPerCallback
+TEST_F(StreamOpen, OpenSLESFramesPerCallback) {
+	const int kRequestedFramesPerCallback = 417;
+	NumFramesCallback callback;
```
dturner: Indentation is off on these 2 lines

tests/testStreamOpen.cpp
```diff
@@ -20,6 +20,15 @@
 
 using namespace oboe;
 
+class NumFramesCallback : public AudioStreamCallback {
```
dturner: I feel like `NumFramesCallback` doesn't make it obvious what this class is doing. Could it be renamed to something like `CallbackSizeMonitor`

tests/testStreamOpen.cpp
```diff
@@ -20,6 +20,15 @@
...
+public:
+    DataCallbackResult onAudioReady(AudioStream *oboeStream, void *audioData, int32_t numFrames) override {
+		framesPerCallback = numFrames;
+        return DataCallbackResult::Continue;
+    }
+    
+    std::atomic<int32_t> framesPerCallback{0};
```
dturner: Maybe add a comment saying `// This is exposed publicly so that the number of frames per callback can be tested`.

tests/testStreamOpen.cpp
philburk: Done. Fixed editor preferences.

tests/testStreamOpen.cpp
philburk: Done

PR #799, opened 2020-03-02
Refactored SampleBuffer into SampleSource & SampleBuffer

android-usb:
Also... Fixed up namespaces for library refactor.

### DIFF
--- a/samples/drumthumper/src/main/cpp/DrumPlayerJNI.cpp
+++ b/samples/drumthumper/src/main/cpp/DrumPlayerJNI.cpp
@@ -22,7 +22,7 @@
 
 #include <android/log.h>
 
-#include <player/OneShotSampleBuffer.h>
+#include <player/OneShotSampleSource.h>
 #include <player/SimpleMultiPlayer.h>
 
 static const char* TAG = "DrumPlayerJNI";
@@ -32,7 +32,7 @@ static const char* TAG = "DrumPlayerJNI";
 extern "C" {
 #endif
 
-using namespace wavlib;
+using namespace iolib;
 
 static SimpleMultiPlayer sDTPlayer;
 
--- a/samples/drumthumper/src/main/cpp/DrumPlayerJNI.cpp
+++ b/samples/drumthumper/src/main/cpp/DrumPlayerJNI.cpp
@@ -40,22 +40,32 @@ static SimpleMultiPlayer sDTPlayer;
  * Native (JNI) implementation of DrumPlayer.setupAudioStreamNative()
  */
 JNIEXPORT void JNICALL Java_com_google_oboe_sample_drumthumper_DrumPlayer_setupAudioStreamNative(
-        JNIEnv* env, jobject, jint numSampleBuffers, jint numChannels, jint sampleRate) {
+        JNIEnv* env, jobject, jint numChannels, jint sampleRate) {
     __android_log_print(ANDROID_LOG_INFO, TAG, "%s", "init()");
 
     // we know in this case that the sample buffers are all 1-channel, 41K
-    sDTPlayer.setupAudioStream(numSampleBuffers, numChannels, sampleRate);
+    sDTPlayer.setupAudioStream(numChannels, sampleRate);
 }
 
 /**
  * Native (JNI) implementation of DrumPlayer.teardownAudioStreamNative()
  */
-JNIEXPORT void JNICALL Java_com_google_oboe_sample_drumthumper_DrumPlayer_teardownAudioStreamNative(
-        JNIEnv* env, jobject, jint numSampleBuffers, jint numChannels, jint sampleRate) {
+JNIEXPORT void JNICALL Java_com_google_oboe_sample_drumthumper_DrumPlayer_teardownAudioStreamNative(JNIEnv* , jobject) {
     __android_log_print(ANDROID_LOG_INFO, TAG, "%s", "deinit()");
 
     // we know in this case that the sample buffers are all 1-channel, 44.1K
     sDTPlayer.teardownAudioStream();
+}
+
+/**
+ * Native (JNI) implementation of DrumPlayer.allocSampleDataNative()
+ */
+JNIEXPORT void JNICALL Java_com_google_oboe_sample_drumthumper_DrumPlayer_allocSampleDataNative(
+        JNIEnv* env, jobject, jint numSampleBuffers) {
+    __android_log_print(ANDROID_LOG_INFO, TAG, "%s", "allocSampleDataNative()");
+
+    // we know in this case that the sample buffers are all 1-channel, 44.1K
+    sDTPlayer.allocSampleData(numSampleBuffers);
 }
 
 /**
@@ -68,6 +78,13 @@ JNIEXPORT void JNICALL Java_com_google_oboe_sample_drumthumper_DrumPlayer_loadWa
     env->GetByteArrayRegion (bytearray, 0, len, reinterpret_cast<jbyte*>(buf));
     sDTPlayer.loadSampleDataFromAsset(buf, len, index);
     delete[] buf;
+}
+
+/**
+ * Native (JNI) implementation of DrumPlayer.nloadWavAssetsNative()
+ */
+JNIEXPORT void JNICALL Java_com_google_oboe_sample_drumthumper_DrumPlayer_unloadWavAssetsNative(JNIEnv* env, jobject) {
+    sDTPlayer.unloadSampleData();
 }
 
 /**
--- a/samples/drumthumper/src/main/cpp/DrumPlayerJNI.cpp
+++ b/samples/drumthumper/src/main/cpp/DrumPlayerJNI.cpp
@@ -81,7 +81,7 @@ JNIEXPORT void JNICALL Java_com_google_oboe_sample_drumthumper_DrumPlayer_loadWa
 }
 
 /**
- * Native (JNI) implementation of DrumPlayer.nloadWavAssetsNative()
+ * Native (JNI) implementation of DrumPlayer.unloadWavAssetsNative()
  */
 JNIEXPORT void JNICALL Java_com_google_oboe_sample_drumthumper_DrumPlayer_unloadWavAssetsNative(JNIEnv* env, jobject) {
     sDTPlayer.unloadSampleData();
--- a/samples/drumthumper/src/main/java/com/google/oboe/sample/drumthumper/DrumThumperActivity.kt
+++ b/samples/drumthumper/src/main/java/com/google/oboe/sample/drumthumper/DrumThumperActivity.kt
@@ -171,19 +171,18 @@ class DrumThumperActivity : AppCompatActivity(), TriggerPad.DrumPadTriggerListen
     }
 
     override fun onStop() {
-        super.onStop()
-
         if (mUseDeviceChangeFallback) {
             mAudioMgr!!.unregisterAudioDeviceCallback(mDeviceListener)
         }
 
         mDrumPlayer.teardownAudioStream()
+
+        super.onStop()
     }
 
     override fun onDestroy() {
-        super.onDestroy()
-
         mDrumPlayer.unloadWavAssets();
+        super.onDestroy()
     }
 
     //
--- a/samples/drumthumper/src/main/java/com/google/oboe/sample/drumthumper/DrumThumperActivity.kt
+++ b/samples/drumthumper/src/main/java/com/google/oboe/sample/drumthumper/DrumThumperActivity.kt
@@ -38,7 +38,7 @@ class DrumThumperActivity : AppCompatActivity(), TriggerPad.DrumPadTriggerListen
 
     private var mDrumPlayer = DrumPlayer()
 
-    private val mUseDeviceChangeFallback = true
+    private val mUseDeviceChangeFallback = false
     private val mSwitchTimerMs = 500L
 
     private var mDevicesInitialized = false
@@ -104,16 +104,17 @@ class DrumThumperActivity : AppCompatActivity(), TriggerPad.DrumPadTriggerListen
 
         mAudioMgr = getSystemService(Context.AUDIO_SERVICE) as AudioManager
 
-        if (mUseDeviceChangeFallback) {
-            mAudioMgr!!.registerAudioDeviceCallback(mDeviceListener, null)
-        }
-
-        mDrumPlayer.setupAudioStream()
+        mDrumPlayer.allocSampleData()
         mDrumPlayer.loadWavAssets(getAssets())
     }
 
     override fun onStart() {
         super.onStart()
+        if (mUseDeviceChangeFallback) {
+            mAudioMgr!!.registerAudioDeviceCallback(mDeviceListener, null)
+        }
+
+        mDrumPlayer.setupAudioStream()
     }
 
     override fun onResume() {
@@ -172,15 +173,17 @@ class DrumThumperActivity : AppCompatActivity(), TriggerPad.DrumPadTriggerListen
     override fun onStop() {
         super.onStop()
 
+        if (mUseDeviceChangeFallback) {
+            mAudioMgr!!.unregisterAudioDeviceCallback(mDeviceListener)
+        }
+
+        mDrumPlayer.teardownAudioStream()
     }
 
     override fun onDestroy() {
         super.onDestroy()
 
-        if (mUseDeviceChangeFallback) {
-            mAudioMgr!!.unregisterAudioDeviceCallback(mDeviceListener)
-        }
-        mDrumPlayer.teardownAudioStream()
+        mDrumPlayer.unloadWavAssets();
     }
 
     //
--- a/samples/drumthumper/src/main/java/com/google/oboe/sample/drumthumper/DrumThumperActivity.kt
+++ b/samples/drumthumper/src/main/java/com/google/oboe/sample/drumthumper/DrumThumperActivity.kt
@@ -110,11 +110,12 @@ class DrumThumperActivity : AppCompatActivity(), TriggerPad.DrumPadTriggerListen
 
     override fun onStart() {
         super.onStart()
+
+        mDrumPlayer.setupAudioStream()
+
         if (mUseDeviceChangeFallback) {
             mAudioMgr!!.registerAudioDeviceCallback(mDeviceListener, null)
         }
-
-        mDrumPlayer.setupAudioStream()
     }
 
     override fun onResume() {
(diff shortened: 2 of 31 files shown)


samples/iolib/src/main/cpp/player/SampleBuffer.h
```diff
@@ -17,44 +17,40 @@
...
+    virtual AudioProperties getProperties() const { return mAudioProperties; }
 
+    const float* getSampleData() { return mSampleData; }
+    int32_t getNumSampleFrames() { return mNumSampleFrames; }
 
 protected:
-    AudioProperties mProperties;
+    AudioProperties mAudioProperties;
 
     float*  mSampleData;
```
philburk: Consider using a std::shared_ptr<float[]> for mSampleData.

samples/iolib/src/main/cpp/player/SampleSource.cpp
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright (C) 2019 The Android Open Source Project
```
philburk: 2020?

samples/iolib/src/main/cpp/player/SampleSource.h
```diff
@@ -0,0 +1,61 @@
+/*
+ * Copyright (C) 2019 The Android Open Source Project
```
philburk: 2020?

samples/iolib/src/main/cpp/player/SimpleMultiPlayer.cpp
```diff
@@ -104,7 +106,14 @@ void SimpleMultiPlayer::setupAudioStream(int32_t numSampleBuffers, int32_t chann
     mChannelCount = channelCount;
     mSampleRate = sampleRate;
 
-    mSampleBuffers = new OneShotSampleBuffer[mNumSampleBuffers = numSampleBuffers];
+    mNumSampleBuffers = numSampleBuffers;
+    mSampleBuffers = new SampleBuffer*[mNumSampleBuffers];
```
philburk: Consider using std::unique_ptr<SampleBuffer>

samples/iolib/src/main/cpp/player/SimpleMultiPlayer.cpp
```diff
@@ -126,32 +135,36 @@ void SimpleMultiPlayer::loadSampleDataFromAsset(byte* dataBytes, int32_t dataLen
...
 
 void SimpleMultiPlayer::unloadSampleData() {
     for (int32_t bufferIndex = 0; bufferIndex < mNumSampleBuffers; bufferIndex++) {
-        mSampleBuffers[bufferIndex].unloadSampleData();
+        mSampleBuffers[bufferIndex]->unloadSampleData();
     }
-    delete[] mSampleBuffers;
-    mSampleBuffers = nullptr;
+
+    delete[] mSampleSources;
```
philburk: Where does mSampleBuffers get deleted?

samples/iolib/src/main/cpp/player/SampleSource.h
android-usb: Done.

samples/iolib/src/main/cpp/player/SimpleMultiPlayer.cpp
android-usb: Now, in the destructor.

samples/iolib/src/main/cpp/player/OneShotSampleSource.cpp
```diff
@@ -18,41 +18,45 @@
...
                          ? std::min(numFrames, numSampleFrames - mCurFrameIndex)
                          : 0;
 
     if (numWriteFrames != 0) {
         // Mix in the samples
         int32_t lastIndex = mCurFrameIndex + numWriteFrames;
+        const float* sampleData = mSampleBuffer->getSampleData();
         for(int32_t index = 0; index < numWriteFrames; index++) {
-            outBuff[index] += mSampleData[mCurFrameIndex++];
+            outBuff[index] += sampleData[mCurFrameIndex++];
```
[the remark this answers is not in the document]
android-usb: Everything about this app assumes mono. I will update to support stereo (and explicit handling of various channel formats) in a future version.

samples/iolib/src/main/cpp/player/SimpleMultiPlayer.cpp
android-usb: Done, although I feel it makes the sample less clear.

samples/iolib/src/main/cpp/player/SampleBuffer.cpp
```diff
@@ -18,24 +18,22 @@
...
+void SampleBuffer::loadSampleData(parselib::WavStreamReader* reader) {
     mAudioProperties.channelCount = reader->getNumChannels();
     mAudioProperties.sampleRate = reader->getSampleRate();
 
     reader->positionToAudio();
 
     mNumSampleFrames = reader->getNumSampleFrames() * reader->getNumChannels();
-    mSampleData = new float[mNumSampleFrames];
-    reader->getDataFloat(mSampleData, reader->getNumSampleFrames());
+    mSampleData = std::make_shared<float*>(new float[mNumSampleFrames]);
```
philburk: Should this be make_shared<float[]>(mNumSamples) ?

samples/iolib/src/main/cpp/player/SampleBuffer.cpp
```diff
@@ -18,24 +18,22 @@
...
 namespace iolib {
 
-void SampleBuffer::loadSampleData(WavStreamReader* reader) {
+void SampleBuffer::loadSampleData(parselib::WavStreamReader* reader) {
     mAudioProperties.channelCount = reader->getNumChannels();
     mAudioProperties.sampleRate = reader->getSampleRate();
 
     reader->positionToAudio();
 
     mNumSampleFrames = reader->getNumSampleFrames() * reader->getNumChannels();
```
philburk: This should be mNumSamples. It is not equal to numFrames except when mono.

samples/iolib/src/main/cpp/player/SampleBuffer.h
```diff
@@ -33,21 +31,22 @@ struct AudioProperties {
...
 
     // Data load/unload
-    void loadSampleData(WavStreamReader* reader);
+    void loadSampleData(parselib::WavStreamReader* reader);
     void unloadSampleData();
 
     virtual AudioProperties getProperties() const { return mAudioProperties; }
 
-    const float* getSampleData() { return mSampleData; }
+    std::shared_ptr<float*> getSampleData() { return mSampleData; }
```
philburk: float[]

samples/iolib/src/main/cpp/player/SimpleMultiPlayer.cpp
```diff
@@ -48,9 +50,10 @@ DataCallbackResult SimpleMultiPlayer::onAudioReady(AudioStream *oboeStream, void
 
     memset(audioData, 0, numFrames * sizeof(float));
 
+    OneShotSampleSource *** sources = mSampleSources.get();
```
philburk: Triple *!

samples/iolib/src/main/cpp/player/SampleBuffer.cpp
android-usb: I don't think so. The std::shared_ptr<> wraps a float*, not a float[]. Perhaps Mikhail can chime in on this?

samples/iolib/src/main/cpp/player/SampleBuffer.cpp
android-usb: The whole thing needs to be generalized for arbitrary channels. Still, done.

samples/iolib/src/main/cpp/player/SimpleMultiPlayer.cpp
android-usb: That seems to be how that unique_ptr<> works. Maybe Mikhail can provide some insight?

samples/iolib/src/main/cpp/player/SampleBuffer.cpp
mnaganov: Well, prior to C++17 dealing with shared_ptrs to arrays was tricky because there was no dedicated support for that (unlike what unique_ptr has), and the culprit is that if you allocate an array using 'new[]' operator, you **must** delete it with 'delete[]' operator (this is what unique_ptr does). So prior to C++17 you have to provide a custom deleter if you use shared_ptr with a C array.

Since C++17 this issue has been fixed in the language and now it's OK to use shared_ptr with C arrays, and even support for 'operator[]' to access array elements was added to shared_ptr. However, to use the proper deleter you must declare your shared_ptr as std::shared_ptr<float[]>, not <float*>.

See more explanations and examples here: https://stackoverflow.com/a/13062069/4477684

Phil, what C++ standard version does Oboe require? If it's C++17, you can go an easy way, otherwise there is more technical work to do.

samples/iolib/src/main/cpp/player/SimpleMultiPlayer.cpp
mnaganov: Paul, you are using an incorrect type for your unique_ptrs. If I understand it correctly, you have an array of pointers. For that, your unique_ptr must be declared as: std::unique_ptr<SampleBuffer*[]>. Please do not decay an array into a pointer because then unique_ptr will use 'delete' instead of 'delete[]' for an array, and that's a memory leak.

BTW, who manages these SampleBuffer instances that we are pointing to? Why it's OK to use raw pointers?

philburk: Would it be better to have a unique_ptr to an array of unique_ptrs to SampleBuffers?

samples/iolib/src/main/cpp/player/SimpleMultiPlayer.h
```diff
@@ -60,16 +58,16 @@ class SimpleMultiPlayer : public AudioStreamCallback  {
...
 
     // Audio attributs
     int32_t mChannelCount;
     int32_t mSampleRate;
 
     // Sample Data
     int32_t mNumSampleBuffers;
-    SampleBuffer**          mSampleBuffers;
-    OneShotSampleSource**   mSampleSources;
+    std::unique_ptr<SampleBuffer**> mSampleBuffers;
```
dturner, line 69: If you're just trying to define an array of `SampleBuffer`s you can write `std::unique_ptr<SampleBuffer[]>`

samples/iolib/src/main/cpp/player/SimpleMultiPlayer.h
```diff
@@ -60,16 +58,16 @@ class SimpleMultiPlayer : public AudioStreamCallback  {
...
+    std::unique_ptr<OneShotSampleSource**>   mSampleSources;
```
dturner, line 70: Same as above

samples/iolib/src/main/cpp/player/SimpleMultiPlayer.h
```diff
@@ -60,16 +58,16 @@ class SimpleMultiPlayer : public AudioStreamCallback  {
 
 private:
     // Oboe Audio Stream
-    AudioStream *mAudioStream { nullptr };
+    oboe::AudioStream *mAudioStream { nullptr };
```
dturner, line 61: You can use a `ManagedStream` here to avoid the raw pointer

samples/iolib/src/main/cpp/player/SimpleMultiPlayer.h
```diff
@@ -22,24 +22,22 @@
...
     // Inherited from oboe::AudioStreamCallback
-    DataCallbackResult onAudioReady(AudioStream *oboeStream, void *audioData,
+    oboe::DataCallbackResult onAudioReady(oboe::AudioStream *oboeStream, void *audioData,
             int32_t numFrames) override;
-    virtual void onErrorAfterClose(AudioStream *oboeStream, Result error) override;
-    virtual void onErrorBeforeClose(AudioStream * oboeStream, Result error) override;
+    virtual void onErrorAfterClose(oboe::AudioStream *oboeStream, oboe::Result error) override;
+    virtual void onErrorBeforeClose(oboe::AudioStream * oboeStream, oboe::Result error) override;
 
     void setupAudioStream(int32_t numSampleBuffers, int32_t channelCount, int32_t sampleRate);
```
dturner, line 42: Does `SimpleMultiPlayer` make sense without an audio stream? If not, then move these parameters into the constructor and initialize them there.

samples/iolib/src/main/cpp/player/SampleBuffer.h
dturner: +1

samples/iolib/src/main/cpp/player/SampleBuffer.cpp
dturner: Why is this a `shared_ptr` at all? Isn't `SampleBuffer` the sole owner of the sample data? If so, you can change to `make_unique<float[]>(mNumSamples)`.

android-usb: I'm going to put this off until after this merge. I will seek more guidance from the template gods after that.

samples/iolib/src/main/cpp/player/SimpleMultiPlayer.cpp
android-usb: I will revisit this.

samples/iolib/src/main/cpp/player/SimpleMultiPlayer.h
android-usb: Made this std::vector<SampleBuffer*> as per our discussion.

samples/iolib/src/main/cpp/player/SimpleMultiPlayer.h
android-usb: It does because the stream will get allocated/deallocated (created/destroyed) separately from the SampleBuffer/Sample source pairs based on application life-cycle.

samples/iolib/src/main/cpp/player/SimpleMultiPlayer.h
android-usb: Made this std::vector<OneShotSampleSource*> as per our discussion.

samples/iolib/src/main/cpp/player/SimpleMultiPlayer.h
android-usb: Done. (I think)

samples/iolib/src/main/cpp/player/SampleBuffer.h
android-usb: I haven't been able to get the compiler to accept that.

samples/drumthumper/src/main/cpp/DrumPlayerJNI.cpp
```diff
@@ -70,6 +80,13 @@ JNIEXPORT void JNICALL Java_com_google_oboe_sample_drumthumper_DrumPlayer_loadWa
     delete[] buf;
 }
 
+/**
+ * Native (JNI) implementation of DrumPlayer.nloadWavAssetsNative()
```
philburk, line 84: c/nload/unload/

samples/drumthumper/src/main/java/com/google/oboe/sample/drumthumper/DrumThumperActivity.kt
```diff
@@ -142,23 +164,26 @@ class DrumThumperActivity : AppCompatActivity(), TriggerPad.DrumPadTriggerListen
...
     override fun onPause() {
         super.onPause()
-
-        mAudioMgr!!.unregisterAudioDeviceCallback(mDeviceListener)
     }
 
     override fun onStop() {
-        mDrumPlayer.teardownAudioStream()
         super.onStop()
+
```
philburk, line 175: the onStop() sequence should be in reverse order to the onStart() sequence.
eg. call super.onStop() at the end.

samples/drumthumper/src/main/java/com/google/oboe/sample/drumthumper/DrumThumperActivity.kt
```diff
@@ -142,23 +164,26 @@ class DrumThumperActivity : AppCompatActivity(), TriggerPad.DrumPadTriggerListen
...
+            mAudioMgr!!.unregisterAudioDeviceCallback(mDeviceListener)
+        }
+
+        mDrumPlayer.teardownAudioStream()
     }
 
     override fun onDestroy() {
         super.onDestroy()
+
+        mDrumPlayer.unloadWavAssets();
```
philburk, line 186: move before super.onDestroy()_

samples/iolib/src/main/cpp/player/DataSource.h
```diff
@@ -19,22 +19,18 @@
...
 
 /*
  * Defines an interface for audio data sources for the SimpleMultiPlayer class.
  */
 class DataSource {
 public:
     virtual ~DataSource() {};
-    virtual AudioProperties getProperties() const  = 0;
+
+    virtual void mixAudio(float* outBuff, int numFrames) =0;
```
philburk, line 31: = 0

samples/drumthumper/src/main/java/com/google/oboe/sample/drumthumper/DrumThumperActivity.kt
```diff
@@ -104,16 +104,17 @@ class DrumThumperActivity : AppCompatActivity(), TriggerPad.DrumPadTriggerListen
...
-        }
-
-        mDrumPlayer.setupAudioStream()
+        mDrumPlayer.allocSampleData()
         mDrumPlayer.loadWavAssets(getAssets())
     }
 
     override fun onStart() {
         super.onStart()
+        if (mUseDeviceChangeFallback) {
```
philburk, line 113: The start and stop are still not symmetric.  Is ABC-BCA, should probably be ABC-CBA
Can there be a problem if the device callback occurs before the audio is setup?
Or vice versa.

samples/iolib/src/main/cpp/player/SimpleMultiPlayer.cpp
```diff
@@ -104,73 +104,73 @@ bool SimpleMultiPlayer::openStream() {
...
     for (int32_t bufferIndex = 0; bufferIndex < mNumSampleBuffers; bufferIndex++) {
-        (*buffers)[bufferIndex]->unloadSampleData();
+        mSampleBuffers[bufferIndex]->unloadSampleData();
+        delete mSampleBuffers[bufferIndex];
+        delete mSampleSources[bufferIndex];
     }
 
-    mSampleSources.reset();
-    mSampleSources.reset();
+    mSampleSources.clear();
```
philburk, line 153: Should one of these be mSampleBuffers?

android-usb: ACTUALLY done now.

PR #800, opened 2020-03-02
Add steps for using prefab package. Fixes #230

### LINKED ISSUE #230: Provide Oboe binary distribution
Discussion started on https://github.com/felipecsl/ktnes/issues/4

Instead of having users clone Oboe in order to use it, It would be desirable to have Oboe distributed as an `aar` package for Android app consumption, since that supports including the precompiled `libOboe.so` files for each supported native platform along with header files.

dturner:


docs/GettingStarted.md
```diff
@@ -1,12 +1,36 @@
...
+## Option 1) Using pre-built binaries and headers
+Oboe is distributed as a [prefab](https://github.com/google/prefab) package via [Google Maven](https://maven.google.com/web/index.html) (search for "oboe"). [Prefab support was added](https://android-developers.googleblog.com/2020/02/native-dependencies-in-android-studio-40.html) to [Android Studio Preview 4.0 Canary 9](https://developer.android.com/studio/preview) so you'll need to be using this version of Android Studio or above. 
+
+Add the oboe dependency to your app's `build.gradle`. Replace "1.3.0" with the [latest stable version](https://github.com/google/oboe/releases/) of Oboe: 
+
+    dependencies {
+        implementation 'com.google.oboe:oboe:1.3.0'
+    }
+
+Prefab isn't enabled by default (this may change in future versions of Android Studio) so enable it by adding following to your `gradle.properties`:
```
philburk: ...file, which is int the top folder of your app.

docs/GettingStarted.md
```diff
@@ -1,12 +1,36 @@
...
+
+    dependencies {
+        implementation 'com.google.oboe:oboe:1.3.0'
+    }
+
+Prefab isn't enabled by default (this may change in future versions of Android Studio) so enable it by adding following to your `gradle.properties`: 
+
+    # Enables Prefab
+    android.enablePrefab=true
+    # Work around https://issuetracker.google.com/149575364
```
philburk: Given the number of workarounds, we may want to warn developers up top that this is new and experimental and ask for feedback. Also to check back soon for updated instructions. We need to figure out a way to tell developers when to remove the workarounds in "gradle.properties" or they will be there forever.

docs/GettingStarted.md
```diff
@@ -1,12 +1,36 @@
...
+    android.enableParallelJsonGen=false
+    # 4.0.0 canary 9 defaults to Prefab 1.0.0-alpha3, which is not the latest.
+    android.prefabVersion=1.0.0-alpha5
+
+Include and link to oboe by updating your `CMakeLists.txt`: 
+
+    find_package (oboe REQUIRED CONFIG)
+    target_link_libraries(app oboe::oboe)
```
philburk: Add to list of possibly many link libraries.

docs/GettingStarted.md
dturner: Good point. Text updated.

Issue #796, opened 2020-02-28
App crashes randomly in LiveEffects sample app


mahaju:
Android Version(s): 8.0,10
Android Device(s): Samsung Galaxy S7 (android 8), Samsung Galaxy S10 (android 10)
Oboe Version: 

**Short Description**
App sometimes crashes at random times

**Steps To Reproduce**
Turn audio output on and off (Start button press), alternate between AAudio and OpenSL ES, and various mic sources

**Expected behavior**
Audio output should stop when pressing the stop button, start when pressing the start button

**Actual behavior**
App sometimes crashes, sometimes when pressing the start/stop button, sometimes while pressing the radio buttons to alternate between AAudio and OpenSL ES. Very seldom it has crashed while just letting the app play audio normally, but even in those cases, start/stop button had been already pressed multiple times, and AAudio/OpenSL ES have been alternated between.

**Additional context**
Almost all of the crashes have happened when going from no audio output state to audio output state, that is, when pressing the button to start audio, not when pressing the button to stop audio. Program crashes in Samsung Galaxy S7 and S10, but has caused the S10 to actually restart a few times.

Here is a logcat for one of the crashes:

```
    --------- beginning of crash
2020-02-28 12:07:26.600 16997-17135/com.google.sample.oboe.liveeffect A/libc: Fatal signal 11 (SIGSEGV), code 1, fault addr 0xffffffffffffffe8 in tid 17135 (Binder:16997_5)
2020-02-28 12:07:26.663 17148-17148/? I/crash_dump64: obtaining output fd from tombstoned
2020-02-28 12:07:26.664 3342-3342/? I//system/bin/tombstoned: received crash request for pid 16997
2020-02-28 12:07:26.666 17148-17148/? I/crash_dump64: performing dump of process 16997 (target tid = 17135)
2020-02-28 12:07:26.666 17148-17148/? A/DEBUG: *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
2020-02-28 12:07:26.666 17148-17148/? A/DEBUG: Build fingerprint: 'samsung/heroltektt/heroltektt:8.0.0/R16NW/G930KKKU2ESF1:user/release-keys'
2020-02-28 12:07:26.666 17148-17148/? A/DEBUG: Revision: '8'
2020-02-28 12:07:26.666 17148-17148/? A/DEBUG: ABI: 'arm64'
[...]

PR #806, opened 2020-03-12
LiveEffect: handle failure to open the stream

philburk:
Used to crash on a nullptr.
Now it passes the result up to Java.
Also uses ManagedStream to prevent leaked resources.

Bug: #796

### DIFF
--- a/samples/LiveEffect/src/main/cpp/LiveEffectEngine.cpp
+++ b/samples/LiveEffect/src/main/cpp/LiveEffectEngine.cpp
@@ -18,7 +18,6 @@
 #include <assert.h>
 #include <logging_macros.h>
 
-
 LiveEffectEngine::LiveEffectEngine() {
     assert(mOutputChannelCount == mInputChannelCount);
 }
@@ -28,7 +27,6 @@ LiveEffectEngine::~LiveEffectEngine() {
     mFullDuplexPass.stop();
     closeStream(mPlayStream);
     closeStream(mRecordingStream);
-
 }
 
 void LiveEffectEngine::setRecordingDeviceId(int32_t deviceId) {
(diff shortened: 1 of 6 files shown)


samples/LiveEffect/src/main/cpp/LiveEffectEngine.cpp
```diff
@@ -43,19 +41,25 @@ bool LiveEffectEngine::isAAudioSupported() {
...
         if (isOn) {
-            openStreams();
-            mFullDuplexPass.start();
+            success = openStreams() == oboe::Result::OK;
+            if (success) {
+                mFullDuplexPass.start();
+                mIsEffectOn = isOn;
+            }
         } else {
+            mIsEffectOn = isOn;
```
ggfan: nit:  if possible, move this one down to the last one: after closeStream(mRecordingStream);? ( only change the state after   everything is done )

samples/LiveEffect/src/main/cpp/LiveEffectEngine.cpp
```diff
@@ -18,7 +18,6 @@
 #include <assert.h>
```
ggfan: nit: include <cassert>

samples/LiveEffect/src/main/cpp/jni_bridge.cpp
```diff
@@ -42,17 +42,17 @@ Java_com_google_sample_oboe_liveEffect_LiveEffectEngine_delete(JNIEnv *env,
...
     if (engine == nullptr) {
         LOGE(
             "Engine is null, you must call createEngine before calling this "
             "method");
-        return;
+        return JNI_FALSE;
     }
 
-    engine->setEffectOn(isEffectOn);
+    return (jboolean) engine->setEffectOn(isEffectOn);
```
ggfan: nit: explicitly translate it from true/false to JNI_TRUE/JNI_FALSE ( kindly avoid implicitly depending JNI_TRUE was defined to be 1 )

samples/LiveEffect/src/main/cpp/LiveEffectEngine.cpp
philburk: I changed it to <cassert>

ggfan:
LGTM!

PR #821, opened 2020-04-15
Add openSharedStream() to prevent deletion while executing onError callbacks

philburk:


include/oboe/AudioStream.h
```diff
@@ -418,6 +418,22 @@ class AudioStream : public AudioStreamBase {
     ResultWithValue<int32_t> waitForAvailableFrames(int32_t numFrames,
                                                     int64_t timeoutNanoseconds);
 
+    /*
```
dturner, line 421: Potentially use friends to avoid exposing these methods in public API

src/common/AudioStreamBuilder.cpp
```diff
@@ -186,4 +186,14 @@ Result AudioStreamBuilder::openManagedStream(oboe::ManagedStream &stream) {
...
+    sharedStream.reset();
+    AudioStream *streamptr;
+    auto result = openStream(&streamptr);
+    sharedStream.reset(streamptr);
+    // Save a weak_ptr in the stream for use with callbacks.
+    streamptr->setWeakThis(sharedStream);
+    return result;
+}
+
 } // namespace oboe
```
philburk, line 199: add blank line

apps/OboeTester/app/src/main/cpp/OboeStreamCallbackProxy.cpp
```diff
@@ -85,6 +85,8 @@ void OboeStreamCallbackProxy::onErrorBeforeClose(oboe::AudioStream *audioStream,
     if (mCallback != nullptr) {
         mCallback->onErrorBeforeClose(audioStream, error);
     }
+    usleep(2000 * 1000); // FIXME - sleep to provoke a race condition
```
philburk, line 88: remove before release

include/oboe/AudioStream.h
dturner: Or just move to `AudioStreamBase`

philburk: AudioStreamBuilder cannot access a protected member of a superclass on another object!
So I had to use a friend class.

src/common/AudioStreamBuilder.cpp
```diff
@@ -186,4 +186,31 @@ Result AudioStreamBuilder::openManagedStream(oboe::ManagedStream &stream) {
...
+    auto result = openStream(&streamptr);
+    if (result == Result::OK) {
+        sharedStream.reset(streamptr);
+        // Save a weak_ptr in the stream for use with callbacks.
+        streamptr->setWeakThis(sharedStream);
+    }
+    return result;
+}
+
+ResultWithValue<std::shared_ptr<AudioStream>> AudioStreamBuilder::openSharedStream() {
```
dturner, line 202: The benefit of this method is that it fully encapsulates stream creation, rather than requiring the caller to create an empty `shared_ptr<AudioStream>` and pass it in as a parameter. It does break from convention though. Slight preference for this method, although happy to go with either one.

include/oboe/AudioStreamBuilder.h
```diff
@@ -399,6 +401,10 @@ class AudioStreamBuilder : public AudioStreamBase {
      */
     Result openManagedStream(ManagedStream &stream);
 
+    Result openSharedStream(std::shared_ptr<oboe::AudioStream> &sharedStream);
+
+    ResultWithValue<std::shared_ptr<oboe::AudioStream>> openSharedStream();
```
dturner, line 406: I would advocate for changing the method name to `openStream`. 

We already have the concept of a "shared audio stream" which means `sharingMode=shared` so `openSharedStream` is likely to cause confusion with users thinking "but I want an EXCLUSIVE stream, not a SHARED stream?!" or similar.

src/common/AudioStreamBuilder.cpp
dturner: Let me rescind my "slight preference for `ResultWithValue` method" and replace it with "strong preference for the original `Result` method". All of our existing users (including all our sample code) will be using the "declare and pass stream pointer as parameter" approach. The former method makes it a very easy refactor, the latter does not.

include/oboe/AudioStreamBuilder.h
philburk: Good point. We still have a "shared_ptr" but I think that is less confusing. One can have a shared_ptr to an eXCLUSIVE stream. I will make these changes.

PR #836, opened 2020-04-27
Default to building both x86 and ARM ABIs

emilieroberts:
Default to building: arm32, arm64, x86_32, and x86_64 ABIs. Handy
for testing intel-based devices like Chromebooks.

### DIFF
--- a/apps/OboeTester/app/build.gradle
+++ b/apps/OboeTester/app/build.gradle
@@ -13,8 +13,7 @@ android {
         externalNativeBuild {
             cmake {
                 cppFlags "-std=c++14"
-                // abiFilters "x86", "armeabi-v7a", "arm64-v8a"
-                // abiFilters "arm64-v8a"
+                abiFilters "x86", "x86_64", "armeabi-v7a", "arm64-v8a"
             }
         }
     }


emilieroberts:
On my 8-core i7-8650U, this increased build time from 39s to 43s.

ggfan:
By default, these 4 ABIs should all built

To speed up local build, there is an option that inside IDE:
  - "Build Variants" (on the left, lower edge of IDE )
  - "Active ABI", choose one of the 4 ABIs: this will only build one ABI and more effective to shorten the build time.

may you try that instead?

emilieroberts:
I personally think for a project of this size (pretty small) it makes sense
to include all 4.

I've wasted tons of time building, copying, and pushing the apk only to
realize I forgot to modify the build.gradle file and don't have the right
ABIs (so yes, this is a bit of a personal convenience).

I also think it's a better base for anyone who uses OboeTester for the base
of their own Oboe apps. Ie. a small increase in build time in order to
support more users is the right default.

Of course, it's not very important so wouldn't insist on it.



Il mer 29 apr 2020, 20:12 Gerry <notifications@github.com> ha scritto:

philburk:
The issue is not the build time increase. As you point out, that is minor.

The issue is that the original code should have built binaries for  x86_64 by default because the ABI filters were commented out. So this change should not be needed. If it is then maybe there is a build bug that we need to understand.

philburk:
@emilieroberts - Did you see my comment from May 2? Shall we close this?

emilieroberts:
Sorry for delay. I believe Android studio defaults to only building ARM64
abis if others are not specifically requested.

This is perhaps a bug we should file against Android Studio (default should
be inclusive, not exclusive).

Let's close this as it's been addressed. Thanks!

Il mar 12 mag 2020, 02:33 Phil Burk <notifications@github.com> ha scritto:

PR #837, opened 2020-04-28
Fixed crash (for double-delete) in unloadSampleData().

android-usb:


samples/iolib/src/main/cpp/player/SampleBuffer.cpp
```diff
@@ -34,7 +34,9 @@ void SampleBuffer::loadSampleData(parselib::WavStreamReader* reader) {
 }
 
 void SampleBuffer::unloadSampleData() {
-    delete[] mSampleData;
+    if (mSampleData != nullptr) {
+        delete[] mSampleData;
```
philburk, line 38: It would be nice to clear the pointer in case this gets called again.

    mSampleData = nullptr;

android-usb: Done. Thnaks.

PR #842, opened 2020-05-02
Re-create hello oboe sample app to use prefab

nisrulz:
I was trying to play around with samples in the repository. To get everything functional with prefab setup, had to do some changes. Figured I could open a PR 😅  

- Made a copy of hello oboe sample app and updated the copy to use kotlin (only conversion + kotlin-fied)
- Using prefab to get oboe library
- Updated dependencies
- Updated AGP to v4.0.0-beta05
- Updated gradle wrapper to v6.3
- Updated kotlin to v1.3.72
- Some cleanup in the project regarding usage of `ndk` block when cmake was already being used
- Setup ndkVersion in build.gradle to get everything working properly
- All sample apps were tested to compile and run ✅

Let me know if anything needs to be reverted 🙇🏻‍♂️


samples/LiveEffect/build.gradle
```diff
@@ -1,17 +1,16 @@
...
     defaultConfig {
         applicationId 'com.google.sample.oboe.liveeffect'
         minSdkVersion 16
-        targetSdkVersion 28
+        targetSdkVersion 29
         versionCode 1
         versionName '1.0'
-        ndk {
-            abiFilters 'armeabi-v7a', 'arm64-v8a', 'x86', 'x86_64'
-        }
```
nisrulz, line 14: This looked redundant, so removed it.

samples/build.gradle
```diff
@@ -16,21 +16,18 @@
...
-    ext.kotlin_version = '1.3.50'
+    ext.kotlin_version = '1.3.72'
 
     repositories {
         google()
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.6.3'
+        classpath 'com.android.tools.build:gradle:4.0.0-beta05'
```
nisrulz, line 27: In order to use prefab, needed to upgrade AGP.

nisrulz:
@philburk I will wait for the review then. I need to know what I need to revert or split out into separate PRs.

The PR in ndk samples looks good too. I had no idea samples related to oboe were also under ndk-samples repo. I thought all of them were now ported to oboe repo under samples dir 😅

philburk:
I went through the code and unfortunately there is very little we can use.
The hello-oboe-with-prefab is not something we will use in this repository because this repo is connected to the source for Oboe. So these samples would not use prefab. That will only be in the NDK samples.

There are some gradle version updates but that should be a different CL. I don't want to discourage you from submitting changes. But a much smaller change would be easier to accept.

dturner:
Thanks very much for submitting this. The way forward is probably this: 

- replace existing hello-oboe with this kotlin version
- include prefab as an option which is disabled by default

Leaving open whilst we decide how and who will do this work....

nisrulz:
@dturner I would say that the java version is kept. Java is a valid language and some people want to use that. Having the sample app in that uses that should be considered.

Regarding your comment:
If we did that, it will still be a big change as @philburk mentioned in his comment. 

So would you suggest I split my PR into smaller parts so it is easier to review and merge? If yes, could you maybe give me a direction towards how the split should happen?

I would like to not throw away this PR, since I did struggle to get it to work initially and I believe others would too unless they see the sample app 😄

ggfan:
@nisrulz:
May you have a look at the [ndk-sample's version](https://github.com/android/ndk-samples/pull/714) of hello-oboe and comment there?

This repo is specialized in Audio processing in C/C++, probably we could leave the java side of the code as-is for now; however, on ndk-samples repo,  we do have intention to move to Kotlin.  Wondering you could help to convert [ndk-samples](https://github.com/android/ndk-samples)' code into Kotlin first, and then we go from there?

nisrulz:
I will take a look in that repo 👍

nisrulz:
@ggfan I took a look at the sample app in ndk-samples repo. 

I have opened a PR for updating hello-oboe sample app: https://github.com/android/ndk-samples/pull/731

That said, I think this PR in itself is very different from the sample app in ndk-sample. I am didnot remove the java sample in this PR rather created a new one with kotlin (using prefab setup for oboe).

If the PR was reviewed by commit, it would make more sense. I put clear commit messages to denote the change I made in a commit. 

Let me know still is not useful. I will close this PR then. 😄

nisrulz:
Hi, any updates on my last comment? 🤔

PR #845, opened 2020-05-06
OboeTester: cleanup debug messages

philburk:


philburk:
@ggfan - why is this failing Travis build? lldb 3.1 missing ?!
This seems unrelated to the minor code change.

philburk:
The Travis failure is because of an issue fixed in #851

PR #864, opened 2020-05-21
corrected intent constant

rpattabi:
The [documentation](https://developer.android.com/reference/android/content/Intent#ACTION_HEADSET_PLUG) of `Intent.ACTION_HEADSET_PLUG` recommends to use `AudioManager.ACTION_HEADSET_PLUG` instead if min sdk _is_ LOLLIPOP. It is unclear if it means min sdk is _at least_ LOLLIPOP. It may be worth mentioning it here.

### DIFF
--- a/docs/notes/disconnect.md
+++ b/docs/notes/disconnect.md
@@ -26,7 +26,7 @@ know that the device has changed. There is a "Test Disconnects" option in
 [OboeTester](https://github.com/google/oboe/tree/master/apps/OboeTester/docs)
 that can be used to diagnose this problem.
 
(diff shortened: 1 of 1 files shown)


rpattabi:
signed the CLA

philburk:
Thanks for the fix. AudioManager.ACTION_HEADSET_PLUG was added in Lollipop.

philburk:
I added a [note](https://github.com/google/oboe/blob/master/docs/notes/disconnect.md#workaround-for-not-disconnecting-properly) about LOLLIPOP as you suggested.

PR #868, opened 2020-06-01
Add script to create prefab package and maven artifact

dturner:
Packaging and changing all the prefab config files was taking a long time so I've created a script to do it. Should save quite a bit of time for future releases.

### DIFF
--- a/prefab_build.sh
+++ b/prefab_build.sh
@@ -1,3 +1,5 @@
+#!/bin/bash
+
 # Test that the prefab binary exists
 if hash prefab 2>/dev/null; then
   echo "Prefab is installed"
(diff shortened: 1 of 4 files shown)


prefab_build.sh
```diff
@@ -0,0 +1,88 @@
+# Test that the prefab binary exists
```
philburk, line 1: Add at top
#!/bin/bash

prefab_build.sh
```diff
@@ -0,0 +1,88 @@
...
+if hash prefab 2>/dev/null; then
+  echo "Prefab is installed"
+else
+  echo "Prefab binary not found. See https://github.com/google/prefab for install instructions"
+  exit 1
+fi
+
+# Get the version string from the source
+major=`grep "#define OBOE_VERSION_MAJOR" include/oboe/Version.h | cut -d' ' -f3`
```
philburk, line 10: Nice!
BTW, Mikhail says the back tick is obsolete and recommends:

    major=$(grep "#define OBOE_VERSION_MAJOR" include/oboe/Version.h | cut -d' ' -f3)

prefab_build.sh
```diff
@@ -0,0 +1,88 @@
...
+    echo "Copying the ${abi} library"
+    cp -v "../${abi}/liboboe.so" "oboe-${version}/prefab/modules/oboe/libs/android.${abi}/"
+  done
+
+  # Verify the prefab packages
+  for abi in ${ABIS[@]}
+  do
+
+    prefab --build-system cmake --platform android --os-version 29 \
+    --stl c++_shared --ndk-version 21 --abi ${abi} \
```
philburk, line 48: indent

prefab_build.sh
dturner: Good to know - thanks!

prefab_build.sh
dturner: Not sure what you mean? It is indented already using 2 spaces.

philburk: In android style, continuation lines are usually indented by 2 levels. That would be an extra 4 spaces in this case relative to line 47.

dturner: I getcha. Thanks for the explanation - I wasn't aware of the rule. Fixed.

PR #877, opened 2020-06-09
Handle EAGAIN result from decoder

dturner:


dturner:
Tested successfully using MP3 supplied from Ajit.

PR #890, opened 2020-06-12
oboe: add Quirk for broken mono on S9

### LINKED ISSUE #824: Audio input pitched down 1 octave on Samsung S9, broken mono
Android Version(s): 10
Android Device(s): Samsung S9 (SM-G960F)
Oboe Version: 1.3.3 and 1.2.4

**Short Description**
Using Oboe in full-duplex, the sound from the microphone sounds pitched down by one octave. This is on a stock version of android 10 (i.e. samsung's official version), on a newly wiped phone. 

**Steps To Reproduce**
Compile and run the FXLab demo app, and press the speaker button in the top right to unmute (happens both with headphones plugged in and without)

**Expected behavior**
Should pass through the signal normally.

**Actual behavior**
The sound sounds as if it is pitched down one octave, additionally, there are discontinuities in the waveform recorded at 3072 frame intervals (buffer size is 96 frames, so that's 32 buffers). Also, very suspiciously, the input buffer is a series of doubled up numbers, as if it's a stereo signal consisting of a mono signal interleaved with itself.

e.g.

-0.062038030475378
-0.062038030475378
-0.055203106254339
-0.055203106254339
-0.052048526704311
-0.052048526704311

I'm guessing the discontinuity is happening because the input buffer hasn't finished being pulled through the app yet (because it's playing at half speed) before it gets overwritten by incoming audio from the system

**Additional context**

(Sound file attached)

I'm happy to keep testing this and other stuff on the S9 if I can be of any help

philburk:
The Samsung S9 does not support mono properly.
It is actually stereo. So on that device, we open a stereo
stream and then convert it to mono by extracting
the first channel.

Fixes #824

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,6 +30,7 @@ set (oboe_sources
     src/flowgraph/ClipToRange.cpp
     src/flowgraph/ManyToMultiConverter.cpp
     src/flowgraph/MonoToMultiConverter.cpp
+    src/flowgraph/MultiToMonoConverter.cpp
     src/flowgraph/RampLinear.cpp
     src/flowgraph/SampleRateConverter.cpp
     src/flowgraph/SinkFloat.cpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,7 +26,8 @@ set (oboe_sources
     src/fifo/FifoController.cpp
     src/fifo/FifoControllerBase.cpp
     src/fifo/FifoControllerIndirect.cpp
-        src/flowgraph/FlowGraphNode.cpp
+    src/flowgraph/FlowGraphNode.cpp
+    src/flowgraph/ChannelCountConverter.cpp
     src/flowgraph/ClipToRange.cpp
     src/flowgraph/ManyToMultiConverter.cpp
     src/flowgraph/MonoToMultiConverter.cpp
--- a/src/common/DataConversionFlowGraph.cpp
+++ b/src/common/DataConversionFlowGraph.cpp
@@ -23,6 +23,7 @@
 
 #include <flowgraph/ClipToRange.h>
 #include <flowgraph/MonoToMultiConverter.h>
+#include <flowgraph/MultiToMonoConverter.h>
 #include <flowgraph/RampLinear.h>
 #include <flowgraph/SinkFloat.h>
 #include <flowgraph/SinkI16.h>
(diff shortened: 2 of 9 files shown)


src/common/QuirksManager.cpp
```diff
@@ -51,11 +52,24 @@ int32_t QuirksManager::DeviceQuirks::clipBufferSize(AudioStream &stream,
...
 }
 
+bool QuirksManager::DeviceQuirks::isAAudioMMapPossible(const AudioStreamBuilder &builder) {
+    bool isSampleRateCompatible =
+            builder.getSampleRate() == oboe::Unspecified
+            || builder.getSampleRate() == kCommonNativeRate
+            || builder.getSampleRateConversionQuality() != SampleRateConversionQuality::None;
+    return builder.getPerformanceMode() == PerformanceMode::LowLatency
+            && isSampleRateCompatible
+            && builder.getChannelCount() <= 2;
```
philburk, line 62: From b/155341396
> When we select Camcorder , it works well because mmap open fails and it works as legacy path.

CMakeLists.txt
```diff
@@ -30,6 +30,7 @@ set (oboe_sources
     src/flowgraph/ClipToRange.cpp
     src/flowgraph/ManyToMultiConverter.cpp
     src/flowgraph/MonoToMultiConverter.cpp
+    src/flowgraph/MultiToMonoConverter.cpp
```
dturner, line 33: Couldn't both `MultiToMono` and `MonoToMulti` be contained within single `ChannelCountConverter` which takes an arbitrary number of input and output channels?

src/common/DataConversionFlowGraph.cpp
```diff
@@ -139,27 +140,43 @@ Result DataConversionFlowGraph::configure(AudioStream *sourceStream, AudioStream
         lastOutput = &mSource->output;
     }
 
+    if (sourceChannelCount != sinkChannelCount
+        && sourceChannelCount != 1
+        && sinkChannelCount != 1
+        ) {
+        LOGW("%s() Channel conversion %d to %d not supported.", __func__,
```
dturner, line 147: `LOGE` since this is an error?

src/common/QuirksManager.cpp
```diff
@@ -115,24 +141,39 @@ bool QuirksManager::isConversionNeeded(
...
-            // b/66967812 AudioRecord does not allow FAST track for stereo capture in O
-            childBuilder.setChannelCount(1);
-            conversionNeeded = true;
-        }
-        // Note that MMAP does not support mono in 8.1. But that would only matter on Pixel 1
-        // phones and they have almost all been updated to 9.0.
+    // Channel Count conversions
+    if (OboeGlobals::areWorkaroundsEnabled()
+            && builder.isChannelConversionAllowed()
+            && builder.getChannelCount() == 2 // stereo?
```
dturner, line 150: Could these be changed to constants? I think we have `kChannelCount*` consts

src/common/QuirksManager.cpp
```diff
@@ -115,24 +141,39 @@ bool QuirksManager::isConversionNeeded(
...
+            && isInput
+            && isLowLatency
+            && (!builder.willUseAAudio() && (getSdkVersion() == __ANDROID_API_O__))
+            ) {
+        // Workaround for heap size regression in O.
+        // b/66967812 AudioRecord does not allow FAST track for stereo capture in O
+        childBuilder.setChannelCount(1);
```
dturner, line 157: Constant?

src/common/QuirksManager.cpp
```diff
@@ -115,24 +141,39 @@ bool QuirksManager::isConversionNeeded(
...
+        conversionNeeded = true;
+        LOGI("QuirksManager::%s() using mono internally for low latency on O", __func__);
+    } else if (OboeGlobals::areWorkaroundsEnabled()
+               && builder.getChannelCount() == 1 // mono?
```
dturner, line 161: Constant?

src/common/QuirksManager.cpp
```diff
@@ -115,24 +141,39 @@ bool QuirksManager::isConversionNeeded(
...
+               && isInput
+               && mDeviceQuirks->isMonoMMapActuallyStereo()
+               && builder.willUseAAudio()
+               // Note: we might use this workaround on a device that supports
+               // MMAP but will use Legacy for this stream.  But this will only happen
+               // on devices that have the broken mono.
+               && mDeviceQuirks->isAAudioMMapPossible(builder)
+               ) {
+        // Workaround for mono actually running in stereo mode.
+        childBuilder.setChannelCount(2); // Use stereo and extract first channel.
```
dturner, line 171: Constant

src/common/DataConversionFlowGraph.cpp
philburk: removed

CMakeLists.txt
philburk: I added a ChannelCountConverter for the previous unsupported case, eg "2:3". But it looks a bit more complex and is slower than the Mono nodes. So I kept them.

src/common/QuirksManager.h
```diff
@@ -110,6 +110,9 @@ class QuirksManager {
 
 private:
 
+    static constexpr int32_t kChannelCountMono = 1;
```
dturner, line 113: Rather than add these I think you can use the `oboe::ChannelCount::*` constants

Issue #874, opened 2020-06-09
LiveEffect Demo does not work with aaudio and Bluetooth headset


chrbsg:
Android version(s):10
Android device(s):Pixel 2 XL
Oboe version:master
App name used for testing (please try to reproduce the issue using the Oboe samples and apps): LIveEffect

**Short description**
Bluetooth headset microphone does not appear to work with aaudio (but works with OpenSL ES)

**Steps to reproduce**
Run LiveEffect on Pixel 2 XL
Select PLT_M70 as both record device and playback device
Select AAudio

**Expected behavior**
Demo should work

**Actual behavior**
No audio

**Any additional context**
Demo works ok under exactly same conditions if "OpenSL ES" is selected. 

Possibly related: #155

philburk:
It does seem related to https://github.com/google/oboe/issues/155#issuecomment-491319585

Should we call am.startBluetoothSco() just in case someone is using a bluetooth headset?

chrbsg:
Does the bluetooth microphone work? It only appears as a Bluetooth telephony device in "Recording device", and does not work.
imho if the dropdowns offer bluetooth devices for selection, then selecting those devices should work. If it is not possible to have a generic initialisation of bluetooth devices, then perhaps the appropriate device-specific init calls can be made after the user performs the device selection, or clicks "start"?

dturner:
@philburk is just waiting on a BT headset - then will investigate further.

philburk:
I got a NewBee BT headset and reproduced the problems you were seeing.

I modified OboeTester so that it starts and stops SCO as needed. See #896 for code you can borrow.

@dturner  - Do you want to make similar changes in Hello-Oboe and LiveEffect?

robertwu1:
Fixed in live effect

PR #896, opened 2020-06-18
OboeTester: start SCO when needed

philburk:
Based on device ID.

Related to #874


robertwu1:
See https://github.com/google/oboe/wiki/TechNote_BluetoothAudio

PR #899, opened 2020-06-18
[samples/DrumThumper] Add build flavor for publishing to Play Store

dturner:
As per guidance here: https://developer.android.com/studio/build/application-id I have created a different build flavor which changes the `applicationId` to `com.mobileer.drumthumper` suitable for publishing to the Play Store.

### DIFF
--- a/samples/drumthumper/build.gradle
+++ b/samples/drumthumper/build.gradle
@@ -22,6 +22,20 @@ android {
         }
     }
 
+    // This sample will also be published on the Google Play Store which needs a different
+    // application id from "com.google" as this is reserved for official Google products.
+    flavorDimensions "publishing-target"
+    productFlavors {
+        codesample {
+            dimension "publishing-target"
+            applicationId "com.google.oboe.samples.drumthumper"
+        }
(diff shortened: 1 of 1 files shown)


samples/drumthumper/build.gradle
```diff
@@ -22,6 +22,20 @@ android {
         }
     }
 
+    // This sample will also be published on the Google Play Store which needs a different
+    // application id from "com.google" as this is reserved for official Google products.
+    flavorDimensions "publishing-target"
+    productFlavors {
+        codesample {
```
philburk, line 29: I interpret the doc differently. I think the flavors are for multiple products in the store, like "free" and "pro".
I think all we need to do is set the applicationId in build.gradle to "com.mobileer.drumthumper", which it is currently.

dturner: By jove, I think you're right! Have changed the package names but left the applicationId as it was. This keeps things very simple - thanks for investigating.

dturner:
I'm going to rebase this into the other PR and we can follow up discussions there.

PR #916, opened 2020-06-30
hello-oboe: store properties, refactor native code

### LINKED ISSUE #879: HelloOboe ignores playback device and other user settings
Android version(s):10
Android device(s): Galaxy S20 Ultra 5G
Oboe version:master
App name used for testing (please try to reproduce the issue using the Oboe samples and apps): HelloOboe

**Short description**
Playback device dropdown appears to be ignored

**Steps to reproduce**
Reproducer 1
Run demo
Select "SM-G988B built-in earphone speaker"
touch screen, listen
select "SM-G988B built-in speaker"
touch screen, listen

Reproducer 2
Run demo
select "built-in earphone speaker"
touch screen
= hear two tones
select channel count 1
touch screen
= hear one tone
select "built-in speaker"
touch screen
= hear one tone
select "built-in earphone speaker"
touch screen
= hear two tones, even though channel count is still 1

**Expected behavior**
Heard audio should change due to different speaker
Channel count should be respected across changing output device

**Actual behavior**
Audio remains the same. I think it's always playing back on earphone speaker.
Channel count goes wrong

**Any additional context**
The same tests on Pixel 2 XL results in a clear audible difference between earphone speaker and non-earphone speaker, with channel count resulting in correct tones.

philburk:
Save properties so the user selections are honored.
Use shared stream.
Use locks to prevent race conditions and crashes.

See #879

### DIFF
--- a/samples/hello-oboe/src/main/cpp/HelloOboeEngine.cpp
+++ b/samples/hello-oboe/src/main/cpp/HelloOboeEngine.cpp
@@ -42,22 +42,21 @@ HelloOboeEngine::HelloOboeEngine()
 double HelloOboeEngine::getCurrentOutputLatencyMillis() {
     if (!mIsLatencyDetectionSupported) return -1.0;
 
-    // Save local copy of shared pointer while we measure latency.
-    std::shared_ptr<oboe::AudioStream> localStream = mStream;
-    if (localStream == nullptr) return -1.0;
+    std::lock_guard<std::mutex> lock(mLock);
+    if (!mStream) return -1.0;
 
     // Get the time that a known audio frame was presented for playing
-    auto result = localStream->getTimestamp(CLOCK_MONOTONIC);
+    auto result = mStream->getTimestamp(CLOCK_MONOTONIC);
     double outputLatencyMillis = -1;
     const int64_t kNanosPerMillisecond = 1000000;
     if (result == oboe::Result::OK) {
         oboe::FrameTimestamp playedFrame = result.value();
         // Get the write index for the next audio frame
-        int64_t writeIndex = localStream->getFramesWritten();
+        int64_t writeIndex = mStream->getFramesWritten();
         // Calculate the number of frames between our known frame and the write index
         int64_t frameIndexDelta = writeIndex - playedFrame.position;
         // Calculate the time which the next frame will be presented
-        int64_t frameTimeDelta = (frameIndexDelta * oboe::kNanosPerSecond) /  (localStream->getSampleRate());
+        int64_t frameTimeDelta = (frameIndexDelta * oboe::kNanosPerSecond) /  (mStream->getSampleRate());
         int64_t nextFramePresentationTime = playedFrame.timestamp + frameTimeDelta;
         // Assume that the next frame will be written at the current time
         using namespace std::chrono;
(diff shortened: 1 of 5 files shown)


samples/hello-oboe/src/main/cpp/HelloOboeEngine.cpp
```diff
@@ -34,25 +34,30 @@
...
+        : mLatencyCallback(std::make_unique<LatencyTuningCallback>(*this)) {
     start();
-    updateLatencyDetection();
 }
 
 double HelloOboeEngine::getCurrentOutputLatencyMillis() {
-    if (!mIsLatencyDetectionSupported) return -1;
+    if (!mIsLatencyDetectionSupported) return -1.0;
+
+    // Save local copy of shared pointer while we measure latency.
```
dturner, line 45: Why is it necessary to copy the shared_ptr to a local var?

philburk: We can delete the object if we only have one shared_ptr.  The code that reopens the stream will close and delete the old stream in mStream when it creates the new one. If we are referencing mStream then we may see it get deleted. By copying the shared_ptr we now hold a reference and can prevent the deletion. Another way to handle this would be to hold the mLock in this method. (I may have this all wrong. So I am glad you are asking.)

samples/hello-oboe/src/main/cpp/HelloOboeEngine.cpp
```diff
@@ -34,25 +34,30 @@
...
+    std::shared_ptr<oboe::AudioStream> localStream = mStream;
```
dturner, line 46: @mnaganov Would mind taking a look at this? We are struggling to understand whether the copying of the `shared_ptr` is necessary.

philburk: I think maybe I should use lock(mLock) here and then I will not need to copy the shared_ptr.

dturner: Been thinking some more about this. The issue here is the assumption that `mStream` will always point to a valid stream object, and in a state where `getTimestamp` and the other methods for latency detection make sense. That isn't necessarily true: 

- `createPlaybackStream` may not have been called yet
- The current playback device may have been unplugged (e.g. headphones) leaving the stream in an indeterminate state of closing/closed/deleted
- The user may have changed the playback device in the drop down so the stream is (by user request) being closed and another stream being restarted, during which time latency detection will be unavailable (although the method will still currently be being called since it's on a separate JVM thread using `TimerTask`). 

This is a common problem: accessing streams and stream properties from multiple threads so it'd be good to sit down and thrash out all the scenarios (especially the stream lifecycle) and how to handle them correctly so we can document/update code samples throughout.

philburk: We do not assume the stream is open. See line 47.

philburk: At the moment HelloOboe is broken. This CL fixes the main problems. Can we make more improvements in the next CL?

philburk: I switched it to use a lock instead of localstream.

samples/hello-oboe/src/main/cpp/HelloOboeEngine.cpp
philburk: I switched it to use a lock instead of localstream.

philburk:
@dturner - hello-oboe is currently broken. I'd like to merge this to fix it and then consider better long term solutions.

samples/hello-oboe/src/main/cpp/HelloOboeEngine.cpp
dturner: > We can delete the object if we only have one shared_ptr. 

I getcha. Because we have 2 threads accessing `HelloOboeEngine` - the main thread and the "update latency timer"  thread we need to copy the `shared_ptr` to avoid the main thread from deleting the stream whilst it's being accessed by the "update latency timer" thread. Is that correct?

rpattabi:
> Save properties so the user selections are honored.
> Use shared stream.
> Use locks to prevent race conditions and crashes.

We are trying to understand if we need locks in our code similar to what is done in Hello Oboe. It is not clear why Hello Oboe needs locks. Which scenarios would otherwise cause race conditions and crashes in Hello Oboe?

We see Oboe Tester also uses shared pointers for streams but no locks.

PR #941, opened 2020-07-09
Add outer namespace to oboe flowgraph

### LINKED ISSUE #930: Oboe unit tests crashing with memory corruption, library cross linking the flowgraph code
Oboe version: <= 1.4.2
App name used for testing:  unit tests

**Short description**
Unit tests that open and close an MMAP stream crash due to memory corruption.
The root cause is because AAudio is creating an object using the flowgraph code in AAudio,
but calling the destructor in Oboe!
The Oboe and AAudio flowgraph code is similar but not identical. So it crashes.

**Steps to reproduce**
Run adb logcat in a separate window.
Follow instructions at: https://github.com/google/oboe/tree/master/tests
to run the tests.

**Expected behavior**
Test runs to completion and and reports number of tests that PASS.

**Actual behavior**
Test stops before final report.
Logcat shows Native Exception and stack trace like the one below.

**Any additional context**
See internal b/160633017

I added -fvisibility=hidden and it seemed to fix the problem.

Here is a stack dump:
07-06 14:32:54.969 F libc    : Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x7afffffff0 in tid 11546 (testOboe), pid 11546 (testOboe)

07-06 14:32:55.046 F DEBUG   :       #00 pc 00000000000444fc  /apex/com.android.runtime/lib64/bionic/libc.so (scudo::Allocator<scudo::AndroidConfig, &(scudo_malloc_postinit)>::deallocate(void*, scudo::Chunk::Origin, unsigned long, unsigned long)+132) (BuildId: e0da623f4ab0530a50a9af9e181b341d)
07-06 14:32:55.046 F DEBUG   :       #01 pc 0000000000074110  /data/user/0/com.google.oboe.tests.unittestrunner/files/testOboe
07-06 14:32:55.046 F DEBUG   :       #02 pc 0000000000035144  /system/lib64/libaaudio_internal.so (AAudioFlowGraph::~AAudioFlowGraph()+148) (BuildId: 572fc676500947b5f7300082d299c641)
07-06 14:32:55.046 F DEBUG   :       #03 pc 0000000000033b80  /system/lib64/libaaudio_internal.so (aaudio::AudioStreamInternalPlay::~AudioStreamInternalPlay()+36) (BuildId: 572fc676500947b5f7300082d299c641)
07-06 14:32:55.046 F DEBUG   :       #04 pc 0000000000003b90  /system/lib64/libaaudio.so (AAudioStream_close.cfi+356) (BuildId: f0582a3365d1efd6d1317e4eb2a35e31)

philburk:
Fixes #930


src/flowgraph/ChannelCountConverter.h
```diff
@@ -48,5 +49,6 @@ namespace flowgraph {
     };
 
 } /* namespace flowgraph */
+} /* namespace oboe */
```
philburk, line 52: Should be namespace FLOWGRAPH_OUTER_NAMESPACE

src/flowgraph/FlowGraphNode.h
```diff
@@ -40,8 +40,16 @@
 
 // Set this to 1 if using it inside the Android framework.
 // This code is kept here so that it can be moved easily between Oboe and AAudio.
+#ifndef FLOWGRAPH_ANDROID_INTERNAL
 #define FLOWGRAPH_ANDROID_INTERNAL 0
+#endif
 
+// Set this to a name that will prevent AAudio from calling into Oboe.
```
dturner, line 47: I think this needs a clearer description/comment since most developers won't know what this refers to.

src/flowgraph/FlowGraphNode.h
```diff
@@ -40,8 +40,16 @@
...
+#ifndef FLOWGRAPH_OUTER_NAMESPACE
+#define FLOWGRAPH_OUTER_NAMESPACE oboe
```
dturner, line 49: Would it make sense to define the namespace based on whether `FLOWGRAPH_ANDROID_INTERNAL` is set to `0`?

philburk:
>  I'm not entirely comfortable with contaminating the oboe codebase to facilitate the
> copying of flowgraph between internal and external sources (this in itself is a code smell),

Sharing flowgraph code between Oboe and AAudio is "code reuse", which is a good thing.
Also we plan to add Oboe to Android as /external/Oboe and may use the oboe::flowgraph code directly from AAudio.
This namespace change will help with that.

I don't think this smells too bad. The alternative is forked code or memory corruption. Both of which smell much worse.

Also, changing the namespace to oboe::flowgraph does not seem so bad. This flowgraph code is part of Oboe.

src/flowgraph/FlowGraphNode.h
philburk: I did that originally and then decided this was better.

I want to be able to set the namespace to aaudio::flowgraph when we use it in AAudio.
Also if someone wants to steal it, this will make it easier to customize the namespace.

src/flowgraph/FlowGraphNode.h
philburk: OK

PR #945, opened 2020-07-13
oboe: cleanup fifo includes

### LINKED ISSUE #927: FifoBuffer.cpp includes AudioClock.h , but doesn't depend on it.
Oboe version:
1.4.2
 
While trying to compile FifoBuffer outside of oboe, it failed when trying to include AudioClock.h (due to missing clock_nanosleep function).
When digging deeper, I found out that it's not actually needed by FifoBuffer.cpp
I suggest to remove the `#include "common/AudioClock.h"` line from FifoBuffer.cpp

philburk:
Remove AudioClock.h and other unnecessary includes.

Fixes #927

### DIFF
--- a/src/fifo/FifoBuffer.cpp
+++ b/src/fifo/FifoBuffer.cpp
@@ -14,18 +14,14 @@
  * limitations under the License.
  */
 
-#include <stdint.h>
-#include <time.h>
-#include <memory.h>
-#include <cassert>
 #include <algorithm>
+#include <memory.h>
+#include <stdint.h>
 
-#include "common/OboeDebug.h"
 #include "fifo/FifoControllerBase.h"
 #include "fifo/FifoController.h"
 #include "fifo/FifoControllerIndirect.h"
 #include "fifo/FifoBuffer.h"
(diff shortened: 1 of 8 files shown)


src/fifo/FifoBuffer.cpp
```diff
@@ -14,18 +14,14 @@
  * limitations under the License.
  */
 
-#include <stdint.h>
-#include <time.h>
-#include <memory.h>
-#include <cassert>
 #include <algorithm>
+#include <memory.h>
+#include <stdint.h>
```
dturner, line 19: We might want to use `<cstdint>` instead of `<stdint.h>` as this seems to be the RightWay™. Source: https://stackoverflow.com/questions/13642827/cstdint-vs-stdint-h

philburk: Hmm. That blog post says that using just <cstdint> is a "bad idea" and that we need to also add:

using std::int32_t;
using std::uint32_t;
etc.

I think that might be a good thing to do throughout Oboe.
Let's maybe do that in a separate CL.

Issue #956, opened 2020-07-23
LiveEffect 


vbarthel-fr:
Android version(s): 10
Android device(s): Essential PH-1
Oboe version: 5b7fe09f39baa2cd62f462d2b4d530f1e63f59f3
App name used for testing: LiveEffect
(Please try to reproduce the issue using the OboeTester or an Oboe sample.)
 
**Short description**

It seems that the sample rate from the playback stream is not used to create the record stream despite the note.

```
// Note: The order of stream creation is important. We create the playback
// stream first, then use properties from the playback stream
// (e.g. sample rate) to create the recording stream. By matching the
// properties we should get the lowest latency path
```

![Screenshot 2020-07-23 at 10 35 50](https://user-images.githubusercontent.com/3321467/88267865-c6936100-ccd1-11ea-9a5b-315f2bcfcdc6.png)

**Steps to reproduce**
- Connect a [Sony WI-200 headphone](https://www.sony.co.uk/electronics/in-ear-headphones/wi-c200).
- Start the LiveEffect demo
- Observe that the sample rate of the playback stream is not used to create the record stream

Note: I have only tested with a Sony WI-200.

**Expected behavior**
- The sample rate of the playback stream is passed to the creation of the record stream.

**Actual behavior**
- `oboe::kUnspecified` is passed to the creation of the record stream.

**Any additional context**

From my understanding of the code, this issue might have been introduced here: https://github.com/google/oboe/pull/522/files#diff-05d4f44cf54f4c39f26379303b1f9d3aL155 `mSampleRate` is no longer set to the playback stream sample rate.

vbarthel-fr:
The associated PR has been merged: I'm closing this issue :)

PR #957, opened 2020-07-23
live-effect: use playback stream sample rate to open record stream

vbarthel-fr:
Related issue: https://github.com/google/oboe/issues/956


vbarthel-fr:
Before:

Screenshot 2020-07-23 at 10 35 50

After:

Screenshot 2020-07-23 at 10 37 15

PR #970, opened 2020-08-06
oboe: delay before closing an AAudio stream

### LINKED ISSUE #967: Prevent callback after close by sleeping before close
See #953

I could only reproduce this when using Bluetooth.

I then tried calling usleep(N*1000) before the stream is closed.
N.      crashes:tries
10 => 0:10
 5 => 0:10
 1 => 1:10 in onDeviceUpdate()
no => 8:11

So it is a pretty tight race. It seems a short delay right before the close is enough to prevent the race. Five msec works OK but ten msec should be a safer value.

Add a sleep in the Oboe close method.

_Originally posted by @philburk in https://github.com/google/oboe/issues/953#issuecomment-669576113_

philburk:
This is a workaround for a possible callback after close that can occur
if close is called immediately after a stop.
This can occur on P, particularly when using Bluetooth devices.

Fixes #967

### DIFF
--- a/src/aaudio/AudioStreamAAudio.cpp
+++ b/src/aaudio/AudioStreamAAudio.cpp
@@ -118,7 +118,7 @@ void AudioStreamAAudio::internalErrorCallback(
     if (oboeStream->wasErrorCallbackCalled()) { // block extra error callbacks
         LOGE("%s() multiple error callbacks called!", __func__);
     } else if (stream != oboeStream->getUnderlyingStream()) {
-        LOGW("%s() stream already closed", __func__); // can happen if there are bugs
+        LOGW("%s() stream already closed or closing", __func__); // can happen if there are bugs
     } else if (sharedStream) {
         // Handle error on a separate thread using shared pointer.
         std::thread t(oboe_aaudio_error_thread_proc_shared, sharedStream,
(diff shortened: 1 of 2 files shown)


src/aaudio/AudioStreamAAudio.cpp
```diff
@@ -298,6 +297,14 @@ Result AudioStreamAAudio::close() {
     // This will delete the AAudio stream object so we need to null out the pointer.
     AAudioStream *stream = mAAudioStream.exchange(nullptr);
     if (stream != nullptr) {
+        // Sometimes a callback can occur shortly after a stream has been stopped and
+        // even after a close. If the stream has been closed then the callback
+        // can access memory that has been freed. That causes a crash.
+        // Two milliseconds may be enough but 10 msec is even safer.
+        // This seems to be less likely in Q but we have seen some reports.
+        if (getSdkVersion() <= __ANDROID_API_Q__) {
+            usleep(kDelayBeforeCloseMillis * 1000);
```
rpattabi, line 306: Clients could call close() even after the stream is closed. In those cases, the forced delay may be unnecessary. Would it help to check the state before enforcing the delay?

philburk: If close () is called a second time then stream would be null and that code would be skipped.

rpattabi: I see now. Thanks for the clarification.

PR #988, opened 2020-08-25
  oboe: fix sample rate converter for input streams

### LINKED ISSUE #987: Oboe sample rate conversion has no callbacks for input streams
Android version(s): Q, N/A
Android device(s): Bonito
Oboe version: current > 1.4.3
 
**Short description**
Oboe sample rate conversion has no callbacks for input streams

**Steps to reproduce**
Launch OboeTester
Tap TEST INPUT
Select SRate: 44100
Select Convert: SRC: Medium
Tap OPEN then START

**Expected behavior**
VU meters responding
#cb advancing

**Actual behavior**
No VU meter activity
#cb = 0

**Any additional context**

SRC: None is OK

I bisected the build using:

    git switch --detach {commit}

and the problem seemed to appear in #395 when I updated the flowgraph code.

philburk:
The callbackCount was not handled correctly.
This could cause a stream to stall and stop calling callbacks.
    
Add unit tests.

Fixes #987


philburk:
This fixes a regression caused by #395 after the 1.4.2 and 1.4.3 releases.

PR #1010, opened 2020-09-14
Simple fix to hello-oboe related to shared_ptr

ggfan:
At the app starting up time, inside the HelloOboeEngine constructor, mStream would always nullptr as no allocation to the raw pointer yet; when calling into HelloOboeEngine::start(), so it would still be nullptr, so check for nullptr would always fail, hence mStream would never get chance to be allocated; when later calling into HelloOboeEngine::tap(), a nullptr is dereferenced, resulting to a crash.

Since builder->openStream() reset the share_ptr, it could tolerate a nullptr. The fix is enable nullptr case in mStream in this sample, there will not be memory leak side-effect.


philburk (approved):
Thanks! I'm not sure how this could have passed before. Your change seems right.

Issue #989, opened 2020-08-25
Add more parameter checking to AudioStreamBuilder


philburk:
We should add some more general parameter checking to AudioStreamBuilder. That will get called in the open() and return an error message specific to certain illegal parameters, eg. ErrorInvalidFormat or ErrorInvalidRate.

_Originally posted by @philburk in https://github.com/google/oboe/pull/553#discussion_r476600888_

ggfan:
working on it

PR #1019, opened 2020-09-19
Add parameter validation to AudioStreamBase

ggfan:
Fixing issue #989 , PTAL

### DIFF
--- a/include/oboe/AudioStreamBase.h
+++ b/include/oboe/AudioStreamBase.h
@@ -195,6 +195,31 @@ protected:
     bool                            mFormatConversionAllowed = false;
     // Control whether and how Oboe can convert sample rates to achieve optimal results.
     SampleRateConversionQuality     mSampleRateConversionQuality = SampleRateConversionQuality::None;
+
+    /** Validate stream parameters that might not be checked in lower layers */
+    virtual Result isValidConfig() {
+        switch (mFormat) {
+            case AudioFormat::Unspecified:
+            case AudioFormat::I16:
+            case AudioFormat::Float:
+                break;
+
+            default:
+                return Result::ErrorInvalidFormat;
+        }
+
+        switch (mSampleRateConversionQuality) {
+            case SampleRateConversionQuality::None:
+            case SampleRateConversionQuality::Fastest:
+            case SampleRateConversionQuality::Low:
+            case SampleRateConversionQuality::Medium:
+            case SampleRateConversionQuality::High:
+            case SampleRateConversionQuality::Best:
+                return Result::OK;
+            default:
+                return Result::ErrorIllegalArgument;
+        }
+    }
 };
 
 } // namespace oboe
--- a/src/common/AudioStreamBuilder.cpp
+++ b/src/common/AudioStreamBuilder.cpp
@@ -187,11 +187,11 @@ Result AudioStreamBuilder::openStream(AudioStream **streamPP) {
 }
 
 Result AudioStreamBuilder::openManagedStream(oboe::ManagedStream &stream) {
+    stream.reset();
     auto result = isValidConfig();
     if (result != Result::OK) {
         return result;
     }
-    stream.reset();
     AudioStream *streamptr;
     result = openStream(&streamptr);
     stream.reset(streamptr);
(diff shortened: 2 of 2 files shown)


include/oboe/AudioStreamBase.h
```diff
@@ -195,6 +195,28 @@ class AudioStreamBase {
     bool                            mFormatConversionAllowed = false;
     // Control whether and how Oboe can convert sample rates to achieve optimal results.
     SampleRateConversionQuality     mSampleRateConversionQuality = SampleRateConversionQuality::None;
+
+    /** Validate the existing stream configurations */
+    virtual Result isValidConfig() {
+        if (mFormat < AudioFormat::Invalid || mFormat > AudioFormat::Float) {
+            return Result::ErrorInvalidFormat;
+        }
+        if ((mSharingMode != SharingMode::Exclusive && mSharingMode != SharingMode::Shared) ||
```
dturner, line 204: This is extremely difficult to parse. Please consider splitting into individual methods e.g. `isSharingModeValid()`, `isPerformanceModeValid()` etc to make it easier to read.

dturner: Additionally, it may be worth logging an error inside each method so the caller knows exactly what was wrong.

include/oboe/AudioStreamBase.h
```diff
@@ -195,6 +195,28 @@ class AudioStreamBase {
...
+            (mDirection != Direction::Input && mDirection != Direction::Output) ||
+            (mPerformanceMode != PerformanceMode::None && mPerformanceMode != PerformanceMode::PowerSaving &&
+             mPerformanceMode != PerformanceMode::LowLatency)    ||
+            (mUsage < Usage::Media || mUsage > Usage::Assistant) ||
+            (mContentType < ContentType::Speech || mContentType > ContentType::Sonification) ||
```
philburk, line 209: No comparison operators for enums.
The goal was to have more informative error codes. So for these I would just let AAudio catch them.

src/common/AudioStreamBuilder.cpp
```diff
@@ -183,17 +187,26 @@ Result AudioStreamBuilder::openStream(AudioStream **streamPP) {
 }
 
 Result AudioStreamBuilder::openManagedStream(oboe::ManagedStream &stream) {
+    auto result = isValidConfig();
+    if (result != Result::OK) {
+        return result;
+    }
     stream.reset();
```
philburk, line 194: Maybe we should reset before the check.

src/common/AudioStreamBuilder.cpp
```diff
@@ -183,17 +187,26 @@ Result AudioStreamBuilder::openStream(AudioStream **streamPP) {
...
     return result;
 }
 
 Result AudioStreamBuilder::openStream(std::shared_ptr<AudioStream> &sharedStream) {
+    auto result = isValidConfig();
+    if (result != Result::OK) {
+        return result;
+    }
+
     sharedStream.reset();
```
philburk, line 207: ditto

philburk (changes requested):

include/oboe/AudioStreamBase.h
ggfan: updated, please take a look again

src/common/AudioStreamBuilder.cpp
ggfan: done

src/common/AudioStreamBuilder.cpp
ggfan: done

include/oboe/AudioStreamBase.h
```diff
@@ -195,6 +195,31 @@ class AudioStreamBase {
...
+    /** Validate stream parameters that might not be checked in lower layers */
+    virtual Result isValidConfig() {
+        switch (mFormat) {
+            case AudioFormat::Unspecified:
+            case AudioFormat::I16:
+            case AudioFormat::Float:
+                break;
+
+            default:
+                return Result::ErrorInvalidFormat;
```
philburk, line 208: Don suggested logging an error message here and at line 220.

philburk (changes requested):

philburk (approved):

include/oboe/AudioStreamBase.h
philburk: Gerry pointed out this is in a header file.

PR #1020, opened 2020-09-20
Fixed compilation error on NDK <21

katyo:


philburk:
Thanks for this fix. Let's merge this now to fix the build. And then maybe we'll move the definition to a more common shared file later.

Issue #1006, opened 2020-09-07
Input preset Unprocessed results in silence on some devices?


rpattabi:
Android version(s): 10 (API 29)
Android device(s): samsung SM-G965F (samsung star2lte) and more?
Oboe version: 1.4.2
App: [Shruti Carnatic Tuner](https://play.google.com/store/apps/details?id=org.kuyil.shruti) v7.0.2

We migrated our app to use oboe and released it recently. This is a tuner app which listens to incoming sound, detects its pitch, and shows the notes. For the input stream, we use `Unprocessed` as input preset. We know `Unprocessed` gives weaker volume levels and we handled it in our app. On the devices we tested, this is working perfectly. However, we started seeing users report that the note detection is not working on some devices. Based on the interaction with a user, we suspect the samsung device mentioned above produces silence in the input stream. We also asked the user to switch off Google Assistant to ensure that it is not interfering with the input which didn't address the problem.

We think that this may be the case with other devices. As a workaround, we asked the user to switch to our _old engine_ which uses OpenSL ES directly without oboe, and it works for this user. No input preset is set while creating this opensl stream, so we assume it uses `Generic` or whichever is the default.

Is it wrong to expect `Unprocessed` to be working on all the devices? Is there anything we could do about this problem? We're particularly interested in getting input signal without any noise suppression which at times alters the pitch of the sound.

rpattabi:
Another device with the same problem:

Android version(s): 8.0 (API 26)
Android device(s): samsung SM-G935F (samsung hero2lte)
Oboe version: 1.4.2

dturner:
Thanks for reporting this. I have an SM-G965F so will attempt to reproduce using OboeTester and investigate further.

rpattabi:
More devices with the same problem:

Device: Nokia Nokia 7 plus (HMD Global B2N_sprout)
Android: 29 (10)

Device: motorola Moto G (5S) (motorola montana)
Android: 25 (7.1.1)
Oboe: 1.4.2

dturner:
I've been testing this today on an Samsung Galaxy S9 SM-G965F (Build PPR.180610.011.G965FXXS7CTA2) running Android 9. 

Here's how I tested this behaviour:

- Load the LiveEffect sample app
- Set the media volume to zero by pressing the volume down button
- Tap start
- Increase the media volume by pressing the volume up button, count the number of presses until feedback occurs. 

I tested this with every input preset (by changing the input stream recording parameters in code). Results:

[...]

PR #1021, opened 2020-09-21
[LiveEffect] Add ability to specify input preset

dturner:
Adds a dropdown list to the LiveEffect sample, allowing the user to specify an input preset. This enables easier debugging of issues like https://github.com/google/oboe/issues/1006.


philburk:
We can always use OboeTester for this kind of debugging. It might be better to keep the examples as simple as possible.

dturner:
Good point. I'll discard this and make the changes to OboeTester instead.

dturner:
Done: https://github.com/google/oboe/pull/1022

PR #1022, opened 2020-09-22
[OboeTester] Put echo test layout inside ScrollView

dturner:
Had to put the echo test layout inside a `ScrollView` as it was too high to fit on a single screen.


dturner:
Will enable easier testing of [this issue](https://github.com/google/oboe/issues/1006)

philburk (approved):

PR #1016, opened 2020-09-17
Convert samples to use shared_ptr stream creation interface

ggfan:
This is for github issue https://github.com/google/oboe/issues/830, please kindly take a look, thank you!

PR #1023, opened 2020-09-24
Minor update to MegaDrone sample for stream creation

ggfan:
To lift stream creation from constructor to a public start() function: this is a followup from [the shared_ptr change pull request](https://github.com/google/oboe/pull/1016), PTAL

### DIFF
--- a/samples/MegaDrone/src/main/cpp/MegaDroneEngine.cpp
+++ b/samples/MegaDrone/src/main/cpp/MegaDroneEngine.cpp
@@ -36,7 +36,8 @@ MegaDroneEngine::MegaDroneEngine(std::vector<int> cpuIds) {
 
 MegaDroneEngine::~MegaDroneEngine() {
     if (mStream) {
-        LOGE("Stop MegaDroneEngine before deleting it!");
+        LOGE("MegaDroneEngine destructor was called without calling stop()."
+             "Please call stop() to ensure stream resources are not leaked.");
         stop();
     }
 }
--- a/samples/MegaDrone/src/main/cpp/MegaDroneEngine.cpp
+++ b/samples/MegaDrone/src/main/cpp/MegaDroneEngine.cpp
@@ -31,13 +31,14 @@
  * @param cpuIds
  */
 MegaDroneEngine::MegaDroneEngine(std::vector<int> cpuIds) {
-
     createCallback(cpuIds);
-    start();
 }
 
 MegaDroneEngine::~MegaDroneEngine() {
-    stop();
+    if (mStream) {
+        LOGE("Stop MegaDroneEngine before deleting it!");
+        stop();
+    }
 }
 
 void MegaDroneEngine::tap(bool isDown) {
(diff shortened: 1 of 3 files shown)


samples/MegaDrone/src/main/cpp/MegaDroneEngine.cpp
```diff
@@ -70,23 +71,26 @@ void MegaDroneEngine::createCallback(std::vector<int> cpuIds){
...
         mStream->start();
+        return true;
     } else {
         LOGE("Failed to create the playback stream. Error: %s", convertToText(result));
+        return false;
     }
 }
 
-void MegaDroneEngine::stop() {
+bool MegaDroneEngine::stop() {
```
dturner, line 88: Why make this a `bool` if it always returns true and none of the callers check the return value?

samples/MegaDrone/src/main/cpp/MegaDroneEngine.cpp
```diff
@@ -31,13 +31,14 @@
...
 MegaDroneEngine::MegaDroneEngine(std::vector<int> cpuIds) {
-
     createCallback(cpuIds);
-    start();
 }
 
 MegaDroneEngine::~MegaDroneEngine() {
-    stop();
+    if (mStream) {
+        LOGE("Stop MegaDroneEngine before deleting it!");
```
dturner, line 39:
```suggestion
        LOGE("MegaDroneEngine destructor was called without calling stop(). Please call stop() to ensure stream resources are not leaked.");
```

samples/MegaDrone/src/main/cpp/MegaDroneEngine.cpp
dturner:
```suggestion
void MegaDroneEngine::stop() {
```

samples/MegaDrone/src/main/cpp/MegaDroneEngine.h
```diff
@@ -40,15 +40,16 @@ class MegaDroneEngine : public IRestartable {
     // from IRestartable
     virtual void restart() override;
 
+    bool start();
+    bool stop();
```
dturner, line 44:
```suggestion
    void stop();
```

dturner (changes requested):
Thanks for doing this. Few minor comments.

dturner (approved):

samples/MegaDrone/src/main/cpp/MegaDroneEngine.cpp
ggfan: done

samples/MegaDrone/src/main/cpp/MegaDroneEngine.h
ggfan: thank you for bring this one up, this is just for the interface consistence purpose with regarding to  ` bool start();`

philburk (approved):

PR #1028, opened 2020-09-29
fixes #1027 handling of skipped chunks during wav parse

### LINKED ISSUE #1027: parselib handling 'other' chunks incorrectly
Android version(s): all
Android device(s): all
Oboe version: 1.4.3
 
**Short description**
The wav reader will crash when asked to parse a wav file with a chunk type other than RIFF, fmt, or data. In the 'else' branch, we want to skip the contents of 'chunk', but we instead try to access the length of mDataChunk, which is null before hitting the data tag.

**Steps to reproduce**
Create a meminputstream with a wav file that contains a chunk type such as 'LIST'

**Expected behavior**
The LIST chunk gets parsed and skipped.

**Actual behavior**
When trying to skip the chunk, it dereferences the wrong chunk throwing an error. 

**Device**
All

jbeta51:
Skipped chunks should now be skipped properly without crashing.

### DIFF
--- a/samples/parselib/src/main/cpp/wav/WavStreamReader.cpp
+++ b/samples/parselib/src/main/cpp/wav/WavStreamReader.cpp
@@ -94,7 +94,7 @@ void WavStreamReader::parse() {
         } else {
             chunk = new WavChunkHeader(tag);
             chunk->read(mStream);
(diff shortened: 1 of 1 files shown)


android-usb (approved):
Yep, that was wrong.

PR #1035, opened 2020-10-08
Add virtual destructor to FlowGraphPort.

jimmytobin2425:
Define a virtual destructor for base class.


philburk:
The failed Travis-CI is because of a known issue in OboeTester.
Rebasing and merging on top of the fix.

PR #1036, opened 2020-10-12
OboeTester: check for errorCallback error code

philburk:
Must be ERROR_DISCONNECTED to pass.
See b/170551030


ggfan (dismissed):
could you add the changes in https://github.com/google/oboe/pull/1038 to confirm the CI could pass?  thank you

Issue #1037, opened 2020-10-12
Travis CI failing for gradle:4.0.0


philburk:
Travis CI for Pull Request #1036 failed with:

    Could not resolve com.android.tools.build:gradle:4.0.0.

The PR did not change any gradle files or other build files.

CI was working 4 days ago when #1035 was pushed.
There have been no commits since then. But when I reran Travis on #1035 it failed!

philburk:
Fixed by #1040

PR #1038, opened 2020-10-12
Make google maven before jcenter for OboeTester

ggfan:
looks like this could be a problem, merging this in?


philburk:
This PR built OK so it appears to fix #1037

philburk (approved):
CI now passing.

ggfan:
Withdrawing this PR, the fix will be merged into Phil's PR.

PR #1040, opened 2020-10-12
Make google maven before jcenter for OboeTester

ggfan:
This would fix CI failures in PR #1036, PTAL.


philburk:
Fixes #1037

philburk (approved):
Builds OK now.

PR #1050, opened 2020-10-22
oboe: change VoicePerformance to VoiceRecognition on P

### LINKED ISSUE #1009: Input stream fails to open on Android <10 when VoicePerformance preset is specified
Android version(s): Android 9
Android device(s): Samsung Galaxy S9
Oboe version: 1.4.3
App name used for testing: LiveEffect
 
**Short description**
When you attempt to open an input stream using the `VoicePerformance` preset on Android 9 or below the stream will fail to open because the `VoicePerformance` preset is only available on Android 10 and above. 

**Steps to reproduce**
On an Android 9 or below device load an app which will open an input stream in the following way. I modified the LiveEffect app for this. 

```
builder->setInputPreset(oboe::VoicePerformance)
builder->openStream(...)
```

**Expected behavior**
This is up for debate but I'd expect a suitable substitute for `VoicePerformance` (maybe `VoiceRecognition`) to be selected and the stream to open successfully. 

**Actual behavior**
The stream fails to open

philburk:
Fixes #1009

### DIFF
--- a/src/aaudio/AudioStreamAAudio.cpp
+++ b/src/aaudio/AudioStreamAAudio.cpp
@@ -213,8 +213,13 @@ Result AudioStreamAAudio::open() {
     }
 
     if (mLibLoader->builder_setInputPreset != nullptr) {
+        aaudio_input_preset_t inputPreset = mInputPreset;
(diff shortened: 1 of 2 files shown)


dturner (approved):
Great - thanks for doing this.

PR #1063, opened 2020-11-03
test and fix timeout in OpenSL ES waitForStateChange()

### LINKED ISSUE #1061: OpenSLES stream.stop(timeoutNanos) may never time out
The Oboe code for stop with a timeout looks odd. If you pass a timeoutNanoseconds > 0 then it can spin forever if the state does not change.
    https://github.com/google/oboe/blob/120b61d9b46ca8559790bce27bdd4c230b11a6cc/src/opensles/AudioStreamOpenSLES.cpp#L384

I think that should be:
    if (timeoutLeftNanos <= 0) {

philburk:
Fixes #1061

### DIFF
--- a/tests/testStreamWaitState.cpp
+++ b/tests/testStreamWaitState.cpp
@@ -64,6 +64,13 @@ protected:
 		EXPECT_EQ(Result::ErrorTimeout, result);
 	}
 
+    void checkWaitNonZeroTimeout() {
+        StreamState next = StreamState::Unknown;
+        int64_t timeout = 200 * kNanosPerMillisecond; // wait for a state change that will never come
+		Result result = mStream->waitForStateChange(mStream->getState(), &next, timeout);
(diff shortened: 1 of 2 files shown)


tests/testStreamWaitState.cpp
```diff
@@ -64,6 +64,13 @@ class TestStreamWaitState : public ::testing::Test {
 		EXPECT_EQ(Result::ErrorTimeout, result);
 	}
 
+    void checkWaitNonZeroTimeout() {
+        StreamState next = StreamState::Unknown;
+        int64_t timeout = 200 * kNanosPerMillisecond; // wait for a state change that will never come
+		Result result = mStream->waitForStateChange(mStream->getState(), &next, timeout);
```
philburk, line 70: Editor added TABS!

philburk:
This became hopelessly destroyed by a git merge.

PR #1079, opened 2020-11-18
OboeTester: more mic info and properties

philburk:
MinSPL, Sensitivity, etc
Also ProAudio Feature, etc.

### DIFF
--- a/apps/OboeTester/app/src/main/java/com/google/sample/oboe/manualtest/DeviceReportActivity.java
+++ b/apps/OboeTester/app/src/main/java/com/google/sample/oboe/manualtest/DeviceReportActivity.java
@@ -19,6 +19,7 @@ package com.google.sample.oboe.manualtest;
 import android.annotation.TargetApi;
 import android.app.Activity;
 import android.content.Context;
+import android.content.pm.PackageManager;
 import android.media.AudioDeviceCallback;
 import android.media.AudioDeviceInfo;
 import android.media.AudioManager;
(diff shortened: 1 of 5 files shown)


apps/OboeTester/app/src/main/java/com/google/sample/oboe/manualtest/DeviceReportActivity.java
```diff
@@ -128,6 +130,27 @@ public String reportAllMicrophones() {
         return report.toString();
     }
 
+    private String getExtraDeviceInfo() {
+        StringBuffer report = new StringBuffer();
+        report.append("\n############################");
+        report.append("\nExtras:\n");
+        String unprocessedSupport = mAudioManager.getParameters(AudioManager.PROPERTY_SUPPORT_AUDIO_SOURCE_UNPROCESSED);
+        report.append("\nSUPPORT_UNPROCESSED : " + ((unprocessedSupport != null) ? "yes" : "no"));
```
rpattabi, line 138: Since "no" could be incorrectly construed as Unprocessed is unavailable, could we have something like "SUPPORT_UNPROCESSED : full/partial"? We still should have "no" in case of pre-Android 7 as Unprocessed is actually not available there.

philburk: Thanks. I want to display what the device is reporting without too much interpretation.
So I changed it to yes/null.

PR #1082, opened 2020-11-18
oboe: some code cleanup suggested by Chrome

### LINKED ISSUE #590: Misc security comments
Aggregating some smaller changes from #352 into one issue.

I know this issue is a non-cohesive collection of smaller issues, but I don't think that any of these sub-issues warranted their own entry.

1) `FifoControllerBase::getFullFramesAvailable()`
Loss of capacity when casting from uint64_t to int32_t

2) `AudioOutputStreamOpenSLES::requestPause()` and  `AudioOutputStreamOpenSLES::requestStop()`
Handle/ignore negative values returned by `getFramesWritten()`

3) `AudioStreamBuffered::write(const void *buffer, ...)`
Why use a `const void*` if it's casted into `void*`

4) `AudioStreamBuffered::transfer` and `AudioStreamBuffered::read`
Handle/ignore the return value from `updateServiceFrameCounter()`

5) `AudioStreamOpenSLES::open()`
Check for overflows in `mFramesPerBurst * kBufferQueueLength`

philburk:
Check for errors from getFramesWritten().
Preserve const pointer in transfer().
Check for errors from updateServiceFrameCounter().
Check for overflows in capacity calculation.

Fixes #590

### DIFF
--- a/src/opensles/AudioStreamBuffered.cpp
+++ b/src/opensles/AudioStreamBuffered.cpp
@@ -110,12 +110,22 @@ int64_t AudioStreamBuffered::predictNextCallbackTime() {
 
 // Common code for read/write.
 // @return Result::OK with frames read/written, or Result::Error*
-ResultWithValue<int32_t> AudioStreamBuffered::transfer(void *buffer,
-                                      int32_t numFrames,
-                                      int64_t timeoutNanoseconds) {
+ResultWithValue<int32_t> AudioStreamBuffered::transfer(
+        void *readBuffer,
+        const void *writeBuffer,
+        int32_t numFrames,
+        int64_t timeoutNanoseconds) {
     // Validate arguments.
-    if (buffer == nullptr) {
-        LOGE("AudioStreamBuffered::%s(): buffer is NULL", __func__);
+    if (readBuffer != nullptr && writeBuffer != nullptr) {
+        LOGE("AudioStreamBuffered::%s(): both buffers are not NULL", __func__);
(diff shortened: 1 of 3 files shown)


src/opensles/AudioStreamBuffered.cpp
```diff
@@ -110,12 +110,22 @@ int64_t AudioStreamBuffered::predictNextCallbackTime() {
 
 // Common code for read/write.
 // @return Result::OK with frames read/written, or Result::Error*
-ResultWithValue<int32_t> AudioStreamBuffered::transfer(void *buffer,
-                                      int32_t numFrames,
-                                      int64_t timeoutNanoseconds) {
+ResultWithValue<int32_t> AudioStreamBuffered::transfer(
```
tguilbert-google, line 113: I haven't looked at the Oboe codebase in a while, so I'm not up to date with styles. However, I wonder if this would be easier to read with as two functions, instead of one where we expect to pass a nullptr every time.

tguilbert-google (approved):
LGTM with one NIT/Suggestion

src/opensles/AudioStreamBuffered.cpp
philburk: Most of the transfer function is the same for read and write. I do not like like to have common code. I suppose I could pass a lambda but that would be getting tricky.

This seemed like an OK compromise where I could keep the benefit of common code and also use const to enforce data protection.

> one where we expect to pass a nullptr every time

Oboe users will never call this function. It is protected and strictly for internal use.

PR #1088, opened 2020-11-19
DrumThumper, fix restart on device change.

android-usb:
Missed the "startStream()" call in onErrorAfterClose() after refactor.


samples/iolib/src/main/cpp/player/SimpleMultiPlayer.cpp
```diff
@@ -65,6 +65,8 @@ void SimpleMultiPlayer::onErrorAfterClose(AudioStream *oboeStream, Result error)
 
     resetAll();
     openStream();
+    startStream();
```
philburk, line 68: What if open fails?

philburk (changes requested):

philburk (approved):

PR #1091, opened 2020-11-20
Update sample code and documentation to use new data and error callback interfaces

dturner:
This PR refactors the sample code away from using a single callback (`AudioStreamCallback`) for both data and error events to using the new `AudioStreamDataCallback` and `AudioStreamErrorCallback` classes. 

The only issue I faced was that the original `DefaultAudioStreamCallback` (used by hello-oboe and megadrone) would attempt to set thread affinity inside the data callback and would reset it inside the error callback through a single field: `mIsThreadAffinitySet`. 

By separating the two callbacks the error callback no longer has access to this field. To solve this I added a `reset` method to `DefaultDataCallback` which can be called to reset the data callback to its initial configuration (in this case, only the thread affinity). 

The responsibility for calling `reset` moves away from the error callback, which just calls `restart` on its `IRestartable` parent. It's up to the parent to call `reset` on the data callback. Example here: https://github.com/google/oboe/compare/master...dturner:split_callbacks?expand=1#diff-300176f2b98467e0664a62ca6d40aa3f5c2401121741bcdea15a787557a6071bR133.

### DIFF
--- a/samples/hello-oboe/src/main/cpp/HelloOboeEngine.cpp
+++ b/samples/hello-oboe/src/main/cpp/HelloOboeEngine.cpp
@@ -35,7 +35,8 @@
  *
  */
 HelloOboeEngine::HelloOboeEngine()
-        : mLatencyCallback(std::make_unique<LatencyTuningCallback>(*this)) {
+        : mLatencyCallback(std::make_unique<LatencyTuningCallback>()),
+        mErrorCallback(std::make_unique<DefaultErrorCallback>(*this)){
 }
 
 double HelloOboeEngine::getCurrentOutputLatencyMillis() {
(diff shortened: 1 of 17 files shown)


samples/hello-oboe/src/main/cpp/HelloOboeEngine.cpp
```diff
@@ -35,7 +35,8 @@
  *
  */
 HelloOboeEngine::HelloOboeEngine()
-        : mLatencyCallback(std::make_unique<LatencyTuningCallback>(*this)) {
+        : mLatencyCallback(std::make_unique<LatencyTuningCallback>()),
+        mErrorCallback(std::make_unique<DefaultErrorCallback>(*this)){
```
philburk, line 39: space

samples/shared/DefaultErrorCallback.h
```diff
@@ -0,0 +1,56 @@
...
+ */
+class DefaultErrorCallback : public oboe::AudioStreamErrorCallback {
+public:
+
+    DefaultErrorCallback(IRestartable &parent): mParent(parent) {}
+    virtual ~DefaultErrorCallback() = default;
+
+    virtual void onErrorAfterClose(oboe::AudioStream *oboeStream, oboe::Result error) override {
+        // Restart the stream when it errors out with disconnect
+        if (error == oboe::Result::ErrorDisconnected) {
```
philburk, line 42: You may want to restart regardless of the error code because of the RQ1A bug.

philburk (changes requested):

philburk (approved):

philburk (approved):
We will coerce the ErrorTimeout to ErrorDisconnect in Oboe.

PR #1096, opened 2020-11-20
Add RQ1A disconnect bug to AndroidAudioHistory.md

philburk:
Document regression in headset plugin handling.

### DIFF
--- a/docs/AndroidAudioHistory.md
+++ b/docs/AndroidAudioHistory.md
@@ -4,7 +4,8 @@ Android audio history
 A list of important audio features, bugs, fixes and workarounds for various Android versions. [(List of all Android Versions)](https://developer.android.com/guide/topics/manifest/uses-sdk-element#ApiLevels)
 
 ### 11.0 R - API 30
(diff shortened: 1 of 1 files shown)


docs/AndroidAudioHistory.md
```diff
@@ -3,13 +3,17 @@ Android audio history
 
 A list of important audio features, bugs, fixes and workarounds for various Android versions. [(List of all Android Versions)](https://developer.android.com/guide/topics/manifest/uses-sdk-element#ApiLevels)
 
+### 11.0 R - API 30
+- Bug in AAudio on RQ1A. A stream is normally disconnected when a **headset is plugged in** or because of other device changes. An AAUDIO_ERROR_DISCONNECTED error code should be passed to the error callback. But a bug in Shared MMAP streams causes a different error to be returned. So if your error callback is checking for AAUDIO_ERROR_DISCONNECTED then it may not respond properly. We recommend **always stopping and closing the stream** regardless of the error code. Oboe does this. So if you are using Oboe callbacks you are OK. This issue was not in the original R release. It was introduced in RQ1A, which is being delivered by OTA starting in November 2020. It will be fixed in a future update.
```
dturner, line 7:
```suggestion
- Bug in **AAudio** on RQ1A. A stream is normally disconnected when a **headset is plugged in** or because of other device changes. An `AAUDIO_ERROR_DISCONNECTED` error code should be passed to the error callback. But a bug in Shared MMAP streams causes a different error to be returned. So if your error callback is checking for `AAUDIO_ERROR_DISCONNECTED` then it may not respond properly. We recommend **always stopping and closing the stream** regardless of the error code. Oboe does this. So if you are using Oboe callbacks you are OK. This issue was not in the original R release. It was introduced in RQ1A, which is being delivered by OTA starting in November 2020. It will be fixed in a future update.

```

dturner: You can format code using backtick `like this`

dturner (approved):

docs/AndroidAudioHistory.md
philburk: Thanks

PR #1105, opened 2020-11-24
oboe: workaround disconnect bug in RQ1A

philburk:
Coerce the ErrorTimeout to ErrorDisconnected if R.
This is to fix b/173928197

This can prevent apps that were looking specifically for the
ErrorDisconnected from failing to restart their streams.


dturner (approved):

philburk:
I tested this using the OboeTester TEST DISCONNECT test on my Pixel 3a running RQ1A.
It fails if "enable Oboe workarounds" is unchecked.
And passes if it is checked.

philburk:
@dturner - Thanks for the idea.

PR #1106, opened 2020-11-25
Update guidance and change link to public URL

dturner:


docs/AndroidAudioHistory.md
```diff
@@ -4,7 +4,7 @@ Android audio history
...
+- Bug in **AAudio** on RQ1A (Oboe works around this issue from version 1.5 onwards). A stream is normally disconnected when a **headset is plugged in** or because of other device changes. An `AAUDIO_ERROR_DISCONNECTED` error code should be passed to the error callback. But a bug in Shared MMAP streams causes `AAUDIO_ERROR_TIMEOUT` to be returned. So if your error callback is checking for `AAUDIO_ERROR_DISCONNECTED` then it may not respond properly. We recommend **always stopping and closing the stream** if the error code is `AAUDIO_ERROR_DISCONNECTED` or `AAUDIO_ERROR_TIMEOUT`. Oboe does this. So if you are using Oboe callbacks you are OK. This issue was not in the original R release. It was introduced in RQ1A, which is being delivered by OTA starting in November 2020. It will be fixed in a future update. Follow it on this public Android issue [b/173928197](https://issuetracker.google.com/173928197).
```
dturner, line 7:
```suggestion
- Bug in **AAudio** on RQ1A (Oboe works around this issue from version 1.5 onwards). A stream is normally disconnected when a **headset is plugged in** or because of other device changes. An `AAUDIO_ERROR_DISCONNECTED` error code should be passed to the error callback. But a bug in Shared MMAP streams causes `AAUDIO_ERROR_TIMEOUT` to be returned. So if your error callback is checking for `AAUDIO_ERROR_DISCONNECTED` then it may not respond properly. We recommend **always stopping and closing the stream** regardless of the error code. Oboe does this. So if you are using Oboe callbacks you are OK. This issue was not in the original R release. It was introduced in RQ1A, which is being delivered by OTA starting in November 2020. It will be fixed in a future update. Follow it on this public Android issue [b/173928197](https://issuetracker.google.com/173928197).

```

dturner:
Abandoning because after editing the file in github it wasn't easy to see the changes any more. Not exactly sure why.

PR #1109, opened 2020-11-25
Added note about higher latency in Java

dturner:
AFAIK Java still has ~20ms greater latency than Oboe. I think it's important to mention this since it's one of the main reasons for Oboe's existence.

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -4,8 +4,8 @@
 
 Oboe is a native library written in C++ which uses the Android NDK. To move data from Java to C++ you can use [JNI](https://developer.android.com/training/articles/perf-jni). 
 
-That said, if you are generating your audio in Java you'll get better performance using the [Java AudioTrack class](https://developer.android.com/reference/android/media/AudioTrack). This can be 
-created with low latency using the AudioTrack.Builder method [`setPerformanceMode(AudioTrack.PERFORMANCE_MODE_LOW_LATENCY)`](https://developer.android.com/reference/android/media/AudioTrack#PERFORMANCE_MODE_LOW_LATENCY).
+That said, if you are generating your audio in Java you'll get better computational performance using the [Java AudioTrack class](https://developer.android.com/reference/android/media/AudioTrack). This can be 
+created with low latency using the AudioTrack.Builder method [`setPerformanceMode(AudioTrack.PERFORMANCE_MODE_LOW_LATENCY)`](https://developer.android.com/reference/android/media/AudioTrack#PERFORMANCE_MODE_LOW_LATENCY). Note that this will give you slightly higher latency (roughly 20ms) than if you were to use Oboe. 
 
 You can dynamically tune the latency of the stream just like in Oboe using [`setBufferSizeInFrames(int)`](https://developer.android.com/reference/android/media/AudioTrack.html#setBufferSizeInFrames(int))
 Also you can use blocking writes with the Java AudioTrack and still get a low latency stream.
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -1,11 +1,10 @@
 # Frequently Asked Questions (FAQ)
 
-## Can I write audio data from Java to Oboe?
+## Can I write audio data from Java/Kotlin to Oboe?
 
 Oboe is a native library written in C++ which uses the Android NDK. To move data from Java to C++ you can use [JNI](https://developer.android.com/training/articles/perf-jni). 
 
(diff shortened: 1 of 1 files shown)


docs/FAQ.md
```diff
@@ -4,8 +4,8 @@
 
 Oboe is a native library written in C++ which uses the Android NDK. To move data from Java to C++ you can use [JNI](https://developer.android.com/training/articles/perf-jni). 
 
-That said, if you are generating your audio in Java you'll get better performance using the [Java AudioTrack class](https://developer.android.com/reference/android/media/AudioTrack). This can be 
-created with low latency using the AudioTrack.Builder method [`setPerformanceMode(AudioTrack.PERFORMANCE_MODE_LOW_LATENCY)`](https://developer.android.com/reference/android/media/AudioTrack#PERFORMANCE_MODE_LOW_LATENCY).
+That said, if you are generating your audio in Java you'll get better computational performance using the [Java AudioTrack class](https://developer.android.com/reference/android/media/AudioTrack). This can be
```
philburk, line 7: If you are generating the audio in Java then you have two options for playback. You can use Java AudioTrack or pass the audio to native code using the JNI and then use Oboe. In both cases the computational performance will be the same. If your device supports MMAP then you can get a low latency stream and may have lower latency than when using Java AudioTrack. But if your device does not support MMAP then AAudio Legacy will not provide a LowLatency stream because you are not using a callback. In that case you may get lower latency with Java AudioTrack than with Oboe.

dturner: Gees, that's complicated! I wonder if it'd be better to move this to a new question: "Does Oboe provide lower latency than Java?" and provide this level of detail there. 

I'll leave this PR open for further discussion.

dturner: AI: Add new question "If I'm already generating audio data in Java using blocking writes, will Oboe give me any lower latency?"

porames0025 (approved):

docs/FAQ.md
dturner: @philburk just trying to get this one closed out. You said "in both cases the computational performance will be the same", however, if you're generating audio data in Java you'll be using a thread with userspace priority, however, if you use Oboe + callback you'll get a SCHED_FIFO thread which may lead to improved computational performance due to less chance of being pre-empted. Is my understanding correct, or am I missing something about AudioTrack?

philburk: When you do the computation in Java then there is no callback. The data is just written using blocking writes either in Java or in Oboe.

By "computational performance" I mean the number of instructions (multiplies, adds, etc.).  SCHED_FIFO does not change computation. It only changes the scheduling for when your task gets to run on a CPU.

This is complicated. A Tech Note is in order.

philburk: >  "If I'm already generating audio data in Java using blocking writes, will Oboe give me any lower latency?"

If the device supports MMAP then yes. If not then Java might be lower latency. We should measure this.

dturner: > It only changes the scheduling for when your task gets to run on a CPU.

True. But on a real-world Android device, with multiple competing processes, an app which has high computational load may be able to use lower buffer sizes using Oboe vs Java blocking writes simply because they can obtain more CPU time. 

Maybe the whole Java blocking write vs Oboe callback is a misguided comparison. I don't know any real-time app which uses blocking writes (maybe Caustic?).

robertwu1: I think the note is getting a bit long with too many exceptions. "Oboe reduces latency, particularly on higher end devices. An alternative is AudioTrack.Builder method [`setPerformanceMode(AudioTrack.PERFORMANCE_MODE_LOW_LATENCY)`](https://developer.android.com/reference/android/media/AudioTrack#PERFORMANCE_MODE_LOW_LATENCY) if you are generating your audio in Java."

Many application developers don't need to know about the intrinsics of MMAP vs AAudio Legacy. Phil's suggestion of a Tech Note would be great to link for those who are curious.

dturner:
```suggestion
Oboe reduces latency, particularly on higher end devices. An alternative is AudioTrack.Builder method [setPerformanceMode(AudioTrack.PERFORMANCE_MODE_LOW_LATENCY)](https://developer.android.com/reference/android/media/AudioTrack#PERFORMANCE_MODE_LOW_LATENCY) for those generating audio in Java.
```

dturner: Great suggestion. I have updated the text accordingly.

robertwu1 (approved):

PR #1117, opened 2020-11-26
Update API reference documentation to 1.5

dturner:


philburk:
Is this fixed?

dturner:
@philburk Not sure what was going on here but yes, the docs are updated to v1.5.

PR #1133, opened 2020-12-10
Add DataPaths test for Speakers and Mics

philburk:
Plays a sine wave through single channels and analyzes the recording
on single channels.
This is designed to check for dead channels.


dturner:
Hey Phil, I was able to run OboeTester with the "Auto Glitch Test" on my Pixel 3. Every test fails but I'm not sure whether that's the expected behaviour.

philburk:
>  I was able to run OboeTester with the "Auto Glitch Test" on my Pixel 3. Every test fails...

That is not expected. There is a known problem on Pixel 3a with a single dead channel for the EARPIECE and the SPEAKER_SAFE.

What version of Android were you using?

Your volume should be turned up to about 80%.

Did you hear the tone for each test? If so then the input side is not working.

apps/OboeTester/app/src/main/java/com/google/sample/oboe/manualtest/AutomatedGlitchActivity.java
```diff
@@ -0,0 +1,111 @@
...
+
+        testConfigurations();
+    }
+
+    private void testConfiguration(int performanceMode,
+                                   int sharingMode,
+                                   int sampleRate) throws InterruptedException {
+        testConfiguration(performanceMode,
+                sharingMode,
+                sampleRate, 1, 2);
```
dturner, line 85: Magic numbers. Could you use `StreamConfiguration.PERFORMANCE_MODE_LOW_LATENCY` and `SampleRate.16000Hz`?

apps/OboeTester/app/src/main/java/com/google/sample/oboe/manualtest/AutomatedGlitchActivity.java
```diff
@@ -0,0 +1,111 @@
...
+                sharingMode,
+                sampleRate, 2, 1);
+    }
+
+    @Override
+    public void runTest() {
+        try {
+            testConfiguration(StreamConfiguration.PERFORMANCE_MODE_LOW_LATENCY,
+                    StreamConfiguration.SHARING_MODE_EXCLUSIVE,
+                    0);
```
dturner, line 96: As above, magic number.

apps/OboeTester/app/src/main/java/com/google/sample/oboe/manualtest/StreamConfiguration.java
```diff
@@ -146,26 +146,27 @@ public void setPerformanceMode(int performanceMode) {
...
     }
 
-    public int getInputPreset() {
-        return mInputPreset;
-    }
-    public void setInputPreset(int inputPreset) {
-        this.mInputPreset = inputPreset;
-    }
-
     static String convertPerformanceModeToText(int performanceMode) {
```
dturner, line 149: Couldn't you use `oboe::convertToText` for this?

apps/OboeTester/app/src/main/java/com/google/sample/oboe/manualtest/StreamConfiguration.java
```diff
@@ -177,11 +178,11 @@ public void setSharingMode(int sharingMode) {
     static String convertSharingModeToText(int sharingMode) {
```
dturner, line 178: Couldn't you use `oboe::convertToText` for this?

dturner (approved):
Had a quick scan through and made some comments. Looks like this is a really great addition to our test suite. LGTM with minor nits.

apps/OboeTester/app/src/main/java/com/google/sample/oboe/manualtest/StreamConfiguration.java
philburk: It would be more work to call through JNI than to just add the Java code. Also I may want shorter text.

apps/OboeTester/app/src/main/java/com/google/sample/oboe/manualtest/StreamConfiguration.java
philburk: ditto

apps/OboeTester/app/src/main/java/com/google/sample/oboe/manualtest/AutomatedGlitchActivity.java
philburk: The 1 and 2 are channels. I could use MONO and STEREO to be more clear. Thanks.

philburk: Done

apps/OboeTester/app/src/main/java/com/google/sample/oboe/manualtest/AutomatedGlitchActivity.java
philburk: Done

PR #1140, opened 2020-12-16
fix: [android sample] fix a crash when calling LiveEffectEngine.setEf…

zhenyuni:
fix: [android sample] fix a crash when calling LiveEffectEngine.setEffectOn(false) in MainActivity.onPause()

### DIFF
--- a/samples/LiveEffect/src/main/cpp/FullDuplexStream.cpp
+++ b/samples/LiveEffect/src/main/cpp/FullDuplexStream.cpp
@@ -112,6 +112,8 @@ oboe::Result FullDuplexStream::stop() {
     if (mInputStream) {
         inputResult = mInputStream->requestStop();
     }
+    setOutputStream(nullptr);
+    setInputStream(nullptr);
     if (outputResult != oboe::Result::OK) {
         return outputResult;
     } else {


samples/LiveEffect/src/main/cpp/FullDuplexStream.cpp
```diff
@@ -112,6 +112,8 @@ oboe::Result FullDuplexStream::stop() {
     if (mInputStream) {
         inputResult = mInputStream->requestStop();
     }
+    setOutputStream(nullptr);
```
philburk, line 115: This would prevent restarting the FullDuplexStream.

philburk (changes requested):
There is no Issue filed for this fix. 
Please create an Issue and explain how to reproduce the crash. 
What devices and OS does it crash on?

samples/LiveEffect/src/main/cpp/FullDuplexStream.cpp
zhenyuni: #1143 
The inputStream and outputStream will be assigned again in LiveEffectEngine::openStreams when restarting.

philburk: Yes, the app will reset them again when you push the button.
But some people will use this class by itself in their own apps.
If you clear the streams then calling FullDuplexStream::start() again will fail.

zhenyuni: I am sorry I wasn't running the latest code. This problem has been fixed in 00039f7 by removing destructor of LiveEffectEngine. It works fine now.

PR #1155, opened 2021-01-06
fxlab: update gradle

philburk:
Fixes broken build:
https://issuetracker.google.com/issues/175895275


android-usb (approved):
This seems to work.

PR #1181, opened 2021-02-01
Suppress warnings of unused variable.

jimmytobin2425:
To suppress the undefined variable error when the LOGD line is not compiled in, we can add in an ignore command. Similar to discussion here: https://herbsutter.com/2009/10/18/mailbag-shutting-up-compiler-warnings/


philburk (approved):
Thanks!

PR #1200, opened 2021-02-25
oboe: fix possible race on close

### LINKED ISSUE #1180: Thread safety gaps in AudioStreamAAudio
There are numerous method, eg. getTimestamp(), that reference the underlying native audio stream.
But that stream could get closed by an AAudio error callback that happened simultaneously.

    // <-- If close() occurred here then the stream would get used but the pointer is invalid!
    if (stream != nullptr) {

We need to lock around references to mAAudioStream.

philburk:
Protect native stream pointer with a shared_lock.

The race may have never ended badly.
But it was theoretically possible. So this prevents it.

Note that this requires that Oboe be built with C++17.
But apps can still use C++14.

Fixes #1180

### DIFF
--- a/apps/OboeTester/app/src/main/cpp/NativeAudioContext.cpp
+++ b/apps/OboeTester/app/src/main/cpp/NativeAudioContext.cpp
@@ -14,11 +14,18 @@
  * limitations under the License.
  */
 
+// Set to 1 for debugging race condition #1180 with mAAudioStream.
+// See also AudioStreamAAudio.cpp in Oboe.
+#define DEBUG_CLOSE_RACE 0
+
 #include <fstream>
 #include <iostream>
+#if DEBUG_CLOSE_RACE
+#include <thread>
+#endif // DEBUG_CLOSE_RACE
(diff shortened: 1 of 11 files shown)


apps/OboeTester/app/src/main/cpp/NativeAudioContext.cpp
```diff
@@ -251,6 +258,20 @@ oboe::Result ActivityContext::start() {
         dataThread = new std::thread(threadCallback, this);
     }
 
+#ifdef DEBUG_CLOSE_RACE
```
dturner, line 261: If we are leaving this in, please add a comment explaining exactly what this directive is used for.

dturner (approved):
See comment

apps/OboeTester/app/src/main/cpp/NativeAudioContext.cpp
philburk: Done.

dturner (approved):

PR #1214, opened 2021-03-05
Fixed (potential) memory leak in WavStreamReader

### LINKED ISSUE #1213: Apparent memory leak in WavStreamReader
WavStreamReader allocates various chunks as raw pointers here:

https://github.com/google/oboe/blob/108a2cd854cc298e348b73a2638abd42bc1a6c32/samples/parselib/src/main/cpp/wav/WavStreamReader.cpp#L83

But I do not see where they are deleted!

They should be declared as std::unique_ptr.
Or, since they are small and are generally all needed when parsing, they could just be declared directly in place instead of being allocated at run-time.

android-usb:
Fixes #1213

### DIFF
--- a/samples/parselib/src/main/cpp/wav/WavStreamReader.cpp
+++ b/samples/parselib/src/main/cpp/wav/WavStreamReader.cpp
@@ -38,8 +38,6 @@ WavStreamReader::WavStreamReader(InputStream *stream) {
     mDataChunk = 0;
 
     mAudioDataStartPos = -1;
-
-//    mChunkMap = new std::map<RiffID, std::shared_ptr<WavChunkHeader>>();
 }
 
 int WavStreamReader::getSampleEncoding() {
--- a/samples/parselib/src/main/cpp/wav/WavStreamReader.cpp
+++ b/samples/parselib/src/main/cpp/wav/WavStreamReader.cpp
@@ -39,7 +39,7 @@ WavStreamReader::WavStreamReader(InputStream *stream) {
 
     mAudioDataStartPos = -1;
 
-    mChunkMap = new std::map<RiffID, WavChunkHeader *>();
+    mChunkMap = new std::map<RiffID, std::shared_ptr<WavChunkHeader>>();
 }
 
 int WavStreamReader::getSampleEncoding() {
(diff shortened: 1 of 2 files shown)


samples/parselib/src/main/cpp/wav/WavStreamReader.h
```diff
@@ -59,13 +59,13 @@ class WavStreamReader {
...
-    WavFmtChunkHeader *mFmtChunk;
-    WavChunkHeader *mDataChunk;
+    std::shared_ptr<WavRIFFChunkHeader> mWavChunk;
+    std::shared_ptr<WavFmtChunkHeader> mFmtChunk;
+    std::shared_ptr<WavChunkHeader> mDataChunk;
 
     long mAudioDataStartPos;
 
-    std::map<RiffID, WavChunkHeader *> *mChunkMap;
+    std::map<RiffID, std::shared_ptr<WavChunkHeader>> *mChunkMap;
```
philburk, line 68: I don't see where this map is used. Can we remove it?
Then we could use std::unique_ptr or, better yet, just declare the Chunks and not use new.

samples/parselib/src/main/cpp/wav/WavStreamReader.cpp
```diff
@@ -39,7 +39,7 @@ WavStreamReader::WavStreamReader(InputStream *stream) {
 
     mAudioDataStartPos = -1;
 
-    mChunkMap = new std::map<RiffID, WavChunkHeader *>();
+    mChunkMap = new std::map<RiffID, std::shared_ptr<WavChunkHeader>>();
```
philburk, line 42: Where does this get deleted?

philburk (changes requested):

samples/parselib/src/main/cpp/wav/WavStreamReader.h
android-usb: It's there so that the WavStreamReader can evolve and also chunks that it doesn't recognize. I'd like to leave it and see how WavStreamReader evolves moving forward.

samples/parselib/src/main/cpp/wav/WavStreamReader.cpp
android-usb: More to the point, why bother allocating it dynamically in any case?

samples/parselib/src/main/cpp/wav/WavStreamReader.cpp
```diff
@@ -39,7 +39,7 @@ WavStreamReader::WavStreamReader(InputStream *stream) {
 
     mAudioDataStartPos = -1;
 
-    mChunkMap = new std::map<RiffID, WavChunkHeader *>();
+//    mChunkMap = new std::map<RiffID, std::shared_ptr<WavChunkHeader>>();
```
philburk, line 42: Dead code.

philburk (approved):
Thanks for fixing this.

samples/parselib/src/main/cpp/wav/WavStreamReader.cpp
android-usb: Done.

philburk (approved):
When you submit the changes, please do a Squash and Merge from the GitHub UI.

PR #1224, opened 2021-03-15
RhythmGame FFMpegExtractor - for clearing packet use av_packet_unref

ashfaq1701:
Currently in FFMpegExtractor for clearing out packets the following two lines are being used,

```
avPacket.size = 0;
avPacket.data = nullptr;
```

This might create memory leak if someone tries to reuse this implementation in some other project. I am writing an application which is reading data from device storage and decoding using FFMpeg. I did found this implementation very compact, organised and ready to be used. But when I used it and profiled memory, I got below,

Screen Shot 2564-03-16 at 01 59 20

Memory block equal to the same size of the decompressed audio remains as ghost memory in RAM.

Doing with the appropriate function,

```
av_packet_unref(&avPacket);
```

Does not leave any memory leak like below,

Screen Shot 2564-03-16 at 02 03 27

I know this is a very small contribution, but I hope it will be merged. :)


porames0025 (approved):

dturner:
Thank you! This is a great find.

PR #1245, opened 2021-04-15
Use an array of callback buffer for OpenSL ES.

### LINKED ISSUE #1165: OpenSL ES should use an array of callback buffers
Android version(s): all
Android device(s): all
Oboe version: 1.5

**Short description**
In https://b.corp.google.com/issues/154868033#comment9 Glenn Kasten wrote:

  AudioStreamOpenSLES.h:    constexpr int kBufferQueueLength = 2; // double buffered for callbacks
but:
  AudioStreamOpenSLES.h:    std::unique_ptr<uint8_t[]>    mCallbackBuffer;
To actually take advantage of kBufferQueueLength > 1, we'll need to change mCallbackBuffer to be an array.
Note that bqCallbackGlue() is called after each buffer is consumed, but does not indicate which buffer was consumed.
You need to keep track of that yourself.  For example, for kBufferQueueLength = 2, then:
  fill a with data [which could be zeroes initially]
  Enqueue(a)
  fill b with data [which could be zeroes initially]
  Enqueue(b)
  bqCallbackGlue() is called when a is consumed
  fill a with data
  Enqueue(a)
  bqCallbackGlue() is called when b is consumed
  fill b with data
  Enqueue(b)
  etc.

flamme:
Fixes #1165

### DIFF
--- a/src/opensles/AudioInputStreamOpenSLES.cpp
+++ b/src/opensles/AudioInputStreamOpenSLES.cpp
@@ -273,11 +273,9 @@ Result AudioInputStreamOpenSLES::requestStart() {
     Result result = setRecordState_l(SL_RECORDSTATE_RECORDING);
     if (result == Result::OK) {
         setState(StreamState::Started);
-        // Enqueue the first `kBufferQueueLength` buffer to start the streaming.
+        // Enqueue the first buffer to start the streaming.
         // This does not call the callback function.
-        for (size_t i = 0; i < kBufferQueueLength; ++i) {
-            enqueueCallbackBuffer(mSimpleBufferQueueInterface);
-        }
+        enqueueCallbackBuffer(mSimpleBufferQueueInterface);
     } else {
         setState(initialState);
     }
--- a/src/opensles/AudioInputStreamOpenSLES.cpp
+++ b/src/opensles/AudioInputStreamOpenSLES.cpp
@@ -273,9 +273,11 @@ Result AudioInputStreamOpenSLES::requestStart() {
     Result result = setRecordState_l(SL_RECORDSTATE_RECORDING);
     if (result == Result::OK) {
         setState(StreamState::Started);
-        // Enqueue the first buffer to start the streaming.
+        // Enqueue the first `kBufferQueueLength` buffer to start the streaming.
         // This does not call the callback function.
-        enqueueCallbackBuffer(mSimpleBufferQueueInterface);
+        for (size_t i = 0; i < kBufferQueueLength; ++i) {
+            enqueueCallbackBuffer(mSimpleBufferQueueInterface);
+        }
     } else {
         setState(initialState);
     }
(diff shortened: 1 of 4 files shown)


tguilbert-google:
I am deferring to philburk@ for the review.

src/opensles/AudioStreamOpenSLES.h
```diff
@@ -121,7 +121,11 @@ class AudioStreamOpenSLES : public AudioStreamBuffered {
     MonotonicCounter              mPositionMillis; // for tracking OpenSL ES service position
 
 private:
-    std::unique_ptr<uint8_t[]>    mCallbackBuffer;
+    uint8_t* getBufferToProcess(int* bufferIdx);
+
+    std::unique_ptr<uint8_t[]>    mCallbackBuffer[kBufferQueueLength];
+    int                           mCallbackBufferIdxApp = 0;
```
philburk, line 127: Change Idx to "Index".

philburk (changes requested):

src/opensles/AudioInputStreamOpenSLES.cpp
```diff
@@ -273,9 +273,11 @@ Result AudioInputStreamOpenSLES::requestStart() {
     Result result = setRecordState_l(SL_RECORDSTATE_RECORDING);
     if (result == Result::OK) {
         setState(StreamState::Started);
-        // Enqueue the first buffer to start the streaming.
+        // Enqueue the first `kBufferQueueLength` buffer to start the streaming.
         // This does not call the callback function.
-        enqueueCallbackBuffer(mSimpleBufferQueueInterface);
+        for (size_t i = 0; i < kBufferQueueLength; ++i) {
```
philburk, line 278: Only enqueue ONE buffer because that will start the callbacks.
If you enqueue more buffers from this thread then they may collide with the enqueuing from the callback thread.

src/opensles/AudioOutputStreamOpenSLES.cpp
```diff
@@ -302,9 +302,11 @@ Result AudioOutputStreamOpenSLES::requestStart() {
         setState(StreamState::Started);
         mLock.unlock();
         if (getBufferDepth(mSimpleBufferQueueInterface) == 0) {
-            // Enqueue the first buffer if needed to start the streaming.
+            // Enqueue the first `kBufferQueueLength` buffer if needed to start the streaming.
             // This might call requestStop() so try to avoid a recursive lock.
-            processBufferCallback(mSimpleBufferQueueInterface);
+            for (size_t i = 0; i < kBufferQueueLength; ++i) {
```
philburk, line 307: ditto

philburk (changes requested):
I think this is the right idea given the API for OpenSL ES.
The current implementation is immediately writing to the AudioTrack buffer.
So the old code was able to run without glitching.
But having one memory buffer per node is more correct.

philburk (approved):
Feel free to Squash and Merge any time.

PR #1255, opened 2021-04-30
Add support for 24 and 32-bit sample formats

philburk:
This is only supported in S or later.

### DIFF
--- a/apps/OboeTester/app/src/main/cpp/FormatConverterBox.h
+++ b/apps/OboeTester/app/src/main/cpp/FormatConverterBox.h
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2021 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef OBOETESTER_FORMATCONVERTERBOX_H
+#define OBOETESTER_FORMATCONVERTERBOX_H
+
+
+#include <unistd.h>
+#include <sys/types.h>
+
+#include "oboe/Oboe.h"
+#include "flowgraph/SinkFloat.h"
+#include "flowgraph/SinkI16.h"
+#include "flowgraph/SinkI24.h"
+#include "flowgraph/SinkI32.h"
+#include "flowgraph/SourceFloat.h"
+#include "flowgraph/SourceI16.h"
+#include "flowgraph/SourceI24.h"
+#include "flowgraph/SourceI32.h"
+
+/**
+ * Use flowgraph modules to convert between the various data formats.
+ */
+
+class FormatConverterBox {
+public:
+    FormatConverterBox(int32_t numSamples, oboe::AudioFormat inputFormat,
+                       oboe::AudioFormat outputFormat);
+
+    void *getOutputBuffer() {
+        return (void *) mOutputBuffer.get();
+    };
+
+    void *getInputBuffer() {
+        return (void *) mInputBuffer.get();
+    };
+
+    /** Convert the data in the conversion buffer to float.
+     */
+    int32_t convert(int32_t numSamples);
+    int32_t convert(int32_t numSamples, const void *inputBuffer);
+    int32_t convert(void *outputBuffer, int32_t numSamples);
+    int32_t convert(void *outputBuffer, int32_t numSamples, const void *inputBuffer);
+
+private:
+    oboe::AudioFormat mInputFormat{oboe::AudioFormat::Invalid};
+    oboe::AudioFormat mOutputFormat{oboe::AudioFormat::Invalid};
+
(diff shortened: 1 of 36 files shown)


apps/OboeTester/app/src/main/cpp/FormatConverterBox.h
```diff
@@ -0,0 +1,70 @@
...
+
+    void *getOutputBuffer() {
+        return (void *) mOutputBuffer.get();
+    };
+
+    void *getInputBuffer() {
+        return (void *) mInputBuffer.get();
+    };
+
+    /** Convert the data in the conversion buffer to float.
```
philburk, line 51: Fix comment.

apps/OboeTester/app/src/main/cpp/FullDuplexStream.cpp
```diff
@@ -112,10 +123,13 @@ oboe::Result FullDuplexStream::start() {
     // Determine maximum size that could possibly be called.
     int32_t bufferSize = getOutputStream()->getBufferCapacityInFrames()
             * getOutputStream()->getChannelCount();
-    if (bufferSize > mBufferSize) {
-        mInputBuffer = std::make_unique<float[]>(bufferSize);
-        mBufferSize = bufferSize;
-    }
+    mInputConverter = std::make_unique<FormatConverterBox>(bufferSize,
+    getInputStream()->getFormat(),
```
philburk, line 127: indent

include/oboe/Definitions.h
```diff
@@ -111,6 +110,17 @@ namespace oboe {
          * Single precision floating points.
          */
         Float = 2, // AAUDIO_FORMAT_PCM_FLOAT,
+
+        /**
+         * Signed 24-bit integers, packed into 3 bytes.
```
philburk, line 115: Add note that these are added in S.

src/common/SourceI24Caller.h
```diff
@@ -0,0 +1,51 @@
...
+    : AudioSourceCaller(channelCount, framesPerCallback, kBytesPerI24Packed) {
+        mConversionBuffer = std::make_unique<uint8_t[]>(
+                kBytesPerI24Packed * channelCount * output.getFramesPerBuffer());
+    }
+
+    int32_t onProcess(int32_t numFrames) override;
+
+    const char *getName() override {
+        return "SourceI24Caller";
+    }
```
philburk, line 43: Add empty line.

apps/OboeTester/app/src/main/cpp/FormatConverterBox.h
dturner: Shouldn't this be "Convert the data in the conversion buffer to the specified format"?

apps/OboeTester/app/src/main/cpp/FormatConverterBox.h
```diff
@@ -0,0 +1,70 @@
...
+     */
+    int32_t convert(int32_t numSamples);
+    int32_t convert(int32_t numSamples, const void *inputBuffer);
+    int32_t convert(void *outputBuffer, int32_t numSamples);
+    int32_t convert(void *outputBuffer, int32_t numSamples, const void *inputBuffer);
```
dturner, line 56: IIUC `FormatConverterBox` can either convert its internal buffers, or the caller can supply their own buffers for conversion. With that in mind, it's tough to tell at a glance what each of these `convert` methods does, how they differ from each other and the order of arguments required to call the desired method. Maybe they could be renamed to `convertInternal`, `convertInput`, `convertOutput` or something similar.

apps/OboeTester/app/src/main/cpp/FullDuplexStream.cpp
```diff
@@ -112,10 +123,13 @@ oboe::Result FullDuplexStream::start() {
...
+    oboe::AudioFormat::Float);
+    mOutputConverter = std::make_unique<FormatConverterBox>(bufferSize,
+            oboe::AudioFormat::Float,
+            getOutputStream()->getFormat());
+
```
dturner, line 132: I'm sure there's a really obvious answer here but why can't the conversion be done in a single stage, rather than having one converter for the input and one for the output with Float being the intermediate format?

apps/OboeTester/app/src/main/cpp/InputStreamCallbackAnalyzer.cpp
```diff
@@ -22,35 +22,19 @@ oboe::DataCallbackResult InputStreamCallbackAnalyzer::onAudioReady(
...
-                float sample = frameData[iChannel];
-                mPeakDetectors[iChannel].process(sample);
-            }
-            frameData += channelCount;
+    mInputConverter->convert(numFrames * channelCount, audioData);
+    float *floatData = (float *) mInputConverter->getOutputBuffer();
+    if (mRecording != nullptr) {
+        mRecording->write(floatData, numFrames);
+    }
+    float *frameData = floatData;
```
dturner, line 31: It seems like `floatData` isn't modified anywhere else, so is there a need for `frameData`?

dturner (approved):
The core code looks great (although I haven't verified the format conversion algorithms - that's much more your domain), made a few comments/suggestions on the OboeTester code which you can take or leave since we have less stringent code standards for OboeTester.

apps/OboeTester/app/src/main/cpp/FullDuplexStream.cpp
philburk: Because I wanted to do the full duplex processing in Float and the input or output streams may be a different format. So I need to process on each end.

apps/OboeTester/app/src/main/cpp/FormatConverterBox.h
philburk: Good idea.

PR #1266, opened 2021-05-13
Fix failing unit tests on pre-L devices

dturner:
I've also split the unit test fixtures into `StreamOpenOutput` and `StreamOpenInput` so that it's easy to run these tests in isolation using: 

```
testOboe --gtest_filter=StreamOpenOutput.*
```

This is useful when testing on the emulator which doesn't allow input streams to be opened, and thus causes all the associated tests to fail.

### DIFF
--- a/tests/testStreamOpen.cpp
+++ b/tests/testStreamOpen.cpp
@@ -38,6 +38,7 @@ class StreamOpen : public ::testing::Test {
 protected:
 
     bool openStream() {
+        mStream = nullptr;
         Result r = mBuilder.openStream(&mStream);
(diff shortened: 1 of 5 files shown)


tests/testStreamOpen.cpp
```diff
@@ -38,6 +38,7 @@ class StreamOpen : public ::testing::Test {
 protected:
 
     bool openStream() {
+        mStream = nullptr;
```
philburk, line 41: It might be worth doing EXPECT_NULL(mStream) in a later PR.

philburk (approved):

PR #1267, opened 2021-05-13
Support float on jellybean

dturner:
Add quirk to allow float playback streams to be created on pre-L devices by using format conversion.

### DIFF
--- a/src/common/QuirksManager.cpp
+++ b/src/common/QuirksManager.cpp
@@ -184,6 +184,17 @@ bool QuirksManager::isConversionNeeded(
         LOGI("QuirksManager::%s() forcing internal format to I16 for low latency", __func__);
     }
 
+    // Add quirk for float output on API <21
+    if (isFloat
+            && !isInput
+            && getSdkVersion() < __ANDROID_API_L__ && builder.isFormatConversionAllowed()) {
+        childBuilder.setFormat(AudioFormat::I16);
+        conversionNeeded = true;
+        LOGI("QuirksManager::%s() float was requested but not supported on pre-L devices, "
+             "creating an underlying I16 stream and using format conversion to provide a float "
+             "stream", __func__);
+    }
+
     // Channel Count conversions
     if (OboeGlobals::areWorkaroundsEnabled()
             && builder.isChannelConversionAllowed()
--- a/src/common/QuirksManager.cpp
+++ b/src/common/QuirksManager.cpp
@@ -187,7 +187,9 @@ bool QuirksManager::isConversionNeeded(
     // Add quirk for float output on API <21
     if (isFloat
             && !isInput
-            && getSdkVersion() < __ANDROID_API_L__ && builder.isFormatConversionAllowed()) {
+            && getSdkVersion() < __ANDROID_API_L__
+            && builder.isFormatConversionAllowed()
+            ) {
         childBuilder.setFormat(AudioFormat::I16);
         conversionNeeded = true;
         LOGI("QuirksManager::%s() float was requested but not supported on pre-L devices, "
--- a/tests/testStreamOpen.cpp
+++ b/tests/testStreamOpen.cpp
@@ -38,7 +38,7 @@ class StreamOpen : public ::testing::Test {
 protected:
 
     bool openStream() {
-        mStream = nullptr;
+        EXPECT_EQ(mStream, nullptr);
         Result r = mBuilder.openStream(&mStream);
         EXPECT_EQ(r, Result::OK) << "Failed to open stream " << convertToText(r);
         EXPECT_EQ(0, openCount) << "Should start with a fresh object every time.";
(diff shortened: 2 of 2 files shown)


src/common/QuirksManager.cpp
```diff
@@ -184,6 +184,17 @@ bool QuirksManager::isConversionNeeded(
         LOGI("QuirksManager::%s() forcing internal format to I16 for low latency", __func__);
     }
 
+    // Add quirk for float output on API <21
+    if (isFloat
+            && !isInput
```
philburk, line 189: Is Float supported for input?!

src/common/QuirksManager.cpp
```diff
@@ -184,6 +184,17 @@ bool QuirksManager::isConversionNeeded(
...
+            && getSdkVersion() < __ANDROID_API_L__ && builder.isFormatConversionAllowed()) {
```
philburk, line 190: I think it is more clear to have each && on its own line.

tests/testStreamOpen.cpp
```diff
@@ -331,6 +331,17 @@ TEST_F(StreamOpenOutput, PlaybackFormatFloatReturnsErrorBeforeLollipop){
     }
 }
 
+TEST_F(StreamOpenOutput, PlaybackFormatFloatReturnsFloatBeforeLollipopWithFormatConversionAllowed){
+    if (getSdkVersion() < __ANDROID_API_L__){
```
philburk, line 335: We might well allow this test to run on every version. It should pass on every version.

tests/testStreamOpen.cpp
```diff
@@ -331,6 +331,17 @@ TEST_F(StreamOpenOutput, PlaybackFormatFloatReturnsErrorBeforeLollipop){
...
+        mBuilder.setDirection(Direction::Output);
```
philburk, line 336: Should we also test for Input?

philburk (changes requested):

tests/testStreamOpen.cpp
```diff
@@ -331,6 +331,17 @@ TEST_F(StreamOpenOutput, PlaybackFormatFloatReturnsErrorBeforeLollipop){
     }
 }
 
+TEST_F(StreamOpenOutput, PlaybackFormatFloatReturnsFloatBeforeLollipopWithFormatConversionAllowed){
```
philburk, line 334: PlaybackFormatFloatReturnsFloatWithFormatConversionAllowed always, not just on Lollipop

src/common/QuirksManager.cpp
dturner: I imagine so, however, I'm unable to test that because I don't have a pre-L device and the emulator doesn't support input streams. I figured it was best to only add code which I could test. We (and by that I mean, you :) could always a quirk for input and then test it on that old phone you have.

src/common/QuirksManager.cpp
dturner: ha, I do too, for some reason I just missed the last one - good spot, thanks.

tests/testStreamOpen.cpp
dturner: That's a broader scoped test though, since on L+ a float stream should be returned regardless of whether format conversion is allowed. This test is specifically for pre-L.

tests/testStreamOpen.cpp
dturner: IIRC float input is only supported on Marshmallow (API 23+), so that would require a different test.

tests/testStreamOpen.cpp
dturner: See previous comments.

tests/testStreamOpen.cpp
philburk: I don't understand why the conditional is needed. The test should pass on every version of Android including pre-L. But OK.

src/common/QuirksManager.cpp
philburk: OK, I will add that and test it on my device.

tests/testStreamOpen.cpp
philburk: We should add a second test that does the exact same thing but for SDK >= L.

tests/testStreamOpen.cpp
philburk: Right. I can add a Quirk for that later.

philburk (approved):

tests/testStreamOpen.cpp
dturner: Yeah, actually you're right. I was thinking about it simply from the point of view that pre-L devices *must* have format conversion enabled to return a float stream, whereas L+ devices it shouldn't matter. There could be a scenario where future Android versions default to a different format (`I32`?) in which case this test should cover that code path. I will address in a follow up CL (reminder: #1270).

Issue #1239, opened 2021-03-31
Engine stops with ErrorTimeout on Samsung A21s when BufferCapacity is specified


apleton:
Android version(s): 10 
Android device(s): Samsung A21s
Oboe version: 1.5.0
App name used for testing: Hello Oboe
 
**Short description**
When I call setBufferCapacityInFrames(8192) on the AudioStreamBuilder in Hello Oboe, the error callback will be called with the error ErrorTimeout (-885). It seems that it's the combination setPerformanceMode(oboe::PerformanceMode::LowLatency) and setBufferCapacityInFrames(8192)  that creates the issue. 

The A21s is the first device where I see that issue. Should I remove setBufferCapacityInFrames call ? What is the best practice here? 

**Steps to reproduce**
- In Samples project, add `->setBufferCapacityInFrames(8192)` to the AudioStreamBuilder in HelloOboeEngine.cpp (method createPlaybackStream). 

```
oboe::Result HelloOboeEngine::createPlaybackStream() {
    oboe::AudioStreamBuilder builder;
    return builder.setSharingMode(oboe::SharingMode::Exclusive)
        ->setPerformanceMode(oboe::PerformanceMode::LowLatency)
        ->setFormat(oboe::AudioFormat::Float)
        ->setDataCallback(mLatencyCallback.get())
        ->setErrorCallback(mErrorCallback.get())
        ->setAudioApi(mAudioApi)
        ->setChannelCount(2)
        ->setDeviceId(mDeviceId)
        ->setFormat(oboe::AudioFormat::Float)
        ->setBufferCapacityInFrames(8192)
        ->openStream(mStream);
}
```

- Launch the HelloOboe app on a Samsung A21s

**Expected behavior**
App emits sound when surface is touched

**Actual behavior**
- No sound when surface is touched
- Error -885 (timeout) printed in logs

**Device**
Samsung A21s

philburk:
@apleton - Thanks for the report. Can you reproduce this in your office? Do you have an A21S?

We would need to see logs from the open call to the error callback. Can you capture some logcat and attach (not paste) it as a file?

If you can get a bugreport file then please create a bug here:

https://issuetracker.google.com/issues?q=componentid:192705

and then attach the bugreport file to that bug.

dturner:
Seems like the A21S has other problems as well: https://github.com/google/oboe/issues/1247

apleton:
Here is the log in the test described in the first post
[...]

PR #1268, opened 2021-05-13
HelloOboe: only open stream once when launched

philburk:
Prevent stream from being opened multiple times when the app is started.
The menu updates were triggering reopen().
Show Toast with error message if open or start fails.

For #1239


dturner (approved):
Thanks for doing all this Phil - looks good.

PR #1273, opened 2021-05-17
Fix calculation related to sample rate conversion for #1256

### LINKED ISSUE #1256: getSampleRate disagrees with getFramesRead for input streams when sample rate conversion is enabled
The calculation here: https://github.com/google/oboe/blob/8041b3891ade1a39dc575fb4e740b6d6f7087d0d/src/common/FilterAudioStream.cpp#L51 makes a ratio that will convert from output to input sample rate. However, `getSampleRate` will always report the requested (converted) sample rate, which is the input sample rate for an output stream, but the output sample rate for an input stream. This means calculations working with `getSampleRate` along with `getFramesRead` on input streams will yield incorrect values. Changing the value of `mRateScaler` to `((double) getSampleRate()) / mChildStream->getSampleRate();` appears to resolve this issue, by setting up a ratio that will consistently convert from the child (native) sample rate to the software (converted) sample rate.

To demonstrate the problem, if we make a function that calculates latency for a stream based on AudioStreamAAudio's calculateLatencyMillis, it will produce implausibly small or negative values for latency when calculating for an input stream that's converting from 48k to 44.1k.

flamme:


philburk (approved):
I tested using the repro steps in #1256 for INPUT and OUTPUT, for 8000 and 96000 Hz and it passed OK.

PR #1276, opened 2021-05-17
Add unit tests for testing getFramesWritten returns correctly.

### LINKED ISSUE #1256: getSampleRate disagrees with getFramesRead for input streams when sample rate conversion is enabled

flamme:
Add unit tests that open input/output stream and process data for a
while and then verify the returned value from getFramesWritten is
correct.

Fixes #1256.

### DIFF
--- a/tests/testStreamFramesWritten.cpp
+++ b/tests/testStreamFramesWritten.cpp
@@ -1,0 +1,88 @@
+/*
+ * Copyright 2021 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <gtest/gtest.h>
+#include <oboe/Oboe.h>
+
+#include <tuple>
+
+using namespace oboe;
+
+class MyCallback : public AudioStreamDataCallback {
+public:
+    DataCallbackResult onAudioReady(AudioStream *oboeStream, void *audioData, int32_t numFrames) override {
+        return DataCallbackResult::Continue;
+    }
+};
+
+using StreamFramesWrittenParams = std::tuple<Direction, int32_t>;
+
+class StreamFramesWritten : public ::testing::Test,
+                            public ::testing::WithParamInterface<StreamFramesWrittenParams> {
+
+protected:
+    void TearDown() override;
+
+    static constexpr int PROCESS_TIME_SECOND = 5;
+
+    AudioStreamBuilder mBuilder;
+    AudioStream *mStream = nullptr;
+};
+
+void StreamFramesWritten::TearDown() {
+    if (mStream != nullptr) {
+        mStream->close();
+        mStream = nullptr;
+    }
+}
+
+TEST_P(StreamFramesWritten, VerifyFramesWritten) {
+    const Direction direction = std::get<0>(GetParam());
+    const int32_t sampleRate = std::get<1>(GetParam());
+
+    AudioStreamDataCallback *callback = new MyCallback();
+    mBuilder.setDirection(direction)
+            ->setFormat(AudioFormat::Float)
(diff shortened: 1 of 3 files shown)


tests/testStreamFramesWritten.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+using StreamFramesWrittenParams = std::tuple<Direction, int32_t>;
+
+class StreamFramesWritten : public ::testing::Test,
+                            public ::testing::WithParamInterface<StreamFramesWrittenParams> {
+
+protected:
+    void TearDown() override;
+
+    static constexpr int PROCESS_TIME_SECOND = 5;
```
philburk, line 39: SECONDS

tests/testStreamFramesWritten.cpp
```diff
@@ -0,0 +1,88 @@
...
+    Result r = mBuilder.openStream(&mStream);
+    ASSERT_EQ(r, Result::OK) << "Failed to open stream." << convertToText(r);
+
+    r = mStream->start();
+    ASSERT_EQ(r, Result::OK) << "Failed to start stream." << convertToText(r);
+    sleep(PROCESS_TIME_SECOND);
+
+    // The frames written should be close to sampleRate * PROCESS_TIME_SECONDS
+    const int64_t framesWritten = mStream->getFramesWritten();
+    EXPECT_NEAR(framesWritten, sampleRate * PROCESS_TIME_SECOND, sampleRate / 2);
```
philburk, line 74: Stream should be stopped and closed to prevent a leak. Closing is not automatic.

tests/testStreamFramesWritten.cpp
```diff
@@ -0,0 +1,88 @@
...
+    mStream = nullptr;
+    Result r = mBuilder.openStream(&mStream);
+    ASSERT_EQ(r, Result::OK) << "Failed to open stream." << convertToText(r);
+
+    r = mStream->start();
+    ASSERT_EQ(r, Result::OK) << "Failed to start stream." << convertToText(r);
+    sleep(PROCESS_TIME_SECOND);
+
+    // The frames written should be close to sampleRate * PROCESS_TIME_SECONDS
+    const int64_t framesWritten = mStream->getFramesWritten();
```
philburk, line 73: Also test framesRead.

tests/testStreamFramesWritten.cpp
```diff
@@ -0,0 +1,88 @@
...
+    EXPECT_NEAR(framesWritten, sampleRate * PROCESS_TIME_SECOND, sampleRate / 2);
+}
+
+INSTANTIATE_TEST_CASE_P(
+        StreamsFramesWrittenTest,
+        StreamFramesWritten,
+        ::testing::Values(
+                StreamFramesWrittenParams({Direction::Output, 8000}),
+                StreamFramesWrittenParams({Direction::Output, 16000}),
```
philburk, line 82: Test a high rate like 96000 instead of 16000.

philburk (changes requested):

tests/testStreamFramesWritten.cpp
flamme: Done

tests/testStreamFramesWritten.cpp
flamme: I have the closing in teardown function. Won't that be enough?

tests/testStreamFramesWritten.cpp
flamme: Done

tests/testStreamFramesWritten.cpp
flamme: Done

tests/testStreamFramesWritten.cpp
philburk: Yes. It's good. I didn't see it.

philburk (approved):

PR #1309, opened 2021-06-15
Update Spinners and CheckBoxes to clean up automatically

robertwu1:
setChildrenEnabled is called whenever the state changes between open, started, paused, stopped, and closed. When the stream is closed, all the spinners and checkboxes should work as expected. Otherwise, disable all of them.

This change fixes the user complaints about MMAP not working as expected when the checkbox is toggled by removing the ability to toggle the checkbox


robertwu1:
@philburk I think you need to add me as a contributor for the project before I can push.

After testing on my android simulator, I added one more spinner that I missed. There are various other controls that are still not greyed out but I don't think this is in the scope of this PR. This PR only fixes the shared spinners and textboxes settings and not the test-specific ones.

philburk (approved):

philburk:
I will squash and merge this after the CI passes.

robertwu1:
Thanks for adding me as a contributor and reviewing the PR Phil. The CI passed so I merged the changes

Issue #1306, opened 2021-06-11
getDeviceId consistently returns 0 when FilterAudioStream is in use


chrismanchester:
When attempting to report audio routes in use from Oboe, we noticed that our getDeviceId calls consistently return 0, even just after re-starting a stream in response to plugging in a new output device.

I noticed that adding `mDeviceId = mChildStream->getDeviceId();` to the constructor of FilterAudioStream fixes this issue.

philburk:
I was able to reproduce this with the following steps in OboeTester:

Launch OboeTester
Tap TEST OUTPUT
Select SRate 16000
Tap OPEN then START
Notice "devId = 2" in the middle of the screen. May be a differ N>0.
Tap CLOSE
Select SRC:Medium. (medium quality sample rate conversion)
Tap OPEN then START

EXPECT:
See "devId=2"

ACTUAL
See "devId=0"

The FilterAudioStream is a proxy class for AudioStream. it is used when we do sample rate conversion in Oboe.
It is supposed to pass commands and properties from the underlying stream. It looks like we missed the device ID property.

PR #1310, opened 2021-06-16
Added mDeviceId to FilterAudioStream

robertwu1:
Fixed a missing parameter in FilterAudioStream. Without this change, when getDeviceId() is called, the device id is set to 0, as it uses the default value. With this change, mDeviceId correctly uses the value from the child stream.

I verified that the issue repros before the change and the fix does indeed fix the problem


philburk (approved):

robertwu1:
Fixes #1306

PR #1318, opened 2021-06-28
Start linear ramps at the initial volume set

robertwu1:
Currently this ramp starts at 0 for the first ramp for no reason. Start this at the initial volume set.

This will help with clicking issues when pressing the volume up and down keys.

Tested via: Installed and ran updated tests via UnitTestRunner app

### DIFF
--- a/src/flowgraph/RampLinear.cpp
+++ b/src/flowgraph/RampLinear.cpp
@@ -33,7 +33,7 @@ void RampLinear::setLengthInFrames(int32_t frames) {
 void RampLinear::setTarget(float target) {
     mTarget.store(target);
     // If the ramp has not been used then start immediately at this level.
-    if (mLastCallCount < 0) {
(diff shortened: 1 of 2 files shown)


src/flowgraph/RampLinear.cpp
```diff
@@ -32,6 +32,10 @@ void RampLinear::setLengthInFrames(int32_t frames) {
 
 void RampLinear::setTarget(float target) {
     mTarget.store(target);
+    // If the ramp has not been used then start immediately at this level.
+    if (mLastCallCount < 0) {
```
philburk, line 36: Could check == kInitialCallCount

philburk (changes requested):

src/flowgraph/RampLinear.cpp
robertwu1: Good idea! Addressed in the second commit!

philburk (approved):

PR #1319, opened 2021-06-28
Fix CalculateLatencyInMillisReturnsClosedIfSupported test

### LINKED ISSUE #1297: Failing test StreamClosedReturnValues.CalculateLatencyInMillisReturnsClosedIfSupported
When executing tests on Pixel 4A the following test fails: 

```
StreamClosedReturnValues.CalculateLatencyInMillisReturnsClosedIfSupported
/Users/donturner/Code/workspace-android/oboe/tests/testStreamClosedMethods.cpp:320: Failure
Expected equality of these values:
  r.error()
    Which is: 4-byte object <81-FC FF-FF>
  Result::ErrorClosed
    Which is: 4-byte object <9B-FC FF-FF>
```

Device info: 

```
ro.product.brand = google
ro.product.manufacturer = Google
ro.product.model = Pixel 4a
ro.product.device = sunfish
ro.product.cpu.abi = arm64-v8a
ro.build.description = sunfish-userdebug S SP1A.210513.004.A2 7366608 dev-keys
ro.hardware = sunfish
ro.hardware.chipname = 
ro.arch = 
| grep aaudio = [aaudio.hw_burst_min_usec]: [2000]
[aaudio.mmap_exclusive_policy]: [2]
[aaudio.mmap_policy]: [2]
```

robertwu1:
AudioStreamAAudio::calculateLatencyMillis calls getTimestamp which returns Result::ErrorInvalidState when the stream is not open. CalculateLatencyInMillisReturnsClosedIfSupported explicitly checks for Result::ErrorClosed.

This was broken in #1200 

There are 3 potential "fixes"
1. Consider Result::ErrorInvalidState to be the correct error code when we query latency. The downside is the changed error code but we were having race conditions before on this anyways.
```cpp
TEST_F(StreamClosedReturnValues, CalculateLatencyInMillisReturnsClosedIfSupported){

    ASSERT_TRUE(openAndCloseStream());

    if (mStream->getAudioApi() == AudioApi::AAudio){
        auto r = mStream->calculateLatencyMillis();
        ASSERT_EQ(r.error(), Result::ErrorClosed); // Change to Result::ErrorInvalidState
    }
}
```
2. Make calculateLatencyMillis return ErrorClosed when there is no audio stream. This is the one implemented in commit 1
3. Make getTimestamp return Result::ErrorClosed instead.

```cpp
Result AudioStreamAAudio::getTimestamp(clockid_t clockId,
                                   int64_t *framePosition,
                                   int64_t *timeNanoseconds) {
    if (getState() != StreamState::Started) {
        return Result::ErrorInvalidState; // Change to Result::ErrorClosed
    }
    std::shared_lock<std::shared_mutex> lock(mAAudioStreamLock);
    if (stream != nullptr) {
        return static_cast<Result>(mLibLoader->stream_getTimestamp(stream, clockId,
                                               framePosition, timeNanoseconds));
    } else {
        return Result::ErrorNull;
    }
}

```
Phil and I discussed the approaches and converged on option 1. I also explicitly labeled the ErrorClosed in the enum for more readability in the logs

Fixes #1297

Tested with UnitTestRunner

### DIFF
--- a/include/oboe/Definitions.h
+++ b/include/oboe/Definitions.h
@@ -184,7 +184,7 @@ namespace oboe {
         Reserved8,
         Reserved9,
         Reserved10,
-        ErrorClosed,
+        ErrorClosed = -869,
     };
 
     /**
--- a/samples/build.gradle
+++ b/samples/build.gradle
@@ -20,7 +20,7 @@
 
 
 buildscript {
-    ext.kotlin_version = '1.3.50'
+    ext.kotlin_version = '1.5.20'
 
     repositories {
         google()
--- a/samples/build.gradle
+++ b/samples/build.gradle
@@ -20,7 +20,7 @@
 
 
 buildscript {
-    ext.kotlin_version = '1.5.20'
+    ext.kotlin_version = '1.3.50'
 
     repositories {
         google()
--- a/src/aaudio/AudioStreamAAudio.cpp
+++ b/src/aaudio/AudioStreamAAudio.cpp
@@ -645,6 +645,15 @@ Result AudioStreamAAudio::getTimestamp(clockid_t clockId,
 }
 
 ResultWithValue<double> AudioStreamAAudio::calculateLatencyMillis() {
+    // Check that the stream is still open before continuing
+    {
+        std::shared_lock<std::shared_mutex> lock(mAAudioStreamLock);
+        AAudioStream *stream = mAAudioStream.load();
+        if (stream == nullptr) {
+            return ResultWithValue<double>(Result::ErrorClosed);
+        }
+    }
+
     // Get the time that a known audio frame was presented.
     int64_t hardwareFrameIndex;
     int64_t hardwareFrameHardwareTime;
--- a/src/aaudio/AudioStreamAAudio.cpp
+++ b/src/aaudio/AudioStreamAAudio.cpp
@@ -645,15 +645,6 @@ Result AudioStreamAAudio::getTimestamp(clockid_t clockId,
 }
 
 ResultWithValue<double> AudioStreamAAudio::calculateLatencyMillis() {
-    // Check that the stream is still open before continuing
-    {
-        std::shared_lock<std::shared_mutex> lock(mAAudioStreamLock);
-        AAudioStream *stream = mAAudioStream.load();
-        if (stream == nullptr) {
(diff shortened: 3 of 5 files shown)


philburk (changes requested):
Let's keep the behavior in calculateLatencyMillis() because it can still return ErrorInvalidState due to a race condition if closed after this null test.
Better to test for ErrorInvalidState or just != Result::OK.

robertwu1:
> Let's keep the behavior in calculateLatencyMillis() because it can still return ErrorInvalidState due to a race condition if closed after this null test.
> Better to test for ErrorInvalidState or just != Result::OK.

Done in iterations 2/3!

dturner:
The problem with returning `ErrorInvalidState` instead of `ErrorClosed` is that it breaks convention. If you call `setBufferSizeInFrames`, `waitForStateChange`, `requestStart` or many other methods when the stream is closed you'll get `ErrorClosed`. 

I suggest that we return `ErrorClosed` when the stream is closed, and `ErrorInvalidState` when the stream is in any other state except STARTED.

robertwu1:
> The problem with returning `ErrorInvalidState` instead of `ErrorClosed` is that it breaks convention. If you call `setBufferSizeInFrames`, `waitForStateChange`, `requestStart` or many other methods when the stream is closed you'll get `ErrorClosed`.
> 
> I suggest that we return `ErrorClosed` when the stream is closed, and `ErrorInvalidState` when the stream is in any other state except STARTED.

We agree that this is the goal but option 2 (https://github.com/google/oboe/pull/1319/commits/4a82d9a3f43156c166ec523e54aead78f0cfb2b4) introduces a race condition. This first checks that we can get an mAAudioStream handle, then whether the stream is open, then checking the mAAudioStream. Phil and I have discussed that "fixing" the unittests is the main priority and we can try to unify the return error codes at a different time

philburk:
> I suggest that we return ErrorClosed when the stream is closed, and ErrorInvalidState when the
> stream is in any other state except STARTED.

I agree with that for a later PR. Let's fix the test as a high priority and then review and cleanup the error return codes next. That may require or adding or changing some tests as well.

philburk (approved):

philburk:
@robertwu1 - if you merge this, make sure to squash the 3 commits.

PR #1322, opened 2021-07-01
Fix switching between SCO and A2DP

robertwu1:
In MainActivity.java, there is code to toggle SCO on and off.

```
    public void onStartStopBluetoothSco(View view) {
        CheckBox checkBox = (CheckBox) view;
        AudioManager myAudioMgr = (AudioManager) getSystemService(Context.AUDIO_SERVICE);
        if (checkBox.isChecked()) {
            myAudioMgr.startBluetoothSco();
        } else {
            myAudioMgr.stopBluetoothSco();
        }
    }
```

This code used in conjunction with the use of startBluetoothSco() and stopBluetoothSco() causes several weird cases with SCO and A2DP. The main one is that when SCO was toggled on, TestAudioActivity would not know and SCO would never turn off.

Oboe's overarching philosophy is to let failures happen, so we removed the automatic sco adjustment in this Pull Request

Tested with Boltune and Sony devices on Pixel 4a 5G device

### DIFF
--- a/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
@@ -75,7 +75,6 @@ abstract class TestAudioActivity extends Activity {
     private MyStreamSniffer mStreamSniffer;
     private CheckBox mCallbackReturnStopBox;
     private int mSampleRate;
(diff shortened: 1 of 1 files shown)


robertwu1:
Fixes #1323

philburk (approved):
Thanks. This prevents the confusing state that can occur by having SCO turned on/off in two places in OboeTester.

PR #1331, opened 2021-07-08
Fix unused-parameter warnings

### LINKED ISSUE #312: Enabling -Wunused-parameter results in lots of "error: unused parameter" messages
I'm attempting to tidy up the Oboe core codebase and (taking a lead from the JUCE developers) am trying to get the compiler to do a bit more work for us. 

Specifically I'm attempting to use the `-Wunused-parameter` flag to provide errors when a parameter is declared but not used.

This happens a lot in the Oboe codebase because we declare empty method bodies in abstract classes, for example here: https://github.com/google/oboe/blob/master/include/oboe/AudioStreamCallback.h#L100

Just wondering, what is best practice here? Should we remove the empty function bodies, use an UNUSED macro, not worry about it? 

Relevant link: https://sites.google.com/site/opensourceconstriubtions/ettl-martin-1/articles/suppress-unused-parameter-compiler-warning

robertwu1:
Fixes #312

The compiler flag -Wunused-parameter outputs some warnings on a couple of files. This is fixed in this PR.

Tested via:
Temporarily added -Wunused-parameter to CMakeLists.txt in top level directory and compiled OboeTester.

### DIFF
--- a/src/aaudio/AudioStreamAAudio.cpp
+++ b/src/aaudio/AudioStreamAAudio.cpp
@@ -344,7 +344,7 @@ Result AudioStreamAAudio::close() {
     }
 }
 
-DataCallbackResult AudioStreamAAudio::callOnAudioReady(AAudioStream *stream,
+DataCallbackResult AudioStreamAAudio::callOnAudioReady(AAudioStream * /*stream*/,
                                                                  void *audioData,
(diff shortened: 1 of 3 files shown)


philburk (approved):
Thanks

Issue #1298, opened 2021-06-07
StabilizedCallback should wait for stream callbacks to settle before taking timings


dturner:
When you first start an audio stream there can be significant jitter on the callback period. For example, consider a stream which, when running optimally, has a 4ms callback period: 

Callback # | Optimal time | Actual time | Early/late 
-|--------------|-------------|----------
1 | 0ms | 0ms | timing model starts
2 | 4ms | 6ms | 2ms late
3 | 8ms | 9ms | 1ms late
4 | 12ms | 13ms | 1ms late
5 | 16ms | 17ms | 1ms late 

The problem with this is that as the stream jitter decreases (in callbacks 3,4,5), the stabilized callback doesn't adjust its timing model, and simply assumes that each callback is running late. This negates its effectiveness since it won't generate the correct amount of load. 

A solution to this would be to wait until the stream has stabilized before starting the timing model, say waiting for 100 callbacks.

philburk:
The decision on how much artificial load to add should be based solely on the execution time of the application. It should not be based on the start time of the callback because that is very erratic, particularly when doing sample rate conversion or when setFramesPerDataCallback().

dturner:
> The decision on how much artificial load to add should be based solely on the execution time of the application. It should not be based on the start time of the callback because that is very erratic

This doesn't work in practice precisely because the start time of the callback is erratic. 

Say you have an application which has a constant execution time, and based on that time we generate a constant amount of artificial load inside each callback with the total execution time being 80% of the callback time (calculated by numFrames / sampleRate). If the callback starts late, the artificial load may push the total execution time over the callback deadline.  

By calculating how late the callback starts based on a timing model we can reduce our artificial load time to avoid missing the deadline. 

> particularly when doing sample rate conversion or when setFramesPerDataCallback().

I'd like to understand how these scenarios affect the callback timings. My guess is it shouldn't matter as long as the following statements hold true: 

- The audio data generated inside the callback is actually being played, not discarded. If it was discarded this breaks the  timing model which assumes all data generated is actually played (`mFrameCount` is used to keep track of that, [code](https://github.com/google/oboe/blob/main/src/common/StabilizedCallback.cpp#L53)). 
- A callback cannot occur until the previous callback has finished.

robertwu1:
StabilizedCallback is no longer used in samples. Marking as closed

PR #1349, opened 2021-07-15
Improve stabilized callback timing model

dturner:
This PR updates the timing model used in `StabilizedCallback`. Essentially the epoch value is not stored until 100 callbacks have occurred. This gives the stream time to settle. 

I've tested that this code works with the MegaDrone app, however, I haven't extensively tested the impact of using original `StabilizedCallback` vs this improved version vs not using it at all. 

I wonder if we could incorporate a "variable CPU load" test into OboeTester?  

Fixes https://github.com/google/oboe/issues/1298

### DIFF
--- a/include/oboe/StabilizedCallback.h
+++ b/include/oboe/StabilizedCallback.h
@@ -22,30 +22,23 @@
 
 namespace oboe {
 
-class StabilizedCallback : public AudioStreamCallback {
+class StabilizedCallback : public AudioStreamDataCallback {
 
 public:
-    explicit StabilizedCallback(AudioStreamCallback *callback);
+    explicit StabilizedCallback(AudioStreamDataCallback *callback);
 
     DataCallbackResult
     onAudioReady(AudioStream *oboeStream, void *audioData, int32_t numFrames) override;
 
-    void onErrorBeforeClose(AudioStream *oboeStream, Result error) override {
-        return mCallback->onErrorBeforeClose(oboeStream, error);
-    }
-
-    void onErrorAfterClose(AudioStream *oboeStream, Result error) override {
-
-        // Reset all fields now that the stream has been closed
+    void resetTimingModel(){
         mFrameCount = 0;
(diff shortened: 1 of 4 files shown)


robertwu1 (approved):
LGTM! We can try something else with averaging/filtering if this doesn't seem to do the trick

philburk:
> I wonder if we could incorporate a "variable CPU load" test into OboeTester?

OboeTester TEST OUTPUT has a "workload" fader.

include/oboe/StabilizedCallback.h
```diff
@@ -22,31 +22,25 @@
 
 namespace oboe {
 
-class StabilizedCallback : public AudioStreamCallback {
+class StabilizedCallback : public AudioStreamDataCallback {
```
philburk, line 25: This is an API change that will cause source level  breaks.
Maybe add a StabilizedDataCallback and deprecate the older compatible definition that uses AudioStreamCallback..

philburk (changes requested):

include/oboe/StabilizedCallback.h
dturner: Good spot, thanks.

dturner:
I will investigate adding `StabilizedCallback` as an option to OboeTester TEST OUTPUT.

philburk:
> I will investigate adding StabilizedCallback as an option to OboeTester TEST OUTPUT.

The UI is getting crowded. Lets discuss that for a later PR.

Udik2021 (approved):

philburk:
We need to rethink this strategy based on the latest research.
Recent analysis indicates that stabilized callback can cause more glitches if the CPU scheduler is using PELT.
[b/202441352](http://b/202441352) | Core migration causes loss of CPU scheduler statistics and drop in CPU frequency, audio then glitches

robertwu1:
Closing stale PR

PR #1351, opened 2021-07-17
oboe: improve DataCallbackResult::Stop handling

### LINKED ISSUE #1230: Review handling of DataCallbackResult::Stop
Consider affect on thread joining before S.
Maybe add a blocker to prevent app callbacks.
Add some unit tests.

philburk:
Use another thread to stop the stream for R and earlier.

Fixes #1230.
Test: use OboeTester "callback return STOP" checkbox

### DIFF
--- a/src/common/AudioStream.cpp
+++ b/src/common/AudioStream.cpp
@@ -104,7 +104,6 @@ Result AudioStream::waitForStateTransition(StreamState startingState,
 
 Result AudioStream::start(int64_t timeoutNanoseconds)
 {
-    mStopThreadAllowed = true;
     Result result = requestStart();
     if (result != Result::OK) return result;
     if (timeoutNanoseconds <= 0) return result;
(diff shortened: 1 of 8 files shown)


robertwu1 (approved):

dturner (approved):
Looks good. Is it worth adding unit test(s) for this?

src/common/AudioStream.cpp
```diff
@@ -104,6 +104,7 @@ Result AudioStream::waitForStateTransition(StreamState startingState,
 
 Result AudioStream::start(int64_t timeoutNanoseconds)
 {
+    mStopThreadAllowed = true;
```
philburk, line 107: Should this be in requestStart()?!

philburk:
Yes, we need unit tests. I think a see a bug that would have been found with a unit test.

src/common/AudioStream.cpp
philburk: Yes, that was a bug.

PR #1354, opened 2021-07-20
Soundboard app

robertwu1:
New Sample that allows users to continuously play sounds. Please look at the README.md for more info


dturner:
This looks really cool. Let's discuss how this could fit with our current samples in our next meeting.

samples/shared/SynthSound.h
```diff
@@ -0,0 +1,139 @@
...
+            mFrequencies[i] = frequency * (i + 1);
+        }
+        updatePhaseIncrement();
+    };
+
+    // Amplitudes from https://epubs.siam.org/doi/pdf/10.1137/S00361445003822
+    inline void setAmplitude(float amplitude) {
+        mAmplitudes[0] = amplitude * .2f;
+        mAmplitudes[1] = amplitude;
+        mAmplitudes[2] = amplitude * .1f;
```
robertwu1, line 73: @philburk I followed your advice and added multiple overtones. I also added attack and delay and sustain. Can you let me know if this logic makes sense with you?

robertwu1: Seems like adding a large attack and the following delay makes the sound worse. Reversed the 6th commit "https://github.com/google/oboe/pull/1354/commits/a6224c9b8374a1f6b4018fd0c35986796faf2e30"

robertwu1: After doing more research, the strong attack followed by the delay makes it like a marimba instead of a piano so reverting makes sense if we wanted a synth sound instead of sounding like a marimba

philburk (changes requested):
Fantastic! You even gave it a cool icon. This is a great app.

samples/SoundBoard/src/main/java/com/google/oboe/samples/soundboard/MainActivity.java
```diff
@@ -0,0 +1,133 @@
...
+            String sampleRateStr = myAudioMgr.getProperty(AudioManager.PROPERTY_OUTPUT_SAMPLE_RATE);
+            int defaultSampleRate = Integer.parseInt(sampleRateStr);
+            String framesPerBurstStr = myAudioMgr.getProperty(AudioManager.PROPERTY_OUTPUT_FRAMES_PER_BUFFER);
+            int defaultFramesPerBurst = Integer.parseInt(framesPerBurstStr);
+
+            native_setDefaultStreamValues(defaultSampleRate, defaultFramesPerBurst);
+        }
+    }
+
+    void createMusicTiles(Context context) {
```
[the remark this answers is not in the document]
philburk: Never mind, you already did.

philburk:
I notice that DrumThumper has an average 32 msec tap-to-tone latency and SoundBoard has 45.
That could be due to the bufferSize, touch input technique, or attack time of the sound.
Worth investigating.

philburk:
The app sounds a bit quite compared to other apps. Maybe boost the volume a bit, even if maxVoices*amplitude > 1.0.

samples/SoundBoard/local.properties
```diff
@@ -0,0 +1,8 @@
+## This file must *NOT* be checked into Version Control Systems,
```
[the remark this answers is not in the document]
robertwu1: Thanks!

samples/SoundBoard/src/main/java/com/google/oboe/samples/soundboard/MusicTileView.java
```diff
@@ -0,0 +1,143 @@
...
+    private ArrayList<Rect> mRectangles;
+    private boolean[] mIsPressedPerRectangle;
+    private Paint mPaint;
+    private SparseArray<PointF> mLocationsOfFingers;
+
+    private long mEngineHandle = 0;
+    private native void tap(long engineHandle, boolean isDown, int audioSource);
+    private boolean isInitialOnDrawCall = true;
+
+    public MusicTileView(Context context, ArrayList<Rect> rectangles, long engineHandle) {
```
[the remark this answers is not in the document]
robertwu1: I don't see a grid being much better than what we are doing here: https://stackoverflow.com/questions/14917982/how-to-draw-a-grid-of-rectangle-shaped-objects. It seems like we need to keep track of them anyways and a 2d array might be more of a pain than it is worth

robertwu1:
> I notice that DrumThumper has an average 32 msec tap-to-tone latency and SoundBoard has 45.
> That could be due to the bufferSize, touch input technique, or attack time of the sound.
> Worth investigating.

Thanks! Instead waiting for the onDraw routine, we can call the sound directly!

samples/shared/SynthSound.h
```diff
@@ -0,0 +1,118 @@
...
+    };
+    // From IRenderableAudio
+    void renderAudio(float *audioData, int32_t numFrames) override {
+        if (mIsWaveOn){
+            for (int i = 0; i < numFrames; ++i) {
+                audioData[i] = 0;
+                for (int j = 0; j < kNumSineWaves; ++j) {
+                    audioData[i] += sinf(mPhases[j]) * mAmplitudes[j] * mAttackCount / kAttackPeriods;
+
+                    mPhases[j].store(mPhases[j] + mPhaseIncrements[j]);
```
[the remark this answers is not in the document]
robertwu1: Thanks!

samples/shared/SynthSound.h
```diff
@@ -0,0 +1,118 @@
...
+                        if (mAttackCount < kAttackPeriods)
+                        {
+                            mAttackCount++;
+                        }
+                    }
+                }
+            }
+        } else {
+            memset(audioData, 0, sizeof(float) * numFrames);
+            // Remove some noise by letting the current sine wave complete.
```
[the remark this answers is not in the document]
robertwu1: This is great! Thanks Phil!

samples/SoundBoard/src/main/java/com/google/oboe/samples/soundboard/MusicTileView.java
```diff
@@ -0,0 +1,143 @@
...
+        mRectangles = rectangles;
+        mIsPressedPerRectangle = new boolean[rectangles.size()];
+        mPaint = new Paint();
+        mLocationsOfFingers = new SparseArray<PointF>();
+        mEngineHandle = engineHandle;
+    }
+
+    private int getIndexFromLocation(PointF pointF) {
+        for (int i = 0; i < mRectangles.size(); i++) {
+            if (pointF.x > mRectangles.get(i).left &&
```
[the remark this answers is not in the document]
robertwu1: Rectangle.contains() doesn't handle PointF. It doesn't make sense to cast this so resolving

samples/SoundBoard/src/main/cpp/SoundBoardEngine.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+SoundBoardEngine::~SoundBoardEngine() {
+    if (mStream) {
+        LOGE("SoundBoardEngine destructor was called without calling stop()."
+             "Please call stop() to ensure stream resources are not leaked.");
+        stop();
+    }
+}
+
+void SoundBoardEngine::tapSource(bool isDown, int32_t audioSource) {
```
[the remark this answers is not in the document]
robertwu1: Done!

robertwu1:
> The app sounds a bit quite compared to other apps. Maybe boost the volume a bit, even if maxVoices*amplitude > 1.0.

Thanks! I changed the amplitude to .2 from .7. I noticed when I changed the amplitude to 1.0, the sound quality is really bad, even when I lowered the volume. I presume this is from some part of the stack clipping the audio.

robertwu1:
@philburk Thanks for the thorough review! I addressed all the comments. It seems like compared to a tap to tone latency of 34ms with this app and 35ms with DrumThumper, a random top piano app I downloaded had 67ms latency. After considering that the tap latency of the display, it seems like this app is magnitudes faster than what is currently out there. Hopefully by releasing this sample app, we can get current app creators to convert to Oboe.

@dturner I can depreciate MegaDrone in an upcoming PR if we decide to submit this. I realized that everything you want to test/show with MegaDrone is already done with soundboard.

Thanks for all your suggestions and let me know if you have any questions!

samples/SoundBoard/src/main/cpp/SoundBoardEngine.cpp
```diff
@@ -42,8 +42,12 @@ SoundBoardEngine::~SoundBoardEngine() {
     }
 }
 
-void SoundBoardEngine::tapSource(bool isDown, int32_t audioSource) {
-    mSynth->tapSource(isDown, audioSource);
+void SoundBoardEngine::setNoteOff(int32_t noteIndex) {
```
philburk, line 45: noteOn() is not normally a "set" method because we are triggering an event, not setting/getting a property.

samples/SoundBoard/src/main/cpp/Synth.h
```diff
@@ -54,9 +54,13 @@ class Synth : public IRenderableAudio, public ITappable {
         }
     }
 
-    void tapSource(bool isOn, int32_t source) {
-        mOscs[source].setWaveOn(isOn);
-    };
+    void setNoteOff(int32_t noteIndex) {
+        mOscs[noteIndex].setWaveOn(false);
```
philburk, line 58: I just realized that you are not using a voice allocator. That may be beyond the scope of this example. Or we can consider it for a later PR.

samples/SoundBoard/src/main/java/com/google/oboe/samples/soundboard/MusicTileView.java
```diff
@@ -134,7 +127,14 @@ public boolean onTouchEvent(MotionEvent event) {
 
         // Calling invalidate() will force onDraw() to be called
         if (didImageChange) {
-            isInitialOnDrawCall = false;
+            // Call c++ native code to set each index add/remove sound
+            for (int i = 0; i < mRectangles.size(); i++) {
+                if (mIsPressedPerRectangle[i]) {
+                    setNoteOn(mEngineHandle, i);
```
philburk, line 133: This could cause multiple noteOn() events if a finger is held down while another finger is moved.

samples/shared/SynthSound.h
```diff
@@ -25,22 +25,21 @@ constexpr int32_t kDefaultSampleRate = 48000;
...
 public:
     SynthSound() {
-        for (int i = 0; i < kNumSineWaves; i++) {
-            mFrequencies[i] = kDefaultFrequency * (i + 1);
-        }
     }
     ~SynthSound() = default;
 
     void setWaveOn(bool isWaveOn) {
         if (!mIsWaveOn && isWaveOn) {
```
philburk, line 41: This edge detection should occur at the UI level. Then you could use the same UI to trigger MIDI events, which do not use the set/get paradigm.

If a noteOn() occurs while a note is already on then it should retrigger the envelope. This may not be obvious if using a steady sustain value.

samples/shared/SynthSound.h
```diff
@@ -65,54 +62,42 @@ class SynthSound : public IRenderableAudio {
...
+                mPhase.store(mPhase + mPhaseIncrement);
+                if (mPhase > kTwoPi) {
+                    mPhase.store(mPhase - kTwoPi);
                 }
             }
         } else {
             memset(audioData, 0, sizeof(float) * numFrames);
             // Remove some noise by letting the current sine wave complete.
             // Since all the frequencies are multiples of mPhases[0], if the first wave complete,
             // so should the others. Use an extra release period to reduce even more noise.
```
philburk, line 80: This comment does not match the code.

samples/shared/SynthSound.h
```diff
@@ -65,54 +62,42 @@ class SynthSound : public IRenderableAudio {
...
-    std::array<std::atomic<double>, kNumSineWaves> mFrequencies;
+    std::atomic<float> mPhase;
+    std::atomic<double> mPhaseIncrement;
+    std::atomic<double> mFrequency { kDefaultFrequency };
     std::atomic<int32_t> mSampleRate { kDefaultSampleRate };
     void updatePhaseIncrement(){
-        for (int i = 0; i < kNumSineWaves; i++) {
-            mPhaseIncrements[i] = (kTwoPi * mFrequencies[i]) / static_cast<double>(mSampleRate);
-        }
+        mPhaseIncrement = kTwoPi * mFrequency / static_cast<double>(mSampleRate);
```
philburk, line 100: A divide is surprisingly inefficient compared to a multiply. Cache the inverse and use that in your real-time code. This is particularly important if you a modulating the frequency.

    mSamplePeriod = 1.0 / mSampleRate;

then later

    mPhaseIncrement = ..... * mSamplePeriod;

samples/SoundBoard/src/main/java/com/google/oboe/samples/soundboard/MusicTileView.java
```diff
@@ -0,0 +1,143 @@
...
+import static android.os.Process.getExclusiveCores;
+
+public class MusicTileView extends View {
+    private ArrayList<Rect> mRectangles;
+    private boolean[] mIsPressedPerRectangle;
+    private Paint mPaint;
+    private SparseArray<PointF> mLocationsOfFingers;
+
+    private long mEngineHandle = 0;
+    private native void setNoteOn(long engineHandle, int noteIndex);
```
philburk, line 24: Rather than calling a native method directly, call a TileListener interface that has onTileOn(index) and onTimeOff(index) methods. Then you could reuse this for other use cases including MIDI, or even games that do not use notes.  You can implement onTile(index) to call noteOn(index).

samples/shared/SynthSound.h
```diff
@@ -0,0 +1,103 @@
...
+        mAmplitudes[1] = amplitude;
+        mAmplitudes[2] = amplitude * .1f;
+        mAmplitudes[3] = amplitude * .02f;
+        mAmplitudes[4] = amplitude * .15f;
+    };
+    // From IRenderableAudio
+    void renderAudio(float *audioData, int32_t numFrames) override {
+        if (mIsWaveOn){
+            for (int i = 0; (i < numFrames) && (mMasterAmplitude > kMasterAmplitudeCutOff); ++i) {
+                mMasterAmplitude.store(mMasterAmplitude * kSustainMultiplier);
```
philburk, line 66: Race condition here. You could use an std::atomic<bool> to trigger the "envelope" atomically.

    noteOn() {
       trigger = true; // start a note envelope
       mIsWaveOn = true;
       mAmplitudeScaler = kDecayScaler;
    }

   noteOff() {
       mAmplitudeScaler = kReleaseScaler;
    }

then from your render thread

    if (trigger.exchange(false)) {
        mMasterAmplitude = 1.0;
    } else {
        mMasterAmplitude *= kWhatever;
    }

So mMasterAmplitude will only be modified from a single thread.

samples/shared/SynthSound.h
```diff
@@ -0,0 +1,103 @@
...
+                audioData[i] = 0;
+                for (int j = 0; j < kNumSineWaves; ++j) {
+                    audioData[i] += sinf(mPhase * (j + 1)) * mAmplitudes[j] * mMasterAmplitude;
+                }
+                mPhase.store(mPhase + mPhaseIncrement);
+                if (mPhase > kTwoPi) {
+                    mPhase.store(mPhase - kTwoPi);
+                }
+            }
+        } else {
```
philburk, line 76: You can eliminate this second block. The only difference from above is the amplitude scaler.

philburk (changes requested):

samples/SoundBoard/src/main/cpp/SoundBoardEngine.cpp
robertwu1: Makes sense to me! Done!

samples/SoundBoard/src/main/java/com/google/oboe/samples/soundboard/MusicTileView.java
robertwu1: Thanks for bringing this up! I have changed the logic when the finger has moved

samples/shared/SynthSound.h
robertwu1: Thanks for giving me a suggested fix below. I have done the following instead:

    void noteOn() {
        mTrigger = true; // start a note envelope
        mAmplitudeScaler = kSustainMultiplier;
    }

    void noteOff() {
        mAmplitudeScaler = kReleaseMultiplier;
    }

samples/SoundBoard/src/main/cpp/Synth.h
robertwu1: What is a voice allocator? If you mean dynamically allocating voices instead of using an array, we can do this in another PR. This one is getting unwieldy

samples/shared/SynthSound.h
robertwu1: Nice catch!

samples/shared/SynthSound.h
robertwu1: mPhaseIncrement is never used in any division so it's fine as is. Thanks for bringing this up though! I'll watch out for optimizations like this in the future

samples/SoundBoard/src/main/java/com/google/oboe/samples/soundboard/MusicTileView.java
robertwu1: Done!

samples/shared/SynthSound.h
robertwu1: Thanks!

samples/shared/SynthSound.h
robertwu1: Good idea! Done!

robertwu1:
@philburk Thanks for submitting more comments. Let me know if you have any more feedback!

samples/shared/SynthSound.h
philburk: > mPhaseIncrement is never used in any division so it's fine as is.

But a divide is used to calculate the mPhaseIncrement. It doesn't matter so much here because you are not varying the mPhaseIncrement. But if you add vibrato or portamento then it would be worth making this optimization.

samples/SoundBoard/src/main/java/com/google/oboe/samples/soundboard/MusicTileView.java
```diff
@@ -0,0 +1,147 @@
+package com.google.oboe.samples.soundboard;
+
```
philburk, line 2: Needs AOSP header.

samples/SoundBoard/src/main/java/com/google/oboe/samples/soundboard/NoteListener.java
```diff
@@ -0,0 +1,20 @@
+package com.google.oboe.samples.soundboard;
+
```
philburk, line 2: Needs AOSP header.

samples/SoundBoard/src/main/java/com/google/oboe/samples/soundboard/TileListener.java
```diff
@@ -0,0 +1,7 @@
+package com.google.oboe.samples.soundboard;
+
```
philburk, line 2: Needs AOSP header. OR define inside MusicTileView.

philburk (changes requested):

samples/SoundBoard/src/main/java/com/google/oboe/samples/soundboard/MusicTileView.java
```diff
@@ -19,19 +20,16 @@
     private boolean[] mIsPressedPerRectangle;
     private Paint mPaint;
     private SparseArray<PointF> mLocationsOfFingers;
+    private NoteListener mNoteListener;
```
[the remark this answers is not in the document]
robertwu1: I moved the interface and the class inside MusicTileView. Let me know if this looks right

samples/SoundBoard/src/main/java/com/google/oboe/samples/soundboard/NoteListener.java
robertwu1: Thanks!

samples/SoundBoard/src/main/java/com/google/oboe/samples/soundboard/MusicTileView.java
robertwu1: Thanks!

samples/SoundBoard/src/main/java/com/google/oboe/samples/soundboard/NoteListener.java
robertwu1: Removed this file actually

samples/SoundBoard/src/main/java/com/google/oboe/samples/soundboard/TileListener.java
robertwu1: Removed file

samples/SoundBoard/src/main/java/com/google/oboe/samples/soundboard/MusicTileView.java
```diff
@@ -75,26 +73,38 @@ public boolean onTouchEvent(MotionEvent event) {
...
+                                notesChangedBy[prevIndex]--;
+                            }
+                            if (newIndex != -1) {
+                                notesChangedBy[newIndex]++;
+                            }
                         }
                     }
                 }
+
+                // Now go through the rectangles to see if they have changed
```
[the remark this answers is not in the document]
robertwu1: Yes, I added some comments to the previous for loop

samples/shared/SynthSound.h
robertwu1: I added a comment for curious readers

robertwu1:
@philburk Thanks for adding more comments. They are very helpful for me learning best practices in Java

samples/SoundBoard/src/main/java/com/google/oboe/samples/soundboard/MusicTileView.java
```diff
@@ -20,16 +36,41 @@
     private boolean[] mIsPressedPerRectangle;
     private Paint mPaint;
     private SparseArray<PointF> mLocationsOfFingers;
-    private NoteListener mNoteListener;
+    private TileListener mNoteListener;
+
+    public interface TileListener {
```
philburk, line 41: This goes inside the MusicTileView class as a public inner class.

samples/SoundBoard/src/main/java/com/google/oboe/samples/soundboard/MusicTileView.java
```diff
@@ -20,16 +36,41 @@
...
+        public void onTileOn(int index);
+        public void onTileOff(int index);
+    }
+
+    public class NoteListener implements TileListener {
```
philburk, line 46: This goes in its own file like before. Sorry that I was not clear.

samples/SoundBoard/src/main/java/com/google/oboe/samples/soundboard/MusicTileView.java
```diff
@@ -20,16 +36,41 @@
...
+        public void onTileOn(int index) {
+            noteOn(mEngineHandle, index);
+        }
+
+        public void onTileOff(int index) {
+            noteOff(mEngineHandle, index);
+        }
+    }
+
+    public MusicTileView(Context context, ArrayList<Rect> rectangles, long mEngineHandle) {
```
philburk, line 65: MusicTileView should not know anything about the synthesizer.

samples/SoundBoard/src/main/java/com/google/oboe/samples/soundboard/MusicTileView.java
```diff
@@ -73,6 +114,9 @@ public boolean onTouchEvent(MotionEvent event) {
         switch (maskedAction) {
             // Move each point from it's current point to the new point.
             case MotionEvent.ACTION_MOVE: {
+                // Create an array to check for finger changes as multiple fingers may be on the
+                // same tile. This two-pass algorithm records the difference before changing the
```
philburk, line 118: "changing the figuring"?

samples/shared/SynthSound.h
```diff
@@ -100,6 +100,9 @@ class SynthSound : public IRenderableAudio {
     std::atomic<float> mFrequency { kDefaultFrequency };
     std::atomic<int32_t> mSampleRate { kDefaultSampleRate };
     void updatePhaseIncrement(){
+        // Note how there is a division here. If this file is changed so that updatePhaseIncrement
```
philburk, line 103: Good.

philburk (changes requested):

samples/SoundBoard/src/main/java/com/google/oboe/samples/soundboard/MusicTileView.java
robertwu1: Thanks for going over this with me offline. Done!

samples/SoundBoard/src/main/java/com/google/oboe/samples/soundboard/MusicTileView.java
robertwu1: Thanks!

samples/SoundBoard/src/main/java/com/google/oboe/samples/soundboard/MusicTileView.java
robertwu1: Nice catch!

robertwu1:
@philburk Addressed your last set of comments. Let me know if I can submit this

philburk (approved):
Looks great. Thanks for adding a cool new sample app.

PR #1355, opened 2021-07-20
Update GetPackageName header

robertwu1:
In AAudio, we changed GetPackageName to better reflect its usage. Doing the same in Oboe


robertwu1:
Fixes #1356

philburk (approved):

PR #1381, opened 2021-08-12
Quirk: disable MMAP on SDM8150

### LINKED ISSUE #1121: No audio in ZTE NX629J
Android version(s): 9.0
Android device(s): ZTE NX629J
Oboe version: 1.4.3
App name used for testing: M64Plus FZ
 
**Short description**
I'm getting complaints of some devices having no audio after switching to the OBOE library.

**Steps to reproduce**
Play a game in the app

**Expected behavior**
Oboe should produce audio output

**Actual behavior**
Stream fails to start for an unknown reason.

**Device**
ZTE NX629J at the very least. Probably other "red magic" devices as well.

I don't have access to the device, but it's using a qualcomm SD 855, arm64-v8a, and it reports as supporting aaudio

```
for p in \
    ro.product.brand ro.product.manufacturer ro.product.model \
    ro.product.device ro.product.cpu.abi ro.build.description \
    ro.hardware ro.hardware.chipname ro.arch "| grep aaudio";
    do echo "$p = $(adb shell getprop $p)"; done
```

**Any additional context**

Unfortunately, I don't have access to the specific device, so the best I can do is report that the stream fails to start for specific devices.

If it helps, the application is open source. The audio handling can be found here: https://github.com/mupen64plus-ae/mupen64plus-ae/blob/master/mupen64plus-audio-android/src/AudioHandler.cpp

Also, if you can pinpoint anything that could be improved in the usage of the library, that wouldn't be a bad thing either.

Another data point, I have done some Google searching online and it seems that users complain about no audio in other games as well on those specific devices.

philburk:
MMAP not supported before Q.
But some phones tried to use it.

Fixes #1121

### DIFF
--- a/src/common/QuirksManager.cpp
+++ b/src/common/QuirksManager.cpp
@@ -63,15 +63,12 @@ bool QuirksManager::DeviceQuirks::isAAudioMMapPossible(const AudioStreamBuilder 
 }
 
 /**
- * This is mostly for Exynos quirks. Samsung Mobile uses Qualcomm chips so
+ * This is for Samsung Exynos quirks. Samsung Mobile uses Qualcomm chips so
  * the QualcommDeviceQuirks would apply.
  */
-class SamsungDeviceQuirks : public  QuirksManager::DeviceQuirks {
+class SamsungExynosDeviceQuirks : public  QuirksManager::DeviceQuirks {
 public:
-    SamsungDeviceQuirks() {
-        std::string arch = getPropertyString("ro.arch");
-        isExynos = (arch.rfind("exynos", 0) == 0); // starts with?
-
+    SamsungExynosDeviceQuirks() {
         std::string chipname = getPropertyString("ro.hardware.chipname");
         isExynos9810 = (chipname == "exynos9810");
         isExynos990 = (chipname == "exynos990");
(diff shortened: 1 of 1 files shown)


src/common/QuirksManager.cpp
```diff
@@ -135,12 +139,42 @@ class SamsungDeviceQuirks : public  QuirksManager::DeviceQuirks {
...
+
+    virtual ~QualcommDeviceQuirks() = default;
+
+    bool isMMapSafe(const AudioStreamBuilder &builder) override {
+        // See https://github.com/google/oboe/issues/1121#issuecomment-897957749
+        bool mmapBroken = false;
+        if (isSM8150 && (getSdkVersion() <= __ANDROID_API_P__)) {
+            LOGI("QuirksManager::%s() MMAP not actually supported on this chip."
+                 " Switching off MMAP.", __func__);
+            mmapBroken = true;
```
robertwu1, line 157: nit: isMmapBroken

src/common/QuirksManager.cpp
```diff
@@ -135,12 +139,42 @@ class SamsungDeviceQuirks : public  QuirksManager::DeviceQuirks {
...
+
 QuirksManager::QuirksManager() {
-    std::string manufacturer = getPropertyString("ro.product.manufacturer");
-    if (manufacturer == "samsung") {
+    std::string productManufacturer = getPropertyString("ro.product.manufacturer");
+    if (productManufacturer == "samsung") {
         mDeviceQuirks = std::make_unique<SamsungDeviceQuirks>();
     } else {
-        mDeviceQuirks = std::make_unique<DeviceQuirks>();
+        std::string socManufacturer = getPropertyString("ro.soc.manufacturer");
```
robertwu1, line 172: This code makes it seem that a device will is either a Samsung Device or a Qualcomm device. Would it be possible for devices to have multiple Quirks in the future?

We should worry about this later but I think the next time we add a new Quirk, like let's say for Pixel 4a, we should consider doing some refactor of this.

robertwu1 (approved):

src/common/QuirksManager.cpp
philburk: Good Point. We should distinguish more clearly between the Samsung Exynos quirks and the SamsungMobile quirks. They are quite different.

robertwu1 (approved):

PR #1388, opened 2021-08-27
(OboeTester) Add time when tap to tone is pressed to logs

### LINKED ISSUE #1387: (OboeTester) Add time when tap to tone is pressed to logs
Tap to tone relies on the fingernail pressing on the screen to detect when the test has started. Alternatively, we can simply use the time the tap signal is sent. This helps eliminate screen latency from Tap to Tone.

This first issue is simply adding when the waveform is sent to the logs. Updating tap to tone to use this behavior can come later if needed.

robertwu1:
Fixes #1387

### DIFF
--- a/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TapToToneActivity.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TapToToneActivity.java
@@ -38,6 +38,7 @@ import com.mobileer.miditools.MidiPortConnector;
 import com.mobileer.miditools.MidiTools;
 
 import java.io.IOException;
+import java.sql.Timestamp;
 
 import static com.mobileer.oboetester.MidiTapTester.NoteListener;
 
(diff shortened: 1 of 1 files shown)


apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TapToToneActivity.java
```diff
@@ -116,6 +117,8 @@ private void updateButtons(boolean running) {
 
     void trigger() {
         if (mTapToToneTester.isArmed()) {
+            Timestamp timestamp = new Timestamp(System.currentTimeMillis());
```
philburk, line 120: The mAudioOutTester.trigger() call is very fast so please put this new code after it.
That way the log() will not contribute to the measured latency.

philburk (changes requested):

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TapToToneActivity.java
robertwu1: Thanks!

PR #1390, opened 2021-08-30
(OboeTester) Create automatic data paths tests

### LINKED ISSUE #1294: OboeTester needs automated Data Paths test
Use an Intent similar to the RT latency test.

robertwu1:
Add automatic data paths tests with the following parameters:

> --ez use_input_presets  {boolean}  // Whether to test various input presets. Note use of "-ez"
> --ez use_input_devices  {boolean}  // Whether to test various input devices. Note use of "-ez"
> --ez use_output_devices {boolean}  // Whether to test various output devices. Note use of "-ez"
> --ei single_test_index  {testId}   // Index for testing one specific test

I've updated the markdown file in the docs directory since we have added a new test of tests 

Fixes #1294

### DIFF
--- a/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestDataPathsActivity.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestDataPathsActivity.java
@@ -16,17 +16,22 @@
 
 package com.mobileer.oboetester;
 
+import android.content.Intent;
 import android.app.Activity;
 import android.content.Context;
 import android.media.AudioDeviceInfo;
 import android.media.AudioManager;
 import android.os.Bundle;
+import android.os.Handler;
+import android.os.Looper;
 import android.support.annotation.NonNull;
+import android.util.Log;
 import android.widget.CheckBox;
 import android.widget.TextView;
 
 import com.mobileer.audio_device.AudioDeviceInfoConverter;
 
+import java.io.IOException;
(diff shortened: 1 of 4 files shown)


apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestDataPathsActivity.java
```diff
@@ -553,27 +610,43 @@ public void runTest() {
...
+                mCheckBoxInputPresets.setEnabled(false);
+                mCheckBoxInputDevices.setEnabled(false);
+                mCheckBoxOutputDevices.setEnabled(false);
+            });
+
             mTestResults.clear();
             mDurationSeconds = DURATION_SECONDS;
 
             if (mCheckBoxInputPresets.isChecked()) {
-                runOnUiThread(() -> mCheckBoxInputPresets.setEnabled(false));
```
philburk, line 560: This strange looking code was designed to allow a user to opt out of some sub-tests while the test was running.

philburk (changes requested):

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestDataPathsActivity.java
robertwu1: Thanks for the heads up! I reverted it

PR #1394, opened 2021-08-31
(OboeTester) Add time between callback statistics

### LINKED ISSUE #1393: (OboeTester) Add some way to figure out time between callbacks
For playing audio from speakers, the audio should not be be bursty. However, for playing audio through Bluetooth, the audio should be bursty.

Figure out how to add this info

robertwu1:
Fixes #1393

Verified that speakers are stable 2.0/2.0/2.1 for min/avg/max in ms and bluetooth is bursty 0.0/1.8/8.6

### DIFF
--- a/apps/OboeTester/app/src/main/cpp/OboeStreamCallbackProxy.cpp
+++ b/apps/OboeTester/app/src/main/cpp/OboeStreamCallbackProxy.cpp
@@ -77,5 +77,12 @@ oboe::DataCallbackResult OboeStreamCallbackProxy::onAudioReady(
     double currentCpuLoad = calculationTime * inverseRealTime; // avoid a divide
     mCpuLoad = (mCpuLoad * 0.95) + (currentCpuLoad * 0.05); // simple low pass filter
 
+    int64_t currentTimeNs = getNanoseconds();
+    if (previousCallbackTimeNs != 0) {
+        mCallbackTimeMs = (currentTimeNs - previousCallbackTimeNs)
+                * .000001;
+    }
+    previousCallbackTimeNs = currentTimeNs;
+
     return callbackResult;
 }
--- a/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/AudioStreamBase.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/AudioStreamBase.java
@@ -26,6 +26,7 @@ public abstract class AudioStreamBase {
     private StreamConfiguration mRequestedStreamConfiguration;
     private StreamConfiguration mActualStreamConfiguration;
     AudioStreamBase.DoubleStatistics mLatencyStatistics;
+    AudioStreamBase.DoubleStatistics mTimeBetweenCallbacksStatistics;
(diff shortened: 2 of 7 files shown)


robertwu1:
image

See time between callbacks

philburk (changes requested):
You are sampling a rapidly changing duration and the probability of reading a particular duration is highly skewed based on the time that the duration is in the variable. Consider the case where you always get two callbacks together followed by a pause.  You are more likely to sample during the pause so the average will be weighted towards the lower duration. It would be better to gather the statistics in the native code and then read the slowly changing stats up into Java.

This is different than the latency measurement, which has smoother behavior.

apps/OboeTester/app/src/main/cpp/OboeStreamCallbackProxy.cpp
```diff
@@ -77,5 +77,12 @@ oboe::DataCallbackResult OboeStreamCallbackProxy::onAudioReady(
     double currentCpuLoad = calculationTime * inverseRealTime; // avoid a divide
     mCpuLoad = (mCpuLoad * 0.95) + (currentCpuLoad * 0.05); // simple low pass filter
 
+    int64_t currentTimeNs = getNanoseconds();
+    if (previousCallbackTimeNs != 0) {
+        mCallbackTimeMs = (currentTimeNs - previousCallbackTimeNs)
+                * .000001;
```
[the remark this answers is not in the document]
robertwu1: Just added a constant!

robertwu1:
> You are sampling a rapidly changing duration and the probability of reading a particular duration is highly skewed based on the time that the duration is in the variable. Consider the case where you always get two callbacks together followed by a pause. You are more likely to sample during the pause so the average will be weighted towards the lower duration. It would be better to gather the statistics in the native code and then read the slowly changing stats up into Java.
> 
> This is different than the latency measurement, which has smoother behavior.

Thanks for bringing this up! It's a good idea. I have done as you have suggested in update 2

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/AudioStreamBase.java
```diff
@@ -92,6 +94,8 @@ String dump(int framesPerBurst) {
             }
             StringBuffer buffer = new StringBuffer();
 
+            buffer.append("time between callbacks = " + callbackTimeStr + "\n");
+
```
philburk, line 98: The line from 110 with the xRun# is no longer being displayed on screen. You neeed to change "statusView" at bottom of "stream_config.xml" to have 5 lines instead of 4.

philburk (changes requested):

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/AudioStreamBase.java
robertwu1: Thanks for the heads up! Done in iteration 3

philburk (approved):

Issue #1399, opened 2021-09-13
OboeTester: add test for device switching time


philburk:
For example, plug or unplug a headset and measure time from detection to reopening a new stream.

Based on internal b/137392532

philburk:
Can add a test on the "EXTRAS..." page.

philburk:
Measure several events and show times between them:

1. Java onAudioDevicesAdded/Removed events
2. error callback from AAudio - triggers reopening of the stream in C++
3. time after close() of original stream
4. time after open() of next stream
5. time of first callback

Show total elapsed time.

PR #1403, opened 2021-09-24
(OboeTester) Add Test Plug Latency Activity

robertwu1:
Adds an activity that measures latency when devices are added/removed. See the screenshot for what this looks like
Screenshot_20210924-142412

### DIFF
--- a/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestPlugLatencyActivity.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestPlugLatencyActivity.java
@@ -1,0 +1,214 @@
+/*
+ * Copyright 2019 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mobileer.oboetester;
+
+import android.annotation.TargetApi;
+import android.content.BroadcastReceiver;
+import android.content.Context;
+import android.content.Intent;
+import android.content.IntentFilter;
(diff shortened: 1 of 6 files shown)


apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestPlugLatencyActivity.java
```diff
@@ -0,0 +1,214 @@
...
+                mInstructionsTextView.setText(text);
+            }
+        });
+    }
+
+    public void startAudioTest() throws IOException {
+        startAudio();
+    }
+
+    private long getLatencyMs() {
```
philburk, line 150: This does a lot more than a normal "get" method. Maybe "measureLatencyMs".

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestPlugLatencyActivity.java
```diff
@@ -0,0 +1,214 @@
...
+    private AudioManager mAudioManager;
+
+    private volatile int mPlugCount = 0;
+
+    class MyAudioDeviceCallback extends AudioDeviceCallback {
+        private HashMap<Integer, AudioDeviceInfo> mDevices
+                = new HashMap<Integer, AudioDeviceInfo>();
+
+        @Override
+        public void onAudioDevicesAdded(AudioDeviceInfo[] addedDevices) {
```
philburk, line 57: Ignore the initial flurry of devices added, which includes speakers. Just log the add/remove messages.

philburk (changes requested):
Very good start. Intention was to measure time it takes for a restart of a stream from C++.
See https://github.com/google/oboe/issues/1399#issuecomment-926988265

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestPlugLatencyActivity.java
robertwu1: Thanks!

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestPlugLatencyActivity.java
robertwu1: Good idea!

robertwu1:
Below is the following after commit 3. Let me know how this looks. Thanks!

Screenshot_20210927-120426

robertwu1:
Screenshot_20211004-172151

Phil suggested to use less space per line and to add line breaks. Done in commit 4

PR #1422, opened 2021-10-19
(OboeTester) Add sine amplitude to glitch test

### LINKED ISSUE #870: OboeTester glitch test show magnitude
Show the sine magnitude measured by the locking algorithm.
It will be more stable than the peak.amplitude.

robertwu1:
Fixes #870 

Screenshot_20211019-115338

This page is getting large and unwieldy. Consider cleaning up unnecessary data in another PR.

### DIFF
--- a/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/GlitchActivity.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/GlitchActivity.java
@@ -48,6 +48,7 @@ public class GlitchActivity extends AnalyzerActivity {
     native int getGlitchCount();
     native double getSignalToNoiseDB();
     native double getPeakAmplitude();
+    native double getSineAmplitude();
 
     private GlitchSniffer mGlitchSniffer;
     private NativeSniffer mNativeSniffer = createNativeSniffer();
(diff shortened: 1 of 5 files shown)


apps/OboeTester/app/src/main/java/com/mobileer/oboetester/GlitchActivity.java
```diff
@@ -175,6 +178,7 @@ private String getCurrentStatusReport() {
             message.append(String.format("glitch.frames = %d\n", mLastGlitchFrames));
             message.append(String.format("reset.count = %d\n", mLastResetCount - mStartResetCount));
             message.append(String.format("peak.amplitude = %8.6f\n", mPeakAmplitude));
+            message.append(String.format("sine.amplitude = %8.6f\n", mSineAmplitude));
```
philburk, line 181: This overflows the line count for the TextView so the "glitch.count" line is now missing.

philburk (changes requested):
Thanks for adding this.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/GlitchActivity.java
robertwu1: Nice catch!

philburk (approved):

PR #1432, opened 2021-11-04
(drumthumper) remove redundant semicolon in kotlin

lerendan:
(drumthumper) remove redundant semicolon in kotlin


robertwu1:
Thanks for the changes! They look good to me

PR #1441, opened 2021-11-09
Add a sine wave in round trip latency to account for start-up delay

### LINKED ISSUE #1440: (OboeTester) Round Trip Latency needs to account for start-up time
Android version(s): ALL
Android device(s): ALL
Oboe version: ALL

OboeTester does not send real data when creating the stream in Round Trip Latency. This means start-up latency is not accounted for in some cases like Bluetooth

robertwu1:
Fixes #1440

This fixes the issue of some devices not completely setting up until real data is sent. Since the A4 sine wave is played for 2 seconds, this should be plenty of time for the input and output on a device to be ready.


robertwu1:
Abandoning because Phil has a better solution in mind

PR #1471, opened 2021-12-30
(HelloOboe) Add Bluetooth SCO sample

### LINKED ISSUE #909: call startBluetoothSco() when needed in Oboe samples
See this PR as an example.
https://github.com/google/oboe/pull/896

robertwu1:
Fixes #909

This PR adds an ability to toggle Bluetooth SCO when it is actually set. We should also update the Wiki to document Bluetooth after this PR is submitted.

Tested on Pixel 4a and Pixel Series A.
Verified that SCO does not play correctly before this change.
Toggled between SCO and A2DP and verified that the sound changes with this change.

### DIFF
--- a/samples/hello-oboe/src/main/java/com/google/oboe/samples/hellooboe/MainActivity.java
+++ b/samples/hello-oboe/src/main/java/com/google/oboe/samples/hellooboe/MainActivity.java
@@ -17,9 +17,13 @@
 package com.google.oboe.samples.hellooboe;
 
 import android.app.Activity;
+import android.content.Context;
+import android.media.AudioDeviceInfo;
 import android.media.AudioManager;
 import android.os.Build;
 import android.os.Bundle;
+
+import androidx.annotation.RequiresApi;
 import androidx.core.view.MotionEventCompat;
 
 import android.view.MotionEvent;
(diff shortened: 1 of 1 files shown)


samples/hello-oboe/src/main/java/com/google/oboe/samples/hellooboe/MainActivity.java
```diff
@@ -106,6 +111,11 @@ protected void onResume() {
         mBufferSizeSpinner.setSelection(SPINNER_DEFAULT_OPTION_INDEX);
         mAudioApiSpinner.setSelection(SPINNER_DEFAULT_OPTION_INDEX);
 
+        if (mScoStarted) {
```
philburk, line 114: Can mScoStarted be true if onPause() works correctly?

philburk (commented):
Thanks for doing this.

samples/hello-oboe/src/main/java/com/google/oboe/samples/hellooboe/MainActivity.java
robertwu1: In the code directly above, onResume sets the options back to the defaults. Thus, we should reset this when this happens

robertwu1: In the code directly above, onResume sets the options back to the defaults. Thus, we should reset this when this happens

robertwu1:
Seems to work fine with just the reset onPause

philburk (approved):

PR #1491, opened 2022-02-02
Bump minSdkVersion to 21 and targetSdkVersion to 31

### LINKED ISSUE #1263: Consider dropping support for older APIs
Lollipop (API 21) and above gives over 94% coverage (according to Android Studio stats): 

image

And Marshmallow (API 23) gives 85%: 

image

I currently have zero devices which run an API <23 which means the only way I can test changes on API 16-22 is through the emulator. This is unreliable and has a [bunch of open issues](https://github.com/google/oboe/issues?q=is%3Aissue+is%3Aopen+emulator). 

I wonder if we should just drop support for these older APIs so we can spend more time improving Oboe for modern devices.

WDYT @philburk @android-usb ?

mmoczkowski:
Closes #1263

### DIFF
--- a/apps/OboeTester/app/build.gradle
+++ b/apps/OboeTester/app/build.gradle
@@ -1,11 +1,11 @@
 apply plugin: 'com.android.application'
 
 android {
-    compileSdkVersion 29
+    compileSdkVersion 31
     defaultConfig {
         applicationId = "com.mobileer.oboetester"
         minSdkVersion 23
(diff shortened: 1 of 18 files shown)


robertwu1:
I tested a subset of the apps: Oboetester, Drumthumper, and Soundboard and everything seems fine with the upgrade to 31.

I'm closing this PR. Thanks for contributing to Oboe!

PR #1503, opened 2022-03-03
DrumThumper: Make hi-hat play one or closed at any given time

### LINKED ISSUE #1502: DrumThumper Open Hat not closing
REPRODUCE
Launch DrumThumper
Tap "Open Hat" then listen to it decay.
Tap "Open Hat" then quickly tap "Close Hat".

EXPECT
OpenHat sound should stop short because of the close.
"Open Hat" and "Closed Hat" are defined as exclusive sounds in General MIDI.

ACTUAL
Open Hat keeps sounding even after the Hat is closed.

robertwu1:
Fixes #1502

### DIFF
--- a/samples/drumthumper/src/main/cpp/DrumPlayerJNI.cpp
+++ b/samples/drumthumper/src/main/cpp/DrumPlayerJNI.cpp
@@ -114,6 +114,13 @@ JNIEXPORT void JNICALL Java_com_plausiblesoftware_drumthumper_DrumPlayer_trigger
 }
 
 /**
+ * Native (JNI) implementation of DrumPlayer.trigger()
+ */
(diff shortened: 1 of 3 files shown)


philburk (commented):
Thanks for doing this. But I am concerned that this implementation will result in an abrupt transition that may be clicky.
A proper transition may require a ramp or crossfade.
I assigned this to Paul because he has the synth background and is the owner for DrumThumper.

android-usb:
Where is the request for this comming from? It's a cool feature sure, but DT's primary raison d'etre is to show techniques for minimum latency playback. Adding other stuff will just obscure that.

philburk (approved):
Sounds great.

PR #1505, opened 2022-03-03
RhythmGame: Add some instructions

### LINKED ISSUE #1504: RhythmGame: add text so it's less confusing when someone stumbles on it
If users are browsing through our samples and downloading them to their phone, using RhythmGame feels like a broken sample.

However, in fact, RhythmGame is a complete sample. The README and associated codelab are actually quite nice to read. It's just that unlike the rest of the samples, RhythmGame does not make sense as a standalone app. It needs instructions to follow along with. We should add instructions in the app somewhere

robertwu1:
This PR clarifies rhythm game to be less confusing.

Fixes #1504

### DIFF
--- a/samples/README.md
+++ b/samples/README.md
@@ -4,7 +4,8 @@ These samples demonstrate how to use the Oboe library:
 
 1. [hello-oboe](hello-oboe): creates an output (playback) stream and plays a
 sine wave when you tap the screen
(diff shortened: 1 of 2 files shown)


robertwu1:
image

philburk (approved):

Issue #1506, opened 2022-03-08
CodeQl gives overflow warning for flowgraph/resampler


ann0see:
**Short description**
CodeQl flags the following lines as high severity issues (probably they are not severe at all, therefore posting it publically): "Multiplication result converted to larger type
High
Multiplication result may overflow 'int' before it is converted to 'size_type'."
See i.e:
https://github.com/github/codeql/blob/17fbbdba34d755fa318d588732f75708aa110be5/cpp/ql/src/Likely%20Bugs/Arithmetic/IntMultToLong.ql
https://github.com/google/oboe/blob/1a7dce497ded081ed127059ae46455aa105612b6/src/flowgraph/resampler/MultiChannelResampler.cpp#L133
https://github.com/google/oboe/blob/1a7dce497ded081ed127059ae46455aa105612b6/src/flowgraph/resampler/MultiChannelResampler.cpp#L32

![image](https://user-images.githubusercontent.com/20726856/157329051-65cd3d0a-10b5-4198-b7c8-bd36b01fc127.png)
![image](https://user-images.githubusercontent.com/20726856/157329132-b80b31b4-2abb-4c39-aff5-f416b25a8fc0.png)
![image](https://user-images.githubusercontent.com/20726856/157329224-83f4caaa-9b0a-4692-8305-00084845bacf.png)
![image](https://user-images.githubusercontent.com/20726856/157329289-e315dce5-a4b1-4eb7-a30a-de42b5680842.png)

**Steps to reproduce**
Clone an app using oboe
Set up CodeQl
Compile and observe warnings from CodeQl as mentioned above
**Expected behavior**
No warnings of CodeQl. Should be fixed by smaller code-refactorings probably.
**Actual behavior**
The error log in CodeQl is warning about issues concerning overflow.
**Device**

/

**Any additional context**

/

robertwu1:
@ann0see Did you change any parameters to get CodeQl to run correctly? When I tried to create codeql.yml, the cpp checker doesn't seem to build to to mismatch in compiler flags

ann0see:
No. I think it worked quite out of the box? That’s our workflow file: https://github.com/jamulussoftware/jamulus/blob/master/.github/workflows/autobuild.yml

ann0see:
The CI https://github.com/google/oboe/runs/5504278249 shows similar issues.

ann0see:
How often do you release? I.e when can we assume these patches out live?

robertwu1:
[...]

PR #1507, opened 2022-03-08
Add static code analysis on commit and fix CodeQL errors

robertwu1:
Create an action to run static code analysis on demand.

After this is created, a separate PR will be created to fix linker issues described in #1506


robertwu1:
Seems like there are issues in the autobuilder. https://github.com/github/codeql-action/issues/824

ann0see:
What happens if you just use a normal build process instead of the auto-build provided by codeql?

robertwu1:
> What happens if you just use a normal build process instead of the auto-build provided by codeql (probably you’re already doing that though)

Changing the build to something normal seems to work fine. Thanks! I will address the issues next week

ann0see:
Ah, now similar issues show up in the CI: https://github.com/google/oboe/runs/5504278249

ann0see:
Great! Thanks for fixing these errors

ann0see:
Do you plan to add codeQl to your repo or just fix the overflow warnings? I think splitting the fixes in a separate PR would make sense such that you can decide if CodeQL should or shouldn’t be added.

robertwu1:
Is there a downside of adding CodeQl to the CI besides the checks taking a bit longer?

.github/workflows/android.yml
```diff
@@ -34,4 +34,20 @@ jobs:
         chmod +x gradlew
         ./gradlew -q clean bundleDebug
         popd
- 
+
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v1
```
philburk, line 39: You might be able to combine the Initialize step with the build step above.

apps/OboeTester/app/src/main/cpp/analyzer/BaseSineAnalyzer.h
```diff
@@ -145,8 +145,8 @@ class BaseSineAnalyzer : public LoopbackProcessor {
     bool transformSample(float sample, float referencePhase) {
         // Track incoming signal and slowly adjust magnitude to account
         // for drift in the DRC or AGC.
-        mSinAccumulator += sample * sinf(referencePhase);
-        mCosAccumulator += sample * cosf(referencePhase);
+        mSinAccumulator += static_cast<double>(sample) * static_cast<double>(sinf(referencePhase));
```
philburk, line 148: You probably only need to cast one operand. The other will get promoted automatically.

apps/OboeTester/app/src/main/cpp/analyzer/GlitchAnalyzer.h
```diff
@@ -169,8 +169,10 @@ class GlitchAnalyzer : public BaseSineAnalyzer {
                 break;
 
             case STATE_WAITING_FOR_LOCK:
-                mSinAccumulator += sample * sinf(mInputPhase);
-                mCosAccumulator += sample * cosf(mInputPhase);
+                mSinAccumulator += static_cast<double>(sample)
```
philburk, line 172: An alternative is to make mSinAccumulator a float.

apps/OboeTester/app/src/main/cpp/analyzer/LatencyAnalyzer.h
```diff
@@ -73,8 +73,8 @@ static double calculateNormalizedCorrelation(const float *a,
 
     // Correlate a against b.
     for (int i = 0; i < windowSize; i++) {
-        float s1 = a[i];
-        float s2 = b[i];
+        double s1 = a[i];
```
philburk, line 76: This code is already too slow.
Maybe better performance to make correlation, sumProducts, and sumSquares all float.

philburk (changes requested):

apps/OboeTester/app/src/main/cpp/analyzer/GlitchAnalyzer.h
robertwu1: Leaving as is because it's used later in division

philburk (approved):

robertwu1:
Fixes #1506

ann0see:
Great thanks for fixing and merging!

PR #1516, opened 2022-03-25
Measure timestamp latency for full duplex stream.

### LINKED ISSUE #1498: OboeTester should show timestamp latency for comparison with measured RT latency
This was requested by QC.
We can generate application timestamps for read and write with accurate timing in the FullDuplex callback.
These can be used by the latency calculation.

We can also do this in CTS-V.

philburk:
Compare with measured acoustic latency.
This will allow us to measure the accuracy of
the timestamps.

The reported value is "timestamp.latency.msec".

Fixes #1498

### DIFF
--- a/apps/OboeTester/app/src/main/cpp/NativeAudioContext.cpp
+++ b/apps/OboeTester/app/src/main/cpp/NativeAudioContext.cpp
@@ -659,14 +659,6 @@ jdouble ActivityRoundTripLatency::measureTimestampLatency() {
     int32_t sampleRate = mFullDuplexLatency->getOutputStream()->getSampleRate();
     int64_t mappedTimeNanos = outputTimeNanos + ((mappedPosition - ouputPosition) * 1e9) / sampleRate;
 
-//    LOGD("ActivityRoundTripLatency::%s() writeReadDelta = %ld ", __func__, writeReadDelta);
-//    LOGD("ActivityRoundTripLatency::%s() inputPosition = %ld ", __func__, inputPosition);
-//    LOGD("ActivityRoundTripLatency::%s() inputTimeNanos = %ld ", __func__, inputTimeNanos);
-//    LOGD("ActivityRoundTripLatency::%s() ouputPosition = %ld ", __func__, ouputPosition);
-//    LOGD("ActivityRoundTripLatency::%s() outputTimeNanos = %ld ", __func__, outputTimeNanos);
-//    LOGD("ActivityRoundTripLatency::%s() mappedPosition = %ld ", __func__, mappedPosition);
-//    LOGD("ActivityRoundTripLatency::%s() mappedTimeNanos = %ld ", __func__, mappedTimeNanos);
-
     // Latency is the difference in time between when a frame was recorded and
     // when its corresponding echo was played.
     return (mappedTimeNanos - inputTimeNanos) * 1.0e-6; // convert nanos to millis
--- a/apps/OboeTester/app/src/main/cpp/NativeAudioContext.cpp
+++ b/apps/OboeTester/app/src/main/cpp/NativeAudioContext.cpp
@@ -632,6 +632,46 @@ void ActivityRoundTripLatency::finishOpen(bool isInput, AudioStream *oboeStream)
     }
 }
 
+// The timestamp latency is the difference between the input
+// and output times for a specific frame.
+// Start with the position and time from an input timestamp.
+// Map the input position to the corresponding position in output
+// and calculate its time.
+// Use the difference between framesWritten and framesRead to
+// convert input positions to output positions.
+jdouble ActivityRoundTripLatency::measureTimestampLatency() {
+    if (!mFullDuplexLatency->isWriteReadDeltaValid()) return -1.0;
+
+    int64_t writeReadDelta = mFullDuplexLatency->getWriteReadDelta();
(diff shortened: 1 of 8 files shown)


apps/OboeTester/app/src/main/cpp/NativeAudioContext.cpp
```diff
@@ -632,6 +632,46 @@ void ActivityRoundTripLatency::finishOpen(bool isInput, AudioStream *oboeStream)
...
+
+    // Map input frame position to the corresponding output frame.
+    int64_t mappedPosition = inputPosition + writeReadDelta;
+    // Calculate when that frame will play.
+    int32_t sampleRate = mFullDuplexLatency->getOutputStream()->getSampleRate();
+    int64_t mappedTimeNanos = outputTimeNanos + ((mappedPosition - ouputPosition) * 1e9) / sampleRate;
+
+//    LOGD("ActivityRoundTripLatency::%s() writeReadDelta = %ld ", __func__, writeReadDelta);
+//    LOGD("ActivityRoundTripLatency::%s() inputPosition = %ld ", __func__, inputPosition);
+//    LOGD("ActivityRoundTripLatency::%s() inputTimeNanos = %ld ", __func__, inputTimeNanos);
```
robertwu1, line 664: Should we remove comments?

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/RoundTripLatencyActivity.java
```diff
@@ -291,11 +327,19 @@ private String getResultString() {
...
 
-    native int getAnalyzerProgress();
-    native int getMeasuredLatency();
     double getMeasuredLatencyMillis() {
         return getMeasuredLatency() * 1000.0 / getSampleRate();
     }
+
+    double getTimestampLatencyMillis() {
+        if (mTimestampLatencyCount <= 0) return -1.0;
+        if (mTimestampLatencyCount <= 0) return -1.0;
```
robertwu1, line 336: repeated

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/RoundTripLatencyActivity.java
```diff
@@ -195,15 +216,28 @@ public String getLastReport() {
         private Runnable runnableCode = new Runnable() {
             @Override
             public void run() {
-                String message;
+                double timestampLatency = -1.0;
+                int state = getAnalyzerState();
+                if (state == STATE_MEASURE_BACKGROUND || state == STATE_IN_PULSE) {
+                    timestampLatency = measureTimestampLatency();
+                    if (timestampLatency > 0) {
```
robertwu1, line 223: Would it be useful to keep track of the latency count somewhere in case this keeps on returning 0 or less?

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/RoundTripLatencyActivity.java
```diff
@@ -195,15 +216,28 @@ public String getLastReport() {
...
+                    }
+                    onAnalyzerDone();
                 } else {
                     message = getProgressText();
                     message += "please wait... " + counter + "\n";
-                    message += convertStateToString(getAnalyzerState());
+                    message += convertStateToString(getAnalyzerState()) + "\n";
 
                     // Repeat this runnable code block again.
                     mHandler.postDelayed(runnableCode, SNIFFER_UPDATE_PERIOD_MSEC);
```
robertwu1, line 243: Wouldn't having this run periodically skew the results towards results that have higher latency compared to if we collected every frame?

apps/OboeTester/app/src/main/cpp/NativeAudioContext.cpp
philburk: I am not convinced the algorithm is correct. So I occasionally stare at these numbers. I made them conditional.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/RoundTripLatencyActivity.java
philburk: Some configurations do not support input timestamps.
Negative latency is impossible so I use -1 to indicate an error or no measurement available.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/RoundTripLatencyActivity.java
philburk: > collected every frame

At what frame rate?

Timestamps are only collected a few times per second. We are probably already oversampling.

robertwu1: Is there a way to sample at the same frequency as the device?

Oversampling is fine. If we know we may be oversampling, we should sample even more frequently. See Nyquist's theorem

robertwu1 (approved):

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/RoundTripLatencyActivity.java
philburk: Over-sampling a function that does not change does not give you more data.
I tested and we are under-sampling. That is fine. We just want to measure the latency several time over a few seconds to see how much variation there. it does not vary much.

PR #1522, opened 2022-03-30
Live Effect: add Java effects

robertwu1:


robertwu1:
Audio Effects add quite a bit of latency and LiveEffect is meant to be an "live" demo. It is not the right place to add AudioFx effects. We are planning to add effect tests to OboeTester instead.

PR #1531, opened 2022-05-09
opensles: use dlsym() to link

### LINKED ISSUE #1530: With minSdk = 30, Oboe causes compilation error as 'slCreateEngine' is deprecated
Android version(s): 11
Android device(s): emulator
Oboe version: 917f8c7
 
**Short description**

Since Android 30, OpenSLES is deprecated, and it will cause build errors when any of our app sources use it (actually build warnings, but the default `-Werror` C++ compilation settings turns the warnings as errors). Currently Oboe uses OpenSLES without any platform tweaks, so if we specify `minSdk 30` in our app `build.gradle`, where I submodule Oboe and use `add_subdirectory()` in my `CMakeLists.txt` the build fails.

My app depends on NdkBinder and thus sets minSdk = 29, but since it has [a blocking issue with minSdk 29 with NDK r23 or later](https://issuetracker.google.com/issues/219987524), setting minSdk to 30 could be a temporary workaround. But now it is incompatible option with Oboe because of this problem.

**Steps to reproduce**

- Open OboeTester.
- change `minSdk 23` to `minSdk 30` in its `app/build.gradle`
- build

**Expected behavior**

- It still builds (maybe by dropping OpenSLES parts? Some decision agreement among the dev. team would be needed)

**Actual behavior**

- /path/to/oboe/src/opensles/EngineOpenSLES.cpp:35:18: error: 'slCreateEngine' is deprecated: first deprecated in Android 30 [-Werror,-Wdeprecated-declarations]

**Device**

- emulator, x86, API Level 31 (should not matter)

**Any additional context**

It is most likely non-issue with binary distribution (prefab) but that's not a compatible solution with other non-prefab native libraries so far.

philburk:
Link with slCreateEngine using dlsym().
OpenSL ES is deprecated in SDK >= 30.

Fixes #1530

### DIFF
--- a/apps/OboeTester/app/build.gradle
+++ b/apps/OboeTester/app/build.gradle
@@ -4,7 +4,7 @@ android {
     compileSdkVersion 31
     defaultConfig {
         applicationId = "com.mobileer.oboetester"
-        minSdkVersion 30
+        minSdkVersion 23
         targetSdkVersion 31
         // Also update the versions in the AndroidManifest.xml file.
         versionCode 57
--- a/apps/OboeTester/app/build.gradle
+++ b/apps/OboeTester/app/build.gradle
@@ -4,7 +4,7 @@ android {
     compileSdkVersion 31
     defaultConfig {
         applicationId = "com.mobileer.oboetester"
-        minSdkVersion 23
+        minSdkVersion 30
         targetSdkVersion 31
         // Also update the versions in the AndroidManifest.xml file.
         versionCode 57
--- a/src/opensles/EngineOpenSLES.cpp
+++ b/src/opensles/EngineOpenSLES.cpp
@@ -14,25 +14,60 @@
  * limitations under the License.
  */
 
+#include <dlfcn.h>
 #include "common/OboeDebug.h"
 #include "EngineOpenSLES.h"
 #include "OpenSLESUtilities.h"
 
 using namespace oboe;
 
+// OpenSL ES is deprecated in SDK 30.
+// So we use custom dynamic linking to access the library.
+#define LIB_OPENSLES_NAME "libOpenSLES.so"
+typedef SLresult  (*prototype_slCreateEngine)(
+        SLObjectItf             *pEngine,
+        SLuint32                numOptions,
+        const SLEngineOption    *pEngineOptions,
+        SLuint32                numInterfaces,
+        const SLInterfaceID     *pInterfaceIds,
+        const SLboolean         *pInterfaceRequired
+);
+static prototype_slCreateEngine gFunction_slCreateEngine = nullptr;
+static void *gLibOpenSlesLibraryHandle = nullptr;
+
+// Load the OpenSL ES library and the one primary entry point.
+static void loadOpenSLES() {
+    if (gFunction_slCreateEngine == nullptr) {
+// Use RTLD_NOW to avoid the unpredictable behavior that RTLD_LAZY can cause.
+// Also resolving all the links now will prevent a run-time penalty later.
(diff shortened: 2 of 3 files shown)


apps/OboeTester/app/build.gradle
```diff
@@ -4,7 +4,7 @@ android {
     compileSdkVersion 31
     defaultConfig {
         applicationId = "com.mobileer.oboetester"
-        minSdkVersion 23
+        minSdkVersion 30
```
philburk, line 7: Go back to 23.

robertwu1: Does changing to 30 work as expected?

src/opensles/EngineOpenSLES.cpp
```diff
@@ -14,12 +14,44 @@
...
+        const SLInterfaceID     *pInterfaceIds,
+        const SLboolean         *pInterfaceRequired
+);
+static prototype_slCreateEngine gFunction_slCreateEngine = nullptr;
+static void *gLibOpenSlesLibraryHandle = nullptr;
+
+// Load the OpenSL ES library and the one primary entry point.
+static void loadOpenSLES() {
+    if (gFunction_slCreateEngine == nullptr) {
+// Use RTLD_NOW to avoid the unpredictable behavior that RTLD_LAZY can cause.
```
robertwu1, line 41: We should indent these comments right?

src/opensles/EngineOpenSLES.cpp
```diff
@@ -30,9 +62,11 @@ SLresult EngineOpenSLES::open() {
 
     SLresult result = SL_RESULT_SUCCESS;
     if (mOpenCount++ == 0) {
+        // load the library and link to it
+        loadOpenSLES();
 
         // create engine
-        result = slCreateEngine(&mEngineObject, 0, NULL, 0, NULL, NULL);
+        result = (*gFunction_slCreateEngine)(&mEngineObject, 0, NULL, 0, NULL, NULL);
```
robertwu1, line 69: Should we check whether gFunction_slCreateEngine is null first?

robertwu1: And in the case whether it is null, what should be our backup? Can we alert the user somehow with a notification?

robertwu1:
Now that OpenSL ES is truly deprecated in some versions of Android, should we add a comment about this in the place where we decide to use either AAudio or OpenSL ES in Oboe?

apps/OboeTester/app/build.gradle
philburk: Yes, I changed minSdkVersion to 30 to repro the build failure. It failed on slCreateEngine.
With this change, it builds OK on SDK 30.
I just forgot to put minSdkVersion back to 23.

src/opensles/EngineOpenSLES.cpp
```diff
@@ -30,9 +62,11 @@ SLresult EngineOpenSLES::open() {
 
     SLresult result = SL_RESULT_SUCCESS;
     if (mOpenCount++ == 0) {
+        // load the library and link to it
+        loadOpenSLES();
```
philburk, line 66: Return an error and bail if this fails.

philburk:
Add a comment to the API choice to say OpenSL ES is deprecated.

src/opensles/EngineOpenSLES.cpp
philburk: I check the link result so now it is guaranteed to be non-null.

robertwu1 (approved):

PR #1537, opened 2022-05-16
OpenSL ES: Set usages if unspecified

### LINKED ISSUE #1536: OpenSL ES volume is broken if usage is set to Unspecified
Internal bug is b/232457419

robertwu1:
Fixes #1536

OboeTester output streams that set Unspecified works as intended now.


philburk:
Some more explanation:
AAudio converts UNSPECIFIED to Media or Music internally.
OpenSL ES was not designed to accept UNSPECIFIED so it needs to be handled in the open() method in Oboe.

PR #1545, opened 2022-05-23
OboeTester: Normalize data when correlating in latency analyzer

robertwu1:


robertwu1:
Fixes #1546 1546

robertwu1:
Abandoning since #1547 does this better

PR #1548, opened 2022-05-25
Add channel mask support for API 32.

### LINKED ISSUE #1409: add setChannelMask() for API 32
For AAudio, when API < 32 we can fail to open if channelMask requested.

OpenSL ES supports a subset of the channelMasks so we can allow that if the requested channelMask is supported.

This is needed for #1400

flamme:
For AAudio, channel mask support was added in API 32. In this change,
channel mask support is added in Oboe for API 32.

Fixes #1409.

### DIFF
--- a/include/oboe/Definitions.h
+++ b/include/oboe/Definitions.h
@@ -486,6 +486,162 @@ namespace oboe {
     };
 
     /**
+     * The channel mask of the audio stream. The underlying type is `uint32_t`.
+     * Use of this enum is convenient
+     */
+    enum class ChannelMask : uint32_t { // aaudio_channel_mask_t
+        /**
+         * Audio channel mask is not specified.
+         */
+        Unspecified = kUnspecified,
+
+        /**
+         * Channel position masks. Use the combinations of the channel position masks.
+         */
+        FRONT_LEFT = 1 << 0,
+        FRONT_RIGHT = 1 << 1,
+        FRONT_CENTER = 1 << 2,
+        LOW_FREQUENCY = 1 << 3,
+        BACK_LEFT = 1 << 4,
+        BACK_RIGHT = 1 << 5,
+        FRONT_LEFT_OF_CENTER = 1 << 6,
+        FRONT_RIGHT_OF_CENTER = 1 << 7,
+        BACK_CENTER = 1 << 8,
+        SIDE_LEFT = 1 << 9,
+        SIDE_RIGHT = 1 << 10,
+        TOP_CENTER = 1 << 11,
+        TOP_FRONT_LEFT = 1 << 12,
+        TOP_FRONT_CENTER = 1 << 13,
+        TOP_FRONT_RIGHT = 1 << 14,
+        TOP_BACK_LEFT = 1 << 15,
+        TOP_BACK_CENTER = 1 << 16,
+        TOP_BACK_RIGHT = 1 << 17,
+        TOP_SIDE_LEFT = 1 << 18,
+        TOP_SIDE_RIGHT = 1 << 19,
+        BOTTOM_FRONT_LEFT = 1 << 20,
+        BOTTOM_FRONT_CENTER = 1 << 21,
+        BOTTOM_FRONT_RIGHT = 1 << 22,
+        LOW_FREQUENCY_2 = 1 << 23,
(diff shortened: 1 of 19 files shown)


include/oboe/Definitions.h
```diff
@@ -485,6 +485,162 @@ namespace oboe {
...
+    enum class ChannelMask : uint32_t { // aaudio_channel_mask_t
+        /**
+         * Audio channel mask is not specified.
+         */
+        Unspecified = kUnspecified,
+
+        /**
+         * Channel position masks. Use the combinations of the channel position masks.
+         */
+        FRONT_LEFT = 1 << 0,
```
philburk, line 501: For consistency, these C++ enums should be Capitalized, eg. FrontLeft, LowFrequency, etc.

philburk (changes requested):
Fantastic. This was much bigger than I thought.

include/oboe/Definitions.h
flamme: I originally thought of keeping them consistent. But I finally make the single channel mask different from the combo ones as the combo ones are more designed for app usage. I was thinking of distinguish them as the using the single channel mask may be rejected by aaudio. Do you think it will be better to keep them consistent?

robertwu1: Would it be possible to use private enum classes if these are not intended to be used by developers?

flamme: We actually allow use those enums by developers if they want to make their combo in aaudio. But their combo will be rejected if it is not the ones known by aaudio.

BTW, I don't have an idea about how to make these enum private and only be used by oboe library.

robertwu1: If we are allowing use of these, we should make them consistent. Maybe we can put a disclaimer with comments

philburk (changes requested):

philburk:
I downloaded this PR and ran it on Raven.
I noticed that the downmix is clipping, which is not the fault of this PR.
I also had a problem when selecting output channelMask of 5.1.4.
Pressing OPEN and START resulted in a crash. I think this is because the sine player can only handle up to 8 channels. So an array is getting over-indexed somewhere.

philburk:
For TEST INPUT, if I select ChannelCount=3 then I get 3 inedpendant mic inputs.
If I select any channelMask with >=3 bits set, e.g. "2.1", "Tri", then I get an error. Maybe this is WAI.

flamme:
For input case, "2.1" and "Tri" are not valid selection. In that case, getting error is WAI

flamme:
> For TEST INPUT, if I select ChannelCount=3 then I get 3 independent mic inputs. If I select any channelMask with >=3 bits set, e.g. "2.1", "Tri", then I get an error. Maybe this is WAI.

For input case, "2.1" and "Tri" are not valid selection. In that case, getting error is WAI

philburk:
> For input case, "2.1" and "Tri" are not valid selection.

Is there a case when they would be valid? I am guessing no.

system/audio.h distinguishes between AUDIO_CHANNEL_IN_* and AUDIO_CHANNEL_OUT_*.
I don't think there is much use case for the input channel masks.
Maybe this can be handled with documentation.

flamme:
> I downloaded this PR and ran it on Raven. I noticed that the downmix is clipping, which is not the fault of this PR. I also had a problem when selecting output channelMask of 5.1.4. Pressing OPEN and START resulted in a crash. I think this is because the sine player can only handle up to 8 channels. So an array is getting over-indexed somewhere.

Yes, our previous max channel count is 8. I have uploaded a new patch to address this.

flamme:
> > For input case, "2.1" and "Tri" are not valid selection.
> 
> Is there a case when they would be valid? I am guessing no.
> 
> system/audio.h distinguishes between AUDIO_CHANNEL_IN_* and AUDIO_CHANNEL_OUT_*. I don't think there is much use case for the input channel masks. Maybe this can be handled with documentation.

Yes, audio.h distinguishes between input and output channel. But when AIDL is introduced, we decided not to distinguish input and output channel mask AAudioStreamParameters::validateChannelMask for the valid input channel mask. For instance, 2.0.2 is a valid one. We can add that in the aaudio documentation.

philburk (approved):
The scrolling channel boxes are nice.

PR #1550, opened 2022-05-26
opensles: clean up after failed open

### LINKED ISSUE #1541: Oboe may need handle the case that openStream failed too many times
Android version(s): 
Android device(s):
Oboe version: latest
App name used for testing: 
 
**Short description**

**Steps to reproduce**
1. disallow cellphone's microphone permission for oboe app
2. try openStream many times
3. then give permission for oboe app
4. openStream maybe failed

By investigate the oboe source code, I think here may have issue, could you help double check?

image

philburk:
There were resource leaks if the open failed.
So now it calls close() internally if open fails.

Fixes #1541

### DIFF
--- a/src/opensles/AudioInputStreamOpenSLES.cpp
+++ b/src/opensles/AudioInputStreamOpenSLES.cpp
@@ -215,6 +215,7 @@ Result AudioInputStreamOpenSLES::open() {
     return Result::OK;
 
 error:
+    close(); // Clean up various OpenSL objects and prevent resource leaks.
     return Result::ErrorInternal; // TODO convert error from SLES to OBOE
 }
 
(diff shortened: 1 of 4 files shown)


src/opensles/AudioInputStreamOpenSLES.cpp
```diff
@@ -225,7 +226,7 @@ Result AudioInputStreamOpenSLES::close() {
     if (getState() == StreamState::Closed){
         result = Result::ErrorClosed;
     } else {
-        requestStop_l();
+        (void) requestStop_l();
```
robertwu1, line 229: I haven't seem this idiom of casting to void before. I'm just curious about the reasoning behind this.

robertwu1 (approved):

src/opensles/AudioInputStreamOpenSLES.cpp
philburk: It says "I know this function returns a result code. But I don't care because I will just keep going even if it returns an error code. So I am not even going to look at it.".

PR #1553, opened 2022-05-27
OboeTester: remove extra \n in xRun#

robertwu1:
Before:
image

After:
image


philburk (approved):
Thanks. I wondered where that strange symbol came from.

PR #1575, opened 2022-07-13
Rename instances of master with main

robertwu1:
Now that we have officially removed the master branch, we should change instances of master to main.

Without this change, there will be an extra automatic link redirection from master to main by Github.


philburk:
What are "instances of master"?
Are these links in the markdown docs?

apps/OboeTester/app/src/main/cpp/util/WaveFileWriter.h
```diff
@@ -15,7 +15,7 @@
  */
 
 // Based on the WaveFileWriter in Java from the open source JSyn library by Phil Burk
-// https://github.com/philburk/jsyn/blob/master/src/com/jsyn/util/WaveFileWriter.java
+// https://github.com/philburk/jsyn/blob/master/src/main/java/com/jsyn/util/WaveFileWriter.java
```
philburk, line 18: JSyn is still using a "master" branch.

philburk (changes requested):

apps/OboeTester/app/src/main/cpp/util/WaveFileWriter.h
robertwu1: Yup, hence this is https://github.com/philburk/jsyn/blob/master/src/main/java/com/jsyn/util/WaveFileWriter.java and not https://github.com/philburk/jsyn/blob/master/src/master/java/com/jsyn/util/WaveFileWriter.java

philburk: Ah! So there is a "main" folder in the "master" branch in JSyn. Got it.

philburk (approved):

PR #1579, opened 2022-07-25
HelloOboe: Use setCommunicationDevice instead of startBluetoothSco

robertwu1:
startBluetoothSco is deprecated in Android T. Instead, we should use setCommunicationDevice.

This PR also cleans up a couple of other warnings from the compiler.

I will update https://github.com/google/oboe/wiki/TechNote_BluetoothAudio after this PR is submitted.


philburk (approved):
Awesome.

PR #1583, opened 2022-08-02
HelloOboe: Disable playback device spinners on OpenSL ES

### LINKED ISSUE #1102: Use of AudioDeviceCallback to select device in samples requires API 23
The device selection code in the samples and in OboeTester use AudioDeviceCallback.
That was added in API 23 so the samples will not run on Lollipop ( API 21,22)

robertwu1:
Fixes #1102

Tested on Nexus 6 with Android API 22.


philburk (changes requested):
Device selection not supported for OpenSLES even on new versions.

philburk (approved):

PR #1584, opened 2022-08-02
fxlab: Restore effect UI when placement changes and on screen rotation

### LINKED ISSUE #834: [fxlab] Screen rotation causes all effects to be lost from view
Repro steps: 
- Add effect
- Rotate screen

Expected: 
- Effects stay in list

Actual:
- Effects are removed

### LINKED ISSUE #1120: FxLab Effect UI state is not restored upon placement in list
Android version(s): Android Pie
Android device(s): Samsung Galaxy S8
Oboe version: master
App name used for testing: FxLab
 
**Short description**
Effect UI state is not restored upon placement in list

**Steps to reproduce**
add an effect, add another effect, change some parameters in the effect, move the effect to another position in the list

**Expected behavior**
the effect's UI's current state should persist

**Actual behavior**
the effect's UI is reset to its initial state

robertwu1:
This PR fixes two bugs relating to fxlab settings not being persistent.
1. Using effectList[index].paramValues[counter] instead of param.defaultValue allows UI to restore on placement swaps.
2. Adding android:configChanges="orientation|screenSize|keyboardHidden" allows effects to stay and audio engine to not be restarted when screen is rotated.

Fixes #1120
Fixes #834

### DIFF
--- a/apps/fxlab/app/build.gradle
+++ b/apps/fxlab/app/build.gradle
@@ -23,7 +23,7 @@ apply plugin: 'kotlin-android-extensions'
 apply plugin: 'kotlin-kapt'
 
 android {
-    compileSdkVersion 29
+    compileSdkVersion 33
     compileOptions {
         sourceCompatibility JavaVersion.VERSION_1_8
         targetCompatibility JavaVersion.VERSION_1_8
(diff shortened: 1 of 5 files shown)


philburk (changes requested):
Consider alternatives to android:configChanges, audio from a persistent object

robertwu1:
> Consider alternatives to android:configChanges, audio from a persistent object

After looking at the alternatives, android:configChanges seems like the correct solution. Changing audio to be a persistent object is non-trivial. Even updating the UI is not easy as we need to save an array as a persistent object.

Let's make the android:configChanges change and if we have many free cycles in the future, we should consider the alternative.

philburk (approved):

PR #1588, opened 2022-08-04
OboeTester: Update Test Output channel box and signal UI

### LINKED ISSUE #1587: OboeTester Test Output Channels and Signal UI show when they shouldn't
In Test Output, the bottom two things we can change seem to do the wrong thing if we change them at the wrong time.

For output channels, changing the channel checkbox between open and start seems to do nothing.

For signals, this does not do anything after a stream has started.

robertwu1:
Fixes #1587

### DIFF
--- a/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestOutputActivity.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestOutputActivity.java
@@ -92,8 +92,6 @@ public final class TestOutputActivity extends TestOutputActivityBase {
 
     public void openAudio() throws IOException {
         super.openAudio();
-        int channelCount = mAudioOutTester.getCurrentAudioStream().getChannelCount();
-        configureChannelBoxes(channelCount);
     }
 
     private void configureChannelBoxes(int channelCount) {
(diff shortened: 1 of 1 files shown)


apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestOutputActivity.java
```diff
@@ -105,9 +103,17 @@ private void configureChannelBoxes(int channelCount) {
...
         configureChannelBoxes(0);
+        mNativeApiSpinner.setEnabled(true);
         super.closeAudio();
     }
 
+    public void startAudio() throws IOException {
+        super.startAudio();
+        int channelCount = mAudioOutTester.getCurrentAudioStream().getChannelCount();
+        configureChannelBoxes(channelCount);
+        mNativeApiSpinner.setEnabled(false);
```
philburk, line 114: mNativeApiSpinner is badly named, by me, so this fix is confusing.
Please rename to mOutputSignalSpinner or something better.
Also rename NativeApiSpinnerListener.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestOutputActivity.java
```diff
@@ -105,9 +103,17 @@ private void configureChannelBoxes(int channelCount) {
 
     public void closeAudio() {
         configureChannelBoxes(0);
+        mNativeApiSpinner.setEnabled(true);
```
philburk, line 106: This stuff should be moved to stopAudio().

philburk (changes requested):

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestOutputActivity.java
philburk: And in pauseAudio().

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestOutputActivity.java
philburk: And pauseAudio()

robertwu1: Changed to all 3. Seems like we need it for pause, stop, and close.

philburk (approved):

PR #1590, opened 2022-08-04
Quirks: Include Vivo and check for Oboe workaround flag

### LINKED ISSUE #201: don't use FLOAT on Vivo devices before M
It was not properly supported. Use PCM16 or reject.

Try  __system_property_get("ro.product.manufacturer", buf);

See https://www.google.com/url?q=https://crbug.com/737188

### LINKED ISSUE #1589: QuirksManager doesn't respect OboeGlobals::areWorkaroundsEnabled()
Several places in QuirksManager.cpp do not respect areWorkaroundsEnabled(). This does not cause any changes to the end user.

This means that OboeTester is doing sample rate conversion and I16 to Float conversions even when the areWorkaroundsEnabled() checkbox is not checked.

I'm not completely sure whether this is a bug or WAI.

robertwu1:
Fixes #201
Fixes #1589

I don't have a Vivo to test this.

I did test Oboe workarounds in OboeTester on a Pixel phone.

### DIFF
--- a/src/common/QuirksManager.cpp
+++ b/src/common/QuirksManager.cpp
@@ -223,11 +223,11 @@ bool QuirksManager::isConversionNeeded(
         conversionNeeded = true;
     }
 
-    // If a SAMPLE RATE is specified for low latency then let the native code choose an optimal rate.
+    // If a SAMPLE RATE is specified for low latency, let the native code choose an optimal rate.
+    // This isn't really a workaround. It is an Oboe feature that is convenient to place here.
     // TODO There may be a problem if the devices supports low latency
     //      at a higher rate than the default.
-    if (OboeGlobals::areWorkaroundsEnabled()
-            && builder.getSampleRate() != oboe::Unspecified
+    if (builder.getSampleRate() != oboe::Unspecified
             && builder.getSampleRateConversionQuality() != SampleRateConversionQuality::None
             && isLowLatency
             ) {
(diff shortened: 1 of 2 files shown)


src/common/QuirksManager.cpp
```diff
@@ -238,14 +250,16 @@ bool QuirksManager::isConversionNeeded(
     }
 
     // Add quirk for float output on API <21
```
philburk, line 252: Change "on API <21" to "where needed".

src/common/QuirksManager.cpp
```diff
@@ -216,7 +226,8 @@ bool QuirksManager::isConversionNeeded(
     // If a SAMPLE RATE is specified for low latency then let the native code choose an optimal rate.
     // TODO There may be a problem if the devices supports low latency
     //      at a higher rate than the default.
-    if (builder.getSampleRate() != oboe::Unspecified
+    if (OboeGlobals::areWorkaroundsEnabled()
```
philburk, line 229: Don't check areWorkaroundsEnabled() here.
We always want to convert in Oboe if the app requests Oboe SRC.
This is more of oboe feature than a bug workaround.

philburk (changes requested):

src/common/QuirksManager.cpp
robertwu1: Should the other two locations be changed or not?

robertwu1: I changed left sample rate default as is and changed Int16 -> Float for input and output streams.

philburk (approved):

PR #1593, opened 2022-08-05
Add API to set sleep time before close.

### LINKED ISSUE #1500: improve sleeptime between stop() and close()
There are ongoing problems with callbacks running after the stop.

In AudioStreamAAudio::close() we stop the stream and then sleep before closing the stream.
    
            usleep(kDelayBeforeCloseMillis * 1000);

We should a call that allows the developer to change that sleep time.

Also the default should be based on the framesPerBurst, which may be 20 msec on some devices.

philburk:
Add: AudioStream::setDelayBeforeCloseMillis(delay)

This can be used to avoid use-after-free bugs
cause by callbacks running when close is called.

Fixes #1500

### DIFF
--- a/src/aaudio/AudioStreamAAudio.cpp
+++ b/src/aaudio/AudioStreamAAudio.cpp
@@ -328,6 +328,8 @@ Result AudioStreamAAudio::open() {
             static_cast<int>(mFormat), static_cast<int>(mSampleRate),
             static_cast<int>(mBufferCapacityInFrames));
 
+    calculateDefaultDelayBeforeCloseMillis();
+
 error2:
     mLibLoader->builder_delete(aaudioBuilder);
     LOGD("AudioStreamAAudio.open: AAudioStream_Open() returned %s",
@@ -355,12 +357,7 @@ Result AudioStreamAAudio::close() {
             // Make sure we are really stopped. Do it under mLock
             // so another thread cannot call requestStart() right before the close.
             requestStop_l(stream);
-            // Sometimes a callback can occur shortly after a stream has been stopped and
-            // even after a close! If the stream has been closed then the callback
-            // can access memory that has been freed. That causes a crash.
-            // This seems to be more likely in Android P or earlier.
-            // But it can also occur in later versions.
-            usleep(kDelayBeforeCloseMillis * 1000);
+            sleepBeforeClose();
         }
         return static_cast<Result>(mLibLoader->stream_close(stream));
     } else {
--- a/src/aaudio/AudioStreamAAudio.h
+++ b/src/aaudio/AudioStreamAAudio.h
@@ -117,9 +117,12 @@ private:
      */
     void launchStopThread();
 
-    // Time to sleep in order to prevent a race condition with a callback after a close().
-    // Two milliseconds may be enough but 10 msec is even safer.
-    static constexpr int kDelayBeforeCloseMillis = 10;
+public:
+    int32_t getMDelayBeforeCloseMillis() const;
+
+    void setDelayBeforeCloseMillis(int32_t mDelayBeforeCloseMillis);
+
+private:
 
     std::atomic<bool>    mCallbackThreadEnabled;
     std::atomic<bool>    mStopThreadAllowed{false};
--- a/src/common/AudioStream.cpp
+++ b/src/common/AudioStream.cpp
@@ -196,4 +196,13 @@ ResultWithValue<FrameTimestamp> AudioStream::getTimestamp(clockid_t clockId) {
     }
 }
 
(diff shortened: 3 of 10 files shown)


src/common/AudioStream.cpp
```diff
@@ -196,4 +196,13 @@ ResultWithValue<FrameTimestamp> AudioStream::getTimestamp(clockid_t clockId) {
     }
 }
 
+void AudioStream::calculateDefaultDelayBeforeCloseMillis() {
```
robertwu1, line 199: Are we guaranteed that calculateDefaultDelayBeforeCloseMillis() runs before the user gets an AudioStreamPointer? If not, setDelayBeforeCloseMillis() may be overriden by calculateDefaultDelayBeforeCloseMillis(). Can we add some comments about the timing windows?

src/aaudio/AudioStreamAAudio.cpp
```diff
@@ -355,12 +357,7 @@ Result AudioStreamAAudio::close() {
             // Make sure we are really stopped. Do it under mLock
             // so another thread cannot call requestStart() right before the close.
             requestStop_l(stream);
```
robertwu1, line 359: It's a bit inconsistent that requestStop_l() in AAudio is in OboeGlobals::areWorkaroundsEnabled() but requestPause_l() in OpenSL ES isn't. Should we change one of them to make them consistent?

robertwu1 (changes requested):
Can run some tests before/after the changes and verify that the timeout increased a lot if you call setDelayBeforeCloseMillis() on a stream? This might be messy to add as a unit test but a quick sanity check before submitting would be good.

philburk:
> Can run some tests before/after the changes and verify that the timeout increased a lot if you call

"timeout"? Are you suggesting that we measure the time it takes to close a stream and assert that the time increases when we call setDelayBeforeCloseMillis() with a higher value?

src/aaudio/AudioStreamAAudio.cpp
philburk: True. I will change the openSL code.

philburk: On second thought. This could cause a regression in some app that has disabled workarounds when using OpenSL ES.  The AAudio code is not supposed to need the pause but we found it does so we consider the pause a workaround. I will leave as-is. But I noticed I forgot to call the sleep from the Input side so I added that.

robertwu1:
> > Can run some tests before/after the changes and verify that the timeout increased a lot if you call
> 
> "timeout"? Are you suggesting that we measure the time it takes to close a stream and assert that the time increases when we call setDelayBeforeCloseMillis() with a higher value?

This test works if we set setDelayBeforeCloseMillis() to a ridiculous number like 1000ms. Otherwise, the test may be too noisy.

Alternatively, you can simply change [testReturnStop.cpp](https://github.com/google/oboe/blob/main/tests/testReturnStop.cpp) locally to call setDelayBeforeCloseMillis() as a sanity check.

philburk:
I added a test that detects the increase in time required to close the stream.
The test was accurate to within a millisecond but I use a bigger tolerance.

robertwu1:
> I added a test that detects the increase in time required to close the stream. The test was accurate to within a millisecond but I use a bigger tolerance.

Great! Are you planning to push your test?

philburk:
> Great! Are you planning to push your test?
Oops. I pushed the wrong branch.
Test now uploaded.

src/aaudio/AudioStreamAAudio.h
```diff
@@ -117,9 +117,12 @@ class AudioStreamAAudio : public AudioStream {
      */
     void launchStopThread();
 
-    // Time to sleep in order to prevent a race condition with a callback after a close().
-    // Two milliseconds may be enough but 10 msec is even safer.
-    static constexpr int kDelayBeforeCloseMillis = 10;
+public:
+    int32_t getMDelayBeforeCloseMillis() const;
```
robertwu1, line 121: Should we make this uint32_t?

robertwu1 (approved):
Left some minor comments. Feel free to submit either without addressing my comments or after addressing my comments.

src/aaudio/AudioStreamAAudio.h
philburk: The [Google C++ style guide](go/cstyle#Integer_Types) says:

> You should not use the unsigned integer types such as uint32_t, unless there is a valid reason
> such as representing a bit pattern rather than a number, or you need defined overflow
> modulo 2^N. In particular, do not use unsigned types to say a number will never be negative.
> Instead, use [assertions](https://engdoc.corp.google.com/eng/doc/devguide/cpp/primer#google_assert) for this.

PR #1594, opened 2022-08-05
OboeTester: use 2 burst min for glitch test

### LINKED ISSUE #1560: OboeTester Glitch test should use BufferSize of 2 bursts
On Raven, the AutoGlitch test fails on test 10, which is not low latency.
The burst size is 1924 and the buffer size is 1924.
We should use a minimum of two bursts for BufferSize

philburk:
Was using 1/2 capacity, which is sometimes only one burst.

Fixes #1560

### DIFF
--- a/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/BaseAutoGlitchActivity.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/BaseAutoGlitchActivity.java
@@ -257,6 +257,7 @@ public class BaseAutoGlitchActivity extends GlitchActivity {
             // Set output size to a level that will avoid glitches.
             AudioStreamBase outStream = mAudioOutTester.getCurrentAudioStream();
             int sizeFrames = outStream.getBufferCapacityInFrames() / 2;
+            sizeFrames = Math.max(sizeFrames, 2 * outStream.getFramesPerBurst());
             outStream.setBufferSizeInFrames(sizeFrames);
             AudioStreamBase inStream = mAudioInputTester.getCurrentAudioStream();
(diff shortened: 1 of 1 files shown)


apps/OboeTester/app/src/main/java/com/mobileer/oboetester/BaseAutoGlitchActivity.java
```diff
@@ -257,6 +257,7 @@ protected TestResult testConfigurations() throws InterruptedException {
             // Set output size to a level that will avoid glitches.
             AudioStreamBase outStream = mAudioOutTester.getCurrentAudioStream();
             int sizeFrames = outStream.getBufferCapacityInFrames() / 2;
+            sizeFrames = Math.max(sizeFrames, 2 * outStream.getFramesPerBurst());
```
robertwu1, line 260: If 2 * outStream.getFramesPerBurst() is greater than outStream.getBufferCapacityInFrames(), we may overflow.

Can we add an additional Math.min(sizeFrames, outStream.getBufferCapacityInFrames())?

robertwu1 (changes requested):

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/BaseAutoGlitchActivity.java
philburk: setBufferSizeInFrames() is clipped in Oboe and AAudio against the capacity. So this is safe.
It returns the actual size after clipping and quantizing to multiples of the burst size.

Also we try to ensure the capacity is at least two bursts.

robertwu1 (approved):

Issue #952, opened 2020-07-21
Huawei P30: AAudio a sound is corrupted and OpenSL has a high latency


banburubi:
Android version(s): 10 (10.1.0.123 (C431E22R2P5))
Android device(s): Huawei P30 (ELE-L29)
Oboe version: 1.4.2
App name used for testing: Similar to LiveEffect
 
**Short description**
**Steps to reproduce**
In voice related application, I'm working on, I'm having some issues on Huawei P30 related to how AAudio and OpenSL behave. On other tested phones of different vendors it works without any issues.

The application has almost the same setup as LiveEffect. There are 2 streams: one for input and another for output, both a requested in Exclusive, Low-latency, I16 mode. Input stream has 1 channel, Output stream has 2 channels. Callback is attached to the output stream. Sample rate and frames per callback are set using this method: https://github.com/google/oboe/blob/master/docs/GettingStarted.md#obtaining-optimal-latency

After requesting streams I'm getting such parameters:

========
AAudio:

Output:
- Frames per burst: 240
- Buffer size in frames: 480
- Channels: 2
- Sample rate: 48000
- Sharing mode: Shared
- Performance mode: Low-latency

Input:
- Frames per burst: 240
- Buffer size in frames: 3840
- Channels: 1
- Sample rate: 48000
- Sharing mode: Shared
- Performance mode: Low-latency

========
OpenSL:

Output:
- Frames per burst: 960
- Buffer size in frames: 1920
- Channels: 2
- Sample rate: 48000
- Sharing mode: Shared
- Performance mode: None

Input:
- Frames per burst: 960
- Buffer size in frames: 15360
- Channels: 1
- Sample rate: 48000
- Sharing mode: Shared
- Performance mode: None

========

**Expected behavior**
On Huawei P30: With AAudio the sound is not corrupted. With OpenSL the latency is low. 

**Actual behavior**
On Huawei P30: With AAudio the sound is corrupted. With OpenSL the latency is high.

Looking forward to your feedback.

Thank you very much.

P.S. Probably these issues are related:
https://github.com/google/oboe/issues/525
https://github.com/google/oboe/issues/386

Best,
Vladymyr.

philburk:
Some Huawei devices do not support mono recording properly. Try recording in stereo and then just use the first channel.

Some Huawei devices disable the FAST mixer so they will have higher than expected latency for OpenSL ES.

We can try to workaround the mono input issue. Can you please tell me what you get for these properties?

[...]

PR #1601, opened 2022-08-06
opensles: implement capacity using buffer queue length

### LINKED ISSUE #744: Adjust OpenSL ES buffer queue length based on capacity
Change kBufferQueueLength constant to mBufferQueueLength variable.
Set mBufferQueueLength based on capacity and callbacksize.

See experiment on the effect of kBufferQueueLength on latency in:
[Why Oboe only enqueue one buffer? #464](https://github.com/google/oboe/issues/464)

[Understanding FramesPerCallback / kBufferQueueLength / Buffer Size #736](https://github.com/google/oboe/issues/736)

According to this internal bug:

b/27819623 | Allow OpenSL ES to specify a total buffer size for a fast track

starting in N, adding more buffers would improve glitch protection. So this change would help devices running 7.0, 7.1 and 8.0. That would be about 1/3 of Android devices. For 8.1 and later, AAudio is available.

[Update: as of August 2022, 17.5% of devices are running < 8.1]

philburk:
The OpenSL ES buffer queue length controls the allocated size of
the internal buffer. So we can use that to make sure there is enough
capacity to protect against glitches when setting a callback size
or a high capacity.

Fixes #744
It might help with #952


src/opensles/AudioStreamOpenSLES.cpp
```diff
@@ -123,35 +123,61 @@ SLresult AudioStreamOpenSLES::finishCommonOpen(SLAndroidConfigurationItf configI
...
-
-        // Calculate the size of a fixed duration high latency buffer based on sample rate.
-        int32_t framesPerHighLatencyBuffer =
-                (kHighLatencyBufferSizeMillis * sampleRate) / kMillisPerSecond;
+int32_t AudioStreamOpenSLES::calculateOptimalBufferQueueLength() {
+    int32_t queueLength = kBufferQueueLengthDefault;
+    int32_t likelyFramesPerBurst = estimateNativeFramesPerBurst();
+    // What is required based on capacity?
+    if (mBufferCapacityInFrames > 0) {
+        int32_t queueLengthFromCapacity = (mBufferCapacityInFrames + likelyFramesPerBurst - 1)
```
robertwu1, line 131: If we are rounding down here, would there be a problem if someone writes the full capacity?

src/opensles/AudioStreamOpenSLES.cpp
```diff
@@ -123,35 +123,61 @@ SLresult AudioStreamOpenSLES::finishCommonOpen(SLAndroidConfigurationItf configI
...
+                / likelyFramesPerBurst;
+        queueLength = std::max(queueLength, queueLengthFromCapacity);
+    }
+    // What is required based on requested callback size?
+    if (mFramesPerCallback > 0) {
+        int32_t minCapacity = 2 * mFramesPerCallback;
```
robertwu1, line 137: Should we use a constant for this 2? Or at least a comment on double buffering?

src/opensles/AudioStreamOpenSLES.cpp
```diff
@@ -123,35 +123,61 @@ SLresult AudioStreamOpenSLES::finishCommonOpen(SLAndroidConfigurationItf configI
...
+ * The best information we have is if DefaultStreamValues::FramesPerBurst
+ * was set by the app based on AudioManager.PROPERTY_OUTPUT_FRAMES_PER_BUFFER.
+ * Without that we just have to guess.
+ * @return
+ */
+int32_t AudioStreamOpenSLES::estimateNativeFramesPerBurst() {
+    int32_t framesPerBurst = DefaultStreamValues::FramesPerBurst;
+    framesPerBurst = std::max(framesPerBurst, 16);
+    // Calculate the size of a fixed duration high latency buffer based on sample rate.
+    int32_t sampleRate = (mSampleRate > 0) ? mSampleRate : 48000;
```
robertwu1, line 157: Should we use DefaultStreamValues::SampleRate as the backup here and 48000 as the backup to the backup?

src/opensles/AudioStreamOpenSLES.cpp
```diff
@@ -123,35 +123,61 @@ SLresult AudioStreamOpenSLES::finishCommonOpen(SLAndroidConfigurationItf configI
...
-            int32_t numBursts = (framesPerHighLatencyBuffer + mFramesPerBurst - 1) / mFramesPerBurst;
-            mFramesPerBurst *= numBursts;
-            LOGD("AudioStreamOpenSLES:%s() NOT low latency, set mFramesPerBurst = %d",
-                 __func__, mFramesPerBurst);
-        }
-        mFramesPerCallback = mFramesPerBurst;
-    }
+            && framesPerBurst < framesPerHighLatencyBuffer) {
+        // Find a multiple of framesPerBurst >= framesPerHighLatencyBuffer.
+        int32_t numBursts = (framesPerHighLatencyBuffer + framesPerBurst - 1) / framesPerBurst;
```
robertwu1, line 166: Seems like we are using this method for finding a multiple in the code in quite a few places. Can we write a short helper function for this so the code is easier to read?

robertwu1 (changes requested):
Added some comments. Thanks!

src/opensles/AudioStreamOpenSLES.cpp
philburk: We are rounding **up** based on the burst size.

src/opensles/AudioStreamOpenSLES.cpp
philburk: Good idea.

src/opensles/AudioStreamOpenSLES.cpp
philburk: OK

src/opensles/AudioStreamOpenSLES.cpp
philburk: Done

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
```diff
@@ -67,7 +67,8 @@ abstract class TestAudioActivity extends Activity {
     public static final int ACTIVITY_DATA_PATHS = 8;
 
     private int mAudioState = AUDIO_STATE_CLOSED;
-    protected int audioManagerFramesPerBurst;
+//    protected int audioManagerSampleRate;
```
philburk, line 70: remove

robertwu1, line 70: nit: remove these

src/opensles/AudioStreamOpenSLES.cpp
```diff
@@ -326,7 +357,7 @@ Result AudioStreamOpenSLES::close_l() {
 SLresult AudioStreamOpenSLES::enqueueCallbackBuffer(SLAndroidSimpleBufferQueueItf bq) {
     SLresult result = (*bq)->Enqueue(
             bq, mCallbackBuffer[mCallbackBufferIndex].get(), mBytesPerCallback);
-    mCallbackBufferIndex = (mCallbackBufferIndex + 1) % kBufferQueueLength;
+    mCallbackBufferIndex = (mCallbackBufferIndex + 1) % kBufferQueueLengthMax;
```
robertwu1, line 360: Should this use mBufferQueueLength in any way? This seems to put data into unused buffers if mBufferQueueLength is small

philburk: This code cycles through allocated buffers. If we do then we should also only allocate mBufferQueueLength. 
The reality of how all this works is not what one would expect. We do not actually queue multiple buffers into the linked list because the list is in the sender thread and basically useless. The "length" is only used as an indirect signal to AudioFlinger to tell it how large a FIFO to allocate.

robertwu1 (approved):

robertwu1:
Feel free to submit once you addressed my remaining comments.

src/opensles/AudioStreamOpenSLES.cpp
philburk: I used mBufferQueueLength to save memory.

PR #1602, opened 2022-08-09
Add OboeExtensions

### LINKED ISSUE #1032: Provide a way to query MMap info from stream
We want to know whether the device supports mmap and also whether our stream uses mmap or not. This will help us to diagnose issues with the stream especially for device specific issues. It will also be useful while reporting issues here.

Currently [AudioStream](https://github.com/google/oboe/blob/master/include/oboe/AudioStream.h) interface provides lots of useful information about the stream such as `isXRunCountSupported()`, `getXRunCount()`, etc. We still miss something like `isMMapUsed()` and `isMMapAvailable()` on this interface.

We looked at MMap related code from QuirkManager and OboeTester. It is not trivial to get this information from the stream. A naive attempt calling `QuirksManager::isMMapUsed()` resulted in a crash on our aaudio device which doesn't support mmap. Apparently, calling  `stream_isMMapUsed` on aaudio internally caused this crash (But, we could get `stream_getFramesPerBurst()` working fine with similar code). OTOH, OboeTester does lot more to get this information. We don't know if we should be using similar approach in our production code.

In short, it would be nice if `AudioStream` provides `isMMapUsed()`. Similarly, `isMMapAvailable()` on suitable interface to learn about the device capability would help.

robertwu1:
Fixes #1032


src/opensles/AudioStreamBuffered.h
```diff
@@ -55,6 +55,10 @@ class AudioStreamBuffered : public AudioStream {
 
     bool isXRunCountSupported() const override;
 
+    bool isMMapUsed() override {
```
philburk, line 58: You could declare this in AudioStream.h as default and then only override it in the AAudio subclass.

include/oboe/AudioStreamBuilder.h
```diff
@@ -499,6 +499,16 @@ class AudioStreamBuilder : public AudioStreamBase {
                 || (mAudioApi == AudioApi::Unspecified && isAAudioRecommended());
     }
 
+    /**
+     * @return true if MMAP will be used based on the current settings.
```
philburk, line 503: We don't know if MMAP will be used until we call AAudio. We can only report whether it is supported on the device.

tests/testStreamOpen.cpp
```diff
@@ -444,3 +444,49 @@ TEST_F(StreamOpenOutput, OutputForOpenSLESPerformanceModeNoneGetBufferSizeInFram
...
+    ASSERT_TRUE(openStream());
+    if (mStream->isMMapUsed()){
+        EXPECT_TRUE(mBuilder.willUseMMap());
+    }
+    ASSERT_TRUE(closeStream());
+}
+
+TEST_F(StreamOpenOutput, OutputMmapExclusiveSupportedShouldMeanItsSupported){
+    mBuilder.setDirection(Direction::Input);
+    if (mBuilder.willUseExclusiveMMap()){
```
philburk, line 482: These tests that expect MMAP are not valid because the code does not make predictions about what will be used. The code is only reporting device support. Also some devices only support MMAP for output on headsets and not on other devices.

philburk (changes requested):
We should only expose isMMapUsed()
isMMapEnabled()
setMMapEnabled()
and isMMapSupported.
Probably in another class like OboeGlobals or OboeExtensions or OboeTest classes.

robertwu1:
> We should only expose isMMapUsed() isMMapEnabled() setMMapEnabled() and isMMapSupported. Probably in another class like OboeGlobals or OboeExtensions or OboeTest classes.

Good idea! I added OboeExtensions with these.

tests/testStreamOpen.cpp
```diff
@@ -444,3 +444,25 @@ TEST_F(StreamOpenOutput, OutputForOpenSLESPerformanceModeNoneGetBufferSizeInFram
...
+        ASSERT_EQ(OboeExtensions::setMMapEnabled(true), 0);
+        ASSERT_TRUE(OboeExtensions::isMMapEnabled());
+        ASSERT_TRUE(openStream());
+        EXPECT_TRUE(OboeExtensions::isMMapUsed(mStream));
+        ASSERT_TRUE(closeStream());
+
+        ASSERT_EQ(OboeExtensions::setMMapEnabled(false), 0);
+        ASSERT_FALSE(OboeExtensions::isMMapEnabled());
+        ASSERT_TRUE(openStream());
+        EXPECT_FALSE(OboeExtensions::isMMapUsed(mStream));
```
philburk, line 459: This will work. It's a good test.

include/oboe/OboeExtensions.h
```diff
@@ -0,0 +1,51 @@
...
+
+#include <stdint.h>
+
+#include "oboe/Definitions.h"
+#include "oboe/AudioStream.h"
+
+namespace oboe {
+
+/**
+ * Extra Oboe APIs to query and set whether MMAP is enabled.
```
philburk, line 28: Comment based on AAudioTesting.h
    The definitions below are only for testing. 
    They are not recommended for use in an application.
    They may change or be removed at any time.

philburk:
What happens if you setMMapEnabled(true) on a device that does not support MMAP?

robertwu1:
> What happens if you setMMapEnabled(true) on a device that does not support MMAP?

Open returns an error.

I addressed your comments. Thanks!

philburk (approved):

PR #1613, opened 2022-08-29
OboeTester: Add ability to add audio effects from stream configs

### LINKED ISSUE #1527: Add AudioFx examples to OboeTester
Should add AEC to Echo test, a new test for output effects, and a new test for input effects.

LiveEffect is not the right place to add effects as output effects don't work with low latency streams.

See #1522 as an example

robertwu1:
Currently, java audio effects are neither stable nor easily testable. This PR adds some basic audio effects to audio stream configs. This way, audio effects can be added to most existing tests.

Fixes #1527

Below are some screenshots:

Screenshot_20220829-164128
Screenshot_20220829-164116

### DIFF
--- a/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/EchoActivity.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/EchoActivity.java
@@ -204,8 +204,4 @@ public class EchoActivity extends TestInputActivity {
     boolean isOutput() {
         return false;
     }
-
-    @Override
-    public void setupEffects(int sessionId) {
-    }
 }
--- a/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/StreamConfigurationView.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/StreamConfigurationView.java
@@ -512,6 +512,7 @@ public class StreamConfigurationView extends LinearLayout {
             mLoudnessEnhancer = new LoudnessEnhancer(sessionId);
             mLoudnessEnhancer.setTargetGain((short) mLoudnessEnhancerSeekBar.getProgress());
         } else {
+            // If AEC is not available, the checkbox will be disabled in initializeViews().
             if (mAcousticEchoCanceler.getEnabled()) {
                 mAcousticEchoCanceler = AcousticEchoCanceler.create(sessionId);
                 if (mAcousticEchoCanceler != null) {
@@ -520,6 +521,7 @@ public class StreamConfigurationView extends LinearLayout {
                     Log.e(TAG, String.format("Could not create AcousticEchoCanceler"));
                 }
             }
+            // If AGC is not available, the checkbox will be disabled in initializeViews().
             if (mAutomaticGainControl.getEnabled()) {
                 mAutomaticGainControl = AutomaticGainControl.create(sessionId);
                 if (mAutomaticGainControl != null) {
--- a/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/StreamConfigurationView.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/StreamConfigurationView.java
@@ -18,12 +18,17 @@ package com.mobileer.oboetester;
 
 import android.content.Context;
 import android.media.AudioManager;
+import android.media.audiofx.AcousticEchoCanceler;
+import android.media.audiofx.AutomaticGainControl;
+import android.media.audiofx.BassBoost;
+import android.media.audiofx.LoudnessEnhancer;
 import android.util.AttributeSet;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.widget.AdapterView;
 import android.widget.ArrayAdapter;
 import android.widget.CheckBox;
+import android.widget.SeekBar;
 import android.widget.Spinner;
 import android.widget.TableRow;
 import android.widget.TextView;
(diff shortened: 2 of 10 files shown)


apps/OboeTester/app/src/main/java/com/mobileer/oboetester/EchoActivity.java
```diff
@@ -204,8 +204,4 @@ public void onStopEcho(View view) {
     boolean isOutput() {
         return false;
     }
-
-    @Override
-    public void setupEffects(int sessionId) {
```
philburk, line 209: If an empty method was declared in TestAudioActivity, instead of an abstract method, then all these empty Overrides would not be needed.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/StreamConfigurationView.java
```diff
@@ -418,4 +486,55 @@ private void onChannelMaskSpinnerSelected() {
...
+
+    public void setupEffects(int sessionId) {
+        if (misOutput) {
+            mBassBoost = new BassBoost(0, sessionId);
+            mBassBoost.setStrength((short) mBassBoostSeekBar.getProgress());
+            mLoudnessEnhancer = new LoudnessEnhancer(sessionId);
+            mLoudnessEnhancer.setTargetGain((short) mLoudnessEnhancerSeekBar.getProgress());
+        } else {
+            mAcousticEchoCanceler = AcousticEchoCanceler.create(sessionId);
+            if (mAcousticEchoCanceler != null) {
```
philburk, line 513: https://developer.android.com/reference/android/media/audiofx/AcousticEchoCanceler#isAvailable()
might be useful here.

philburk (commented):
Nice! Thanks for doing this.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/EchoActivity.java
robertwu1: We are moving settingEffects from the job of an individual test into each stream config. This way output tests can get output effects, input tests can get input effects, and tests that test both can do both.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/StreamConfigurationView.java
robertwu1: Thanks! Added in the latest iteration

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/StreamConfigurationView.java
```diff
@@ -509,17 +512,21 @@ public void setupEffects(int sessionId) {
             mLoudnessEnhancer = new LoudnessEnhancer(sessionId);
             mLoudnessEnhancer.setTargetGain((short) mLoudnessEnhancerSeekBar.getProgress());
         } else {
-            mAcousticEchoCanceler = AcousticEchoCanceler.create(sessionId);
-            if (mAcousticEchoCanceler != null) {
-                mAcousticEchoCanceler.setEnabled(mAcousticEchoCancelerCheckBox.isEnabled());
-            } else {
-                Log.e(TAG, String.format("Could not create AcousticEchoCanceler"));
+            if (mAcousticEchoCanceler.getEnabled()) {
```
philburk, line 515: This looks weird. We query an object to see if it is enabled.
Then we overwrite it with a new object!

philburk: Maybe we need to keep a boolean outside the effect object that we can use before we create the object with the sessionID.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/StreamConfigurationView.java
```diff
@@ -509,17 +512,21 @@ public void setupEffects(int sessionId) {
...
+                } else {
+                    Log.e(TAG, String.format("Could not create AcousticEchoCanceler"));
+                }
             }
-            mAutomaticGainControl = AutomaticGainControl.create(sessionId);
-            if (mAutomaticGainControl != null) {
-                mAutomaticGainControl.setEnabled(mAutomaticGainControlCheckBox.isEnabled());
-            } else {
-                Log.e(TAG, String.format("Could not create AutomaticGainControl"));
+            if (mAutomaticGainControl.getEnabled()) {
```
philburk, line 523: ditto

philburk (changes requested):

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/StreamConfigurationView.java
robertwu1: This is set in lines 230 and 231. I agree this does looks a bit funky. I should add some comments

~~~
mAutomaticGainControlCheckBox.setEnabled(AutomaticGainControl.isAvailable());
     mAcousticEchoCancelerCheckBox.setEnabled(AcousticEchoCanceler.isAvailable());
~~~

philburk: Still funny.

philburk (changes requested):

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/StreamConfigurationView.java
robertwu1: Fixed in latest iteration

philburk (approved):

Issue #1603, opened 2022-08-09
The client receives the callback onErrorAfterClose from oboe even if the audio stream is closed.  [ Same as Issue 1578]


FeiWang888:
It is the same as Issue 1578, https://github.com/google/oboe/issues/1578, which is closed, but I don't get the right answer.
To draw attention, I am not able to reopen the previous one so open a new one. Please reopen 1578 or give a comment here. Thanks.

Android version(s):
Android 8.1+
Android device(s):
Samsung Galaxy Note10+, Samsung Galaxy S21 Ultra 5G, Google Pixel 5, Samsung Galaxy A70, Samsung Galaxy A72, etc..

Oboe version:
OboeVersion1.5.0

App name used for testing:
Cisco WebexMeeting App
https://play.google.com/store/apps/details?id=com.cisco.webex.meetings&hl=en_US&gl=US

Short description
We saw a lot of crash issues reported from the Google Console and receives some client logs for that.
However, it is very hard to reproduce the issue locally
Please give advice about the root causes and how to prevent them.

The key question is why the client receives the callback onErrorAfterClose from oboe even if the audio stream is closed.
Client log:
// stop the stream (thread id 2079)
NATIVE_WME 0:0 07/18/2022 01:46:59 737 I [TID:2079][NATIVE_TID:18036][AudioEngine] [CallID=33558529]WbxAndroidAudioPlaybackNative::StopStream end, stream = 0xb400007afccefd80, stream->getState() = 10,this=0xb400007a27418018
// then close the stream (thread id 2079)
NATIVE_WME 0:0 07/18/2022 01:46:59 778 I [TID:2079][NATIVE_TID:18036][AudioEngine] [CallID=33558529]WbxAndroidAudioPlaybackNative::CloseStream end, stream = 0xb400007afccefd80, stream->getState() = 12,this=0xb400007a27418018
// deconstruct the object (thread id 2079)
NATIVE_WME 0:0 07/18/2022 01:46:59 780 I [TID:2079][NATIVE_TID:18036][AudioEngine] [CallID=33558529]WbxAndroidAudioPlaybackNative::~WbxAndroidAudioPlaybackNative,this=0xb400007a27418018
// receives the callback onErrorAfterClose from oboe (thread id 22077)
[...]

PR #1614, opened 2022-08-30
OboeTester: add error callback test

philburk:
Delete an error callback while it is being used by Oboe.

Try to reproduce issue #1603

### DIFF
--- a/apps/OboeTester/app/src/main/cpp/TestErrorCallback.h
+++ b/apps/OboeTester/app/src/main/cpp/TestErrorCallback.h
@@ -1,0 +1,107 @@
+/*
+ * Copyright 2022 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef OBOETESTER_TEST_ERROR_CALLBACK_H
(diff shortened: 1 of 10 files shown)


apps/OboeTester/app/src/main/cpp/TestErrorCallback.h
```diff
@@ -0,0 +1,106 @@
...
+            usleep(10 * 1000); // Force us to lose the race condition.
+            // The callback was probably deleted by now.
+            LOGE("%s() - error = %s, mMagic = 0x%08X",
+                 __func__, oboe::convertToText(error), mMagic.load());
+            mParent->mCallbackMagic = mMagic.load();
+        }
+
+    private:
+        TestErrorCallback *mParent;
+        static constexpr int32_t kMagicGood = 0x2B4ACAFE;
```
robertwu1, line 95: Can we use something like 0x600DCAFE? It's hard to determine that 0x2B4ACAFE is the intended good value

apps/OboeTester/app/src/main/cpp/TestErrorCallback.h
```diff
@@ -0,0 +1,106 @@
...
+            // reported in #1603.
+            std::thread t([this]() {
+                    this->mParent->cleanup();
+                    LOGE("onErrorBeforeClose called cleanup!");
+                });
+            t.detach();
+        }
+
+        void onErrorAfterClose(oboe::AudioStream *oboeStream, oboe::Result error) override {
+            usleep(10 * 1000); // Force us to lose the race condition.
```
robertwu1, line 86: I'm confused why we need the sleep here. What is the order of events for the bad and good cases?

apps/OboeTester/app/src/main/cpp/TestErrorCallback.h
philburk: Nice! will do.

apps/OboeTester/app/src/main/cpp/TestErrorCallback.h
philburk: We want to try to delete the callback before onErrorAfterClose() is run.
It is a race condition and I want the thread doing the deletion to win the race.
After we get it to FAIL reliably then I will fix it by adding an API that uses a shared_ptr.

PR #1620, opened 2022-09-07
OboeTester: bump gradle to 7.2.2

philburk:


apps/OboeTester/app/build.gradle
```diff
@@ -30,6 +30,7 @@ android {
             path "CMakeLists.txt"
         }
     }
+    ndkVersion '21.4.7075529'
```
robertwu1, line 33: Do we need this? We don't have this in other parts of our codebase in Oboe.

philburk: It builds without this. I will remove it.

robertwu1 (approved):

PR #1626, opened 2022-09-15
oboe: use shared_ptr for callbacks

### LINKED ISSUE #1603: The client receives the callback onErrorAfterClose from oboe even if the audio stream is closed.  [ Same as Issue 1578]

philburk:
Pass data and error callbacks to builder using a shared_ptr. Deprecate methods that pass raw pointers.

This is to avoid the callback getting deleted while the error callback is active.

Fixes #1603

### DIFF
--- a/apps/OboeTester/app/src/main/cpp/TestErrorCallback.cpp
+++ b/apps/OboeTester/app/src/main/cpp/TestErrorCallback.cpp
@@ -24,14 +24,19 @@ using namespace oboe;
 oboe::Result TestErrorCallback::open() {
     mCallbackMagic = 0;
     mDataCallback = std::make_unique<MyDataCallback>();
-    mErrorCallback = std::make_unique<MyErrorCallback>(this);
+    mErrorCallback = std::make_shared<MyErrorCallback>(this);
     AudioStreamBuilder builder;
     oboe::Result result = builder.setSharingMode(oboe::SharingMode::Exclusive)
             ->setPerformanceMode(oboe::PerformanceMode::LowLatency)
             ->setFormat(oboe::AudioFormat::Float)
             ->setChannelCount(kChannelCount)
+#if 0
             ->setDataCallback(mDataCallback.get())
-            ->setErrorCallback(mErrorCallback.get())
+            ->setErrorCallback(mErrorCallback.get()) // This can lead to a crash.
+#else
+            ->setDataCallback(mDataCallback)
+            ->setErrorCallback(mErrorCallback) // shared_ptr avoids a crash
+#endif
             ->openStream(mStream);
     return result;
 }
--- a/apps/OboeTester/app/src/main/cpp/TestErrorCallback.cpp
+++ b/apps/OboeTester/app/src/main/cpp/TestErrorCallback.cpp
@@ -23,7 +23,7 @@ using namespace oboe;
 
 oboe::Result TestErrorCallback::open() {
     mCallbackMagic = 0;
-    mDataCallback = std::make_unique<MyDataCallback>();
+    mDataCallback = std::make_shared<MyDataCallback>();
     mErrorCallback = std::make_shared<MyErrorCallback>(this);
     AudioStreamBuilder builder;
     oboe::Result result = builder.setSharingMode(oboe::SharingMode::Exclusive)
@@ -32,7 +32,7 @@ oboe::Result TestErrorCallback::open() {
             ->setChannelCount(kChannelCount)
 #if 0
             ->setDataCallback(mDataCallback.get())
-            ->setErrorCallback(mErrorCallback.get()) // This can lead to a crash.
+            ->setErrorCallback(mErrorCallback.get()) // This can lead to a crash or FAIL.
 #else
             ->setDataCallback(mDataCallback)
             ->setErrorCallback(mErrorCallback) // shared_ptr avoids a crash
--- a/include/oboe/AudioStreamBuilder.h
+++ b/include/oboe/AudioStreamBuilder.h
@@ -354,11 +354,31 @@ public:
      * <strong>Important: See AudioStreamCallback for restrictions on what may be called
      * from the callback methods.</strong>
      *
+     * We pass a shared_ptr so that the sharedDataCallback object cannot be deleted
+     * before the stream is deleted.
+     *
      * @param dataCallback
      * @return pointer to the builder so calls can be chained
      */
-    AudioStreamBuilder *setDataCallback(oboe::AudioStreamDataCallback *dataCallback) {
+    AudioStreamBuilder *setDataCallback(std::shared_ptr<AudioStreamDataCallback> sharedDataCallback) {
+        // Use this raw pointer in the rest of the code to retain backwards compatibility.
+        mDataCallback = sharedDataCallback.get();
+        // Hold a shared_ptr to protect the raw pointer for the lifetime of the stream.
+        mSharedDataCallback = sharedDataCallback;
+        return this;
+    }
+
+    /**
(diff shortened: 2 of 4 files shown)


apps/OboeTester/app/src/main/cpp/TestErrorCallback.cpp
```diff
@@ -24,14 +24,19 @@ using namespace oboe;
 oboe::Result TestErrorCallback::open() {
     mCallbackMagic = 0;
     mDataCallback = std::make_unique<MyDataCallback>();
```
philburk, line 26: Should be make_shared.

include/oboe/AudioStreamBuilder.h
```diff
@@ -354,11 +354,31 @@ class AudioStreamBuilder : public AudioStreamBase {
...
+
+    /**
+    * Pass a raw pointer to a data callback. This is not recommended because the dataCallback
+    * object might get deleted by the app while it is being used.
+    *
+    * @deprecated Call setDataCallback(std::shared_ptr<AudioStreamDataCallback>) instead.
+    * @param dataCallback
+    * @return pointer to the builder so calls can be chained
+    */
+    AudioStreamBuilder *setDataCallback(AudioStreamDataCallback *dataCallback) {
```
robertwu1, line 379: Should we deprecate this? There may be performance hits with shared_ptr on the dataCallback.

Can we measure timing somehow with some test to verify there is no regression?

include/oboe/AudioStreamBuilder.h
```diff
@@ -354,11 +354,31 @@ class AudioStreamBuilder : public AudioStreamBase {
...
+     * We pass a shared_ptr so that the sharedDataCallback object cannot be deleted
+     * before the stream is deleted.
+     *
      * @param dataCallback
      * @return pointer to the builder so calls can be chained
      */
-    AudioStreamBuilder *setDataCallback(oboe::AudioStreamDataCallback *dataCallback) {
+    AudioStreamBuilder *setDataCallback(std::shared_ptr<AudioStreamDataCallback> sharedDataCallback) {
+        // Use this raw pointer in the rest of the code to retain backwards compatibility.
+        mDataCallback = sharedDataCallback.get();
```
robertwu1, line 365: Are we sure that this works as intended? There is nothing stopping the raw pointer from getting deleted somewhere else in the code.

include/oboe/AudioStreamBuilder.h
philburk: We use the stored raw pointer for all the real-time callbacks. So there should be no performance impact.

include/oboe/AudioStreamBuilder.h
philburk: We control Oboe and can promise not to delete the raw pointer. Maybe we should add a comment to that effect. We did not create the callback and should not directly delete it. That would be a bad bug.

The callback object is created by the app and shared with Oboe. Oboe AudioStream will hold a shared pointer so we know the callback will not get deleted by the app while the AudioStream still exists. And if the AudioStream is created with a shared_ptr then we know it will exist for the duration of the error callback.

It is always possible to delete some raw pointer and mess things up if we try. The goal is to make it possible for apps to  be safe if they try. Before this change, it was hard for apps to be safe even when they tried.

robertwu1 (approved):

Issue #1629, opened 2022-09-21
Samples do not handle disconnect between open and start


philburk:
Android version(s): master
Android device(s): Pixel 6
Oboe version: main

Steps to reproduce

I hacked AudioStreamAAudio::requestStart() so that it returns ErrorDisconnected the first time it is called.

    static bool sFirstTime = true;
    if (sFirstTime) {
        sFirstTime = false;
        return Result::ErrorDisconnected;
    }

This simulates a race condition where a device is unplugged between the Open and the Start.

Expected behavior
App should close the old stream and reopen a new stream if the start call fails.
App should then make sound.

Actual behavior

No sound from the app.

Apps affected include:
* HelloOboe
* SoundBoard

philburk:
@robertwu1 - need fix in SoundBoard too.

PR #1630, opened 2022-09-21
HelloOboe: try again when start fails

philburk:
This handles the case where a DISCONNECT occurs between Open and Start.

See #1629

### DIFF
--- a/samples/hello-oboe/src/main/cpp/HelloOboeEngine.cpp
+++ b/samples/hello-oboe/src/main/cpp/HelloOboeEngine.cpp
@@ -132,30 +132,39 @@ void HelloOboeEngine::restart() {
 
 oboe::Result HelloOboeEngine::start() {
     std::lock_guard<std::mutex> lock(mLock);
-    mIsLatencyDetectionSupported = false;
-    auto result = openPlaybackStream();
-    if (result == oboe::Result::OK){
-        mAudioSource =  std::make_shared<SoundGenerator>(mStream->getSampleRate(),
-                mStream->getChannelCount());
-        mLatencyCallback->setSource(std::dynamic_pointer_cast<IRenderableAudio>(mAudioSource));
+    oboe::Result result = oboe::Result::OK;
+    // It is possible for a stream's device to become disconnected between the Open and the Start.
+    // So if it fails to start, close the old stream and try again.
+    int tryCount = 0;
+    bool startFailed = false;
+    do {
+        mIsLatencyDetectionSupported = false;
+        result = openPlaybackStream();
+        if (result == oboe::Result::OK){
+            mAudioSource =  std::make_shared<SoundGenerator>(mStream->getSampleRate(),
+                    mStream->getChannelCount());
+            mLatencyCallback->setSource(std::dynamic_pointer_cast<IRenderableAudio>(mAudioSource));
 
-        LOGD("Stream opened: AudioAPI = %d, channelCount = %d, deviceID = %d",
-                 mStream->getAudioApi(),
-                 mStream->getChannelCount(),
-                 mStream->getDeviceId());
+            LOGD("Stream opened: AudioAPI = %d, channelCount = %d, deviceID = %d",
+                     mStream->getAudioApi(),
+                     mStream->getChannelCount(),
+                     mStream->getDeviceId());
 
-        result = mStream->start();
-        if (result != oboe::Result::OK) {
-            LOGE("Error starting playback stream. Error: %s", oboe::convertToText(result));
-            mStream->close();
-            mStream.reset();
+            result = mStream->requestStart();
+            if (result != oboe::Result::OK) {
+                LOGE("Error starting playback stream. Error: %s", oboe::convertToText(result));
+                mStream->close();
+                mStream.reset();
+                startFailed = true;
+            } else {
+                mIsLatencyDetectionSupported = (mStream->getTimestamp((CLOCK_MONOTONIC)) !=
+                                                oboe::Result::ErrorUnimplemented);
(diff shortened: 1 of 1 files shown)


samples/hello-oboe/src/main/cpp/HelloOboeEngine.cpp
```diff
@@ -132,30 +132,39 @@ void HelloOboeEngine::restart() {
 
 oboe::Result HelloOboeEngine::start() {
     std::lock_guard<std::mutex> lock(mLock);
```
robertwu1, line 134: Is it ok to hold the lock the whole time with the new set of operations? This is probably fine, but I want just to check with you if you have any concerns

samples/hello-oboe/src/main/cpp/HelloOboeEngine.cpp
```diff
@@ -132,30 +132,39 @@ void HelloOboeEngine::restart() {
...
+                startFailed = true;
+            } else {
+                mIsLatencyDetectionSupported = (mStream->getTimestamp((CLOCK_MONOTONIC)) !=
+                                                oboe::Result::ErrorUnimplemented);
+                startFailed = false;
+            }
         } else {
-            mIsLatencyDetectionSupported = (mStream->getTimestamp((CLOCK_MONOTONIC)) !=
-                                            oboe::Result::ErrorUnimplemented);
+            LOGE("Error creating playback stream. Error: %s", oboe::convertToText(result));
```
robertwu1, line 165: Should we add a startFailed = true here?

samples/hello-oboe/src/main/cpp/HelloOboeEngine.cpp
```diff
@@ -132,30 +132,39 @@ void HelloOboeEngine::restart() {
...
+            result = mStream->requestStart();
+            if (result != oboe::Result::OK) {
+                LOGE("Error starting playback stream. Error: %s", oboe::convertToText(result));
+                mStream->close();
+                mStream.reset();
+                startFailed = true;
+            } else {
+                mIsLatencyDetectionSupported = (mStream->getTimestamp((CLOCK_MONOTONIC)) !=
+                                                oboe::Result::ErrorUnimplemented);
+                startFailed = false;
```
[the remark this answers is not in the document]
philburk: We need to set it false so we can succeed after a failure.

samples/hello-oboe/src/main/cpp/HelloOboeEngine.cpp
philburk: It is a long time. But safe and probably OK. I am going to move all this into an AsyncTask later so it won't hang the UI thread.

samples/hello-oboe/src/main/cpp/HelloOboeEngine.cpp
```diff
@@ -132,30 +132,39 @@ void HelloOboeEngine::restart() {
...
+            }
         } else {
-            mIsLatencyDetectionSupported = (mStream->getTimestamp((CLOCK_MONOTONIC)) !=
-                                            oboe::Result::ErrorUnimplemented);
+            LOGE("Error creating playback stream. Error: %s", oboe::convertToText(result));
         }
-    } else {
-        LOGE("Error creating playback stream. Error: %s", oboe::convertToText(result));
-    }
+    } while (tryCount++ < 3 && startFailed);
```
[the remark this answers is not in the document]
philburk: I will add a sleep when I do the AsyncTask.

samples/hello-oboe/src/main/cpp/HelloOboeEngine.cpp
philburk: I will check to make sure a DISCONNECT cannot occur during open.

philburk: If a DISCONNECT occurs during open then it was possible for the stream to open and then the start would probably fail. AAudio was fixed so that the open could fail due to a disconnect.
See [b/193071260](http://b/193071260) | P2 | AAudio stream should not open OK if disconnected while opening

 I will retry if either the open or the start fails.

samples/hello-oboe/src/main/cpp/HelloOboeEngine.cpp
```diff
@@ -132,30 +132,40 @@ void HelloOboeEngine::restart() {
...
-                 mStream->getChannelCount(),
-                 mStream->getDeviceId());
-
-        result = mStream->start();
-        if (result != oboe::Result::OK) {
-            LOGE("Error starting playback stream. Error: %s", oboe::convertToText(result));
-            mStream->close();
-            mStream.reset();
+    oboe::Result result = oboe::Result::OK;
+    // It is possible for a stream's device to become disconnected during thew open or between
```
robertwu1, line 136: typo in thew

samples/hello-oboe/src/main/cpp/HelloOboeEngine.cpp
```diff
@@ -132,30 +132,40 @@ void HelloOboeEngine::restart() {
...
+    // the Open and the Start.
+    // So if it fails to start, close the old stream and try again.
+    int tryCount = 0;
+    do {
+        if (tryCount > 0) {
+            usleep(20 * 1000); // Sleep between tries to give the system time to settle.
+        }
+        mIsLatencyDetectionSupported = false;
+        result = openPlaybackStream();
```
philburk, line 145: A DISCONNECT can also occur during the open so retry if open fails.

robertwu1 (approved):

PR #1634, opened 2022-09-27
OpenSL ES: Return ErrorInvalidFormat instead of ErrorInternal for invalid format

robertwu1:
OpenSL ES only supports float and I16. We should return that other formats are invalid if OpenSL ES is used. Otherwise, it may be hard for developers to debug.

### DIFF
--- a/src/opensles/AudioOutputStreamOpenSLES.cpp
+++ b/src/opensles/AudioOutputStreamOpenSLES.cpp
@@ -130,9 +130,9 @@ Result AudioOutputStreamOpenSLES::open() {
     }
 
     // OpenSL ES only supports I16 and Float
-    //if (mFormat != AudioFormat::I16 && mFormat != AudioFormat::Float) {
-    //    return Result::ErrorInvalidFormat;
-    //}
+    if (mFormat != AudioFormat::I16 && mFormat != AudioFormat::Float) {
+        return Result::ErrorInvalidFormat;
+    }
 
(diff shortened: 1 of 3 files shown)


src/opensles/AudioOutputStreamOpenSLES.cpp
```diff
@@ -129,6 +129,11 @@ Result AudioOutputStreamOpenSLES::open() {
                   AudioFormat::I16 : AudioFormat::Float;
     }
 
+    // OpenSL ES only supports I16 and Float
+    if (mFormat != AudioFormat::I16 && mFormat != AudioFormat::Float) {
+        return Result::ErrorInvalidFormat;
+    }
```
flamme, line 135: The format validation is the same for input and output. In that case, why not put this in AudioStreamOpenSLES::open?

philburk (changes requested):
Look for other format check in Oboe.

src/opensles/AudioOutputStreamOpenSLES.cpp
robertwu1: Change now in a different spot. At this new spot, I don't think it should be in AudioStreamOpenSLES::open

robertwu1:
> Look for other format check in Oboe.

Done. Thanks for the suggestion!

flamme (approved):

philburk (approved):

PR #1635, opened 2022-09-28
Add MinimalOboe Kotlin app

philburk:
Uses coroutine to call Oboe native to prevent ANR.


samples/minimaloboe/src/androidTest/java/com/example/minimaloboe/ExampleInstrumentedTest.kt
```diff
@@ -0,0 +1,24 @@
...
+import org.junit.runner.RunWith
+
+import org.junit.Assert.*
+
+/**
+ * Instrumented test, which will execute on an Android device.
+ *
+ * See [testing documentation](http://d.android.com/tools/testing).
+ */
+@RunWith(AndroidJUnit4::class)
```
robertwu1, line 16: Can we remove?

samples/minimaloboe/src/test/java/com/example/minimaloboe/ExampleUnitTest.kt
```diff
@@ -0,0 +1,17 @@
...
+import org.junit.Test
+
+import org.junit.Assert.*
+
+/**
+ * Example local unit test, which will execute on the development machine (host).
+ *
+ * See [testing documentation](http://d.android.com/tools/testing).
+ */
+class ExampleUnitTest {
```
robertwu1, line 12: Can we remove this test?

samples/minimaloboe/src/main/cpp/SimpleNoiseMaker.cpp
```diff
@@ -0,0 +1,92 @@
...
+DataCallbackResult SimpleNoiseMaker::MyDataCallback::onAudioReady(
+        AudioStream *audioStream,
+        void *audioData,
+        int32_t numFrames) {
+    // We requested float when we built the stream.
+    float *output = (float *) audioData;
+    // Fill buffer with random numbers to create "white noise".
+    int numSamples = numFrames * kChannelCount;
+    for (int i = 0; i < numSamples; i++) {
+        // drand48() returns a random number between 0.0 and 1.0.
```
robertwu1, line 74: Can we play a sine wave instead so it is pleasant sounding?

samples/minimaloboe/src/main/java/com/example/minimaloboe/MainActivity.kt
```diff
@@ -0,0 +1,110 @@
...
+        return audioPlayer.stopAudio()
+    }
+}
+
+@Composable
+fun MainControls(viewModel: ExampleViewModel = ExampleViewModel()) {
+
+    // State that affects the UI.
+    var started by remember { mutableStateOf(false) }
+    var result by remember { mutableStateOf(1) }
```
robertwu1, line 69: Having result start at 1 is quite strange. It doesn't map to Result::OK (0) or any of the error codes

samples/minimaloboe/src/main/java/com/example/minimaloboe/MainActivity.kt
```diff
@@ -0,0 +1,110 @@
...
+
+    Column {
+        Text(text = "Minimal Oboe!")
+        // START button
+        Button(
```
robertwu1, line 74: nit: the button UI takes some time to run. Sound starts playing way before the UI is done updating. Do you know whether we can change the speed of the UI transitions so it takes less time?

samples/minimaloboe/src/main/java/com/example/minimaloboe/MainActivity.kt
```diff
@@ -0,0 +1,110 @@
...
+class ExampleViewModel : ViewModel() {
+
+    var audioPlayer = AudioPlayer();
+
+    fun startAudio(): Int {
+        return audioPlayer.startAudio()
+
+    }
+
+    fun stopAudio(): Int {
```
robertwu1, line 59: Should we stop audio if the screen rotates?

samples/minimaloboe/src/main/java/com/example/minimaloboe/ui/theme/Type.kt
```diff
@@ -0,0 +1,28 @@
...
+import androidx.compose.ui.unit.sp
+
+// Set of Material typography styles to start with
+val Typography = Typography(
+    body1 = TextStyle(
+        fontFamily = FontFamily.Default,
+        fontWeight = FontWeight.Normal,
+        fontSize = 16.sp
+    )
+    /* Other default text styles to override
```
robertwu1, line 16: nit: remove comments?

samples/minimaloboe/src/main/cpp/MinimalOboeJNI.cpp
```diff
@@ -0,0 +1,64 @@
...
+}
+
+/**
+ * Native (JNI) implementation of AudioPlayer.stopAudioStreamNative()
+ */
+JNIEXPORT jint JNICALL Java_com_example_minimaloboe_AudioPlayer_stopAudioStreamNative(
+        JNIEnv * /* env */, jobject) {
+    __android_log_print(ANDROID_LOG_INFO, TAG, "%s", __func__);
+    Result result1 = sPlayer.stop();
+    Result result2 = sPlayer.close();
```
robertwu1, line 59: Should we call close if stop fails?

samples/minimaloboe/src/main/cpp/SimpleNoiseMaker.cpp
philburk: Some folks hate sine waves. And the sine wave requires state and a phaseIncrement. White Noise is more minimal.

samples/minimaloboe/src/main/java/com/example/minimaloboe/MainActivity.kt
philburk: My buttons are fast on Pixel 7. Which device is slow?

samples/minimaloboe/src/main/java/com/example/minimaloboe/MainActivity.kt
philburk: I like that the sound continues. But the UI gets out of sync. Will fix.

samples/minimaloboe/src/main/java/com/example/minimaloboe/MainActivity.kt
robertwu1: The UI takes about half a second for the full animation from purple -> gray -> black.

robertwu1: This for Pixel 6.

samples/minimaloboe/src/main/java/com/example/minimaloboe/MainActivity.kt
philburk: I also found the sound continues when the app is in the background. Not intended. So I added onPause(). It is better but when I switch tasks the UI can still get out of sync.

samples/minimaloboe/src/main/cpp/MinimalOboeJNI.cpp
philburk: Yes. The close() is needed to delete the resources.
If open() fails we do not start().
but if stop() fails we still want to close().
I added a comment.

robertwu1:
Btw, I thought about Java samples a bit more and Soundboard and Megadrone may be too simple as examples. We simply start the audio engine when the app is resumed and stop the audio engine when the app is paused. This is good for many apps, but not all apps want to have an open audio stream all the time.

For Soundboard and Megadrone, I don't think we should change anything. Letting the UI thread open audio is fine in these cases because having pads that do not play audio at the beginning feels like a bug.

LiveEffect uses a different mechanism. Only when LiveEffectEngine.setEffectOn(bool) is called, should the engine be started or stopped. This seems like a good place to spin off another non-UI thread.

HelloOboe is more complicated than all of samples listed above. It creates an audio stream onResume() and deletes it onPause(). It also closes the current stream and reopens a new stream whenever PlaybackEngine.setChannelCount, PlaybackEngine.setAudioDeviceId, or PlaybackEngine.setAudioApi is called. This is clearly the largest sample despite the name.

I recommend making changes to LiveEffect for a Java sample. For the wiki, we can list out 4 samples in the following chart:

| Type of sample             | Kotlin      | Java       |
|----------------------------|-------------|------------|
| Create engine at on resume | DrumThumper | Soundboard |
| Dynamically create engine  | MinimalOboe | LiveEffect |

philburk:
I agree LiveEffect would be a good candidate for a background task.

Re modifying SoundBoard
>having pads that do not play audio at the beginning feels like a bug.

Ideally the open should just take 100-200 msec. The user should not notice the unresponsive pads. But if the audioserver dies and takes 10 seconds to recover then the ANR in the UI thread will kill the app. That is not acceptable. So all apps should really use a background task to avoid ANRs.

philburk:
@mmoczkowski - Do you know Kotlin/Compose? If so can you review to see if this demonstrates proper techniques? I am new to both.

robertwu1:
> @robertwu1 wrote:
> 
> > This file also needs to a line about DrumThumper
> 
> I don't understand. Can you clarify?

It currently lists out 5 samples. It doesn't list out DrumThumper

samples/minimaloboe/src/main/java/com/example/minimaloboe/MainActivity.kt
```diff
@@ -0,0 +1,119 @@
...
+import androidx.compose.runtime.*
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.lifecycle.ViewModel
+import com.example.minimaloboe.ui.theme.SamplesTheme
+import kotlinx.coroutines.*
+
+class MainActivity : ComponentActivity() {
+
+    var mExampleViewModel = ExampleViewModel()
```
mmoczkowski, line 34: ViewModels should only be instantiated by `ViewModelProvider` ([see here](https://developer.android.com/topic/libraries/architecture/viewmodel#viewmodel-with-dependencies)).  Easiest way to do so:

1. Declare `androidx.activity:activity-ktx` dependency in Gradle
2. Use delegated property to instantiate VM: `val exampleViewModel: ExampleViewModel by viewModels()`

samples/minimaloboe/src/main/java/com/example/minimaloboe/MainActivity.kt
```diff
@@ -0,0 +1,119 @@
...
+                    modifier = Modifier.fillMaxSize(),
+                    color = MaterialTheme.colors.background
+                ) {
+                    MainControls(mExampleViewModel)
+                }
+            }
+        }
+    }
+
+    override fun onStop() {
```
mmoczkowski, line 51: Overriding lifecycle methods is now discouraged in favor of using `LifecycleObserver`. I'd suggest:

1. Make `AudioPlayer` extend the `DefaultLifecycleObserver`
2. Register lifecycle observer in Activity's `onCreate`: `lifecycle.addObserver(player)`

samples/minimaloboe/src/main/java/com/example/minimaloboe/MainActivity.kt
```diff
@@ -0,0 +1,119 @@
...
+        return audioPlayer.startAudio()
+    }
+
+    fun stopAudio(): Int {
+        return audioPlayer.stopAudio()
+    }
+}
+
+@Composable
+fun MainControls(viewModel: ExampleViewModel = ExampleViewModel()) {
```
mmoczkowski, line 72: 1. viewModel shouldn't be instantiated like that, but you can do something very similiar with hilt viewmodel integration `viewModel: ExampleViewModel = viewModel()`. See [here](https://developer.android.com/jetpack/compose/libraries#hilt)
2. Composable should provide a modifier argument. See [here](https://chris.banes.dev/always-provide-a-modifier/)

samples/minimaloboe/src/main/java/com/example/minimaloboe/MainActivity.kt
```diff
@@ -0,0 +1,119 @@
...
+
+    Column {
+        Text(text = "Minimal Oboe!")
+        // START button
+        Button(
+            onClick = {
+                // Start Oboe from a coroutine in case it blocks for too long.
+                // If the AudioServer has died it may take several seconds to recover.
+                // That can cause an ANR if we are starting audio from the main UI thread.
+                GlobalScope.launch {
```
mmoczkowski, line 88: ViewModel should expose non-suspending functions and launch coroutines itself. `GlobalScope` should be avoided, use `viewModelScope` instead.

samples/minimaloboe/src/main/java/com/example/minimaloboe/MainActivity.kt
philburk: Thanks Milosz. Don refactored the app and it now uses LifecycleObserver.

samples/minimaloboe/src/main/java/com/example/minimaloboe/MainActivity.kt
philburk: I tried on Oriole running TM and did not see slow animation.

samples/minimaloboe/src/main/java/com/example/minimaloboe/MainActivity.kt
philburk: Don fixed this.

samples/minimaloboe/src/main/java/com/example/minimaloboe/AudioPlayer.kt
```diff
@@ -0,0 +1,77 @@
...
+    init {
+        // Load the library containing the native code including the JNI functions.
+        System.loadLibrary("minimaloboe")
+    }
+
+    fun setPlaybackEnabled(isEnabled: Boolean) {
+        // Start (and stop) Oboe from a coroutine in case it blocks for too long.
+        // If the AudioServer has died it may take several seconds to recover.
+        // That can cause an ANR if we are starting audio from the main UI thread.
+        GlobalScope.launch {
```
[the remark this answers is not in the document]
dturner: I'm going to answer my own question here and say that using `GlobalScope` is nearly always a bad idea. Despite our `AudioPlayer` being a singleton (`object`) which will live for the lifetime of the application, any jobs executed by `AudioPlayer` should be executed within a scope owned by `AudioPlayer`. For this reason I've just submitted a PR which adds a `CoroutineContext` to this class under which jobs can be executed and cancelled if `AudioPlayer` is destroyed (if, for example, someone changed it from `object` to `class`).

samples/minimaloboe/src/main/java/com/example/minimaloboe/MainActivity.kt
philburk: I will add a modifier argument.

samples/minimaloboe/src/main/java/com/example/minimaloboe/MainActivity.kt
philburk: Let's tweak this later if still a problem.

samples/minimaloboe/src/main/java/com/example/minimaloboe/MainActivity.kt
philburk: I think this obsolete with Don's changes.

samples/minimaloboe/src/main/java/com/example/minimaloboe/AudioPlayer.kt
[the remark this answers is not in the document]
philburk: Thanks Don. I merged that change.

philburk:
@donturner - Thanks for the coroutine scope patch. Requesting final review and Approval.

samples/minimaloboe/src/main/res/values/colors.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <color name="purple_200">#FFBB86FC</color>
+    <color name="purple_500">#FF6200EE</color>
+    <color name="purple_700">#FF3700B3</color>
```
robertwu1, line 5: nit: Seems like the only color defined here that is used is purple_700. Is this ok?

samples/minimaloboe/src/main/res/mipmap-anydpi-v26/ic_launcher.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<adaptive-icon xmlns:android="http://schemas.android.com/apk/res/android">
+    <background android:drawable="@drawable/ic_launcher_background" />
+    <foreground android:drawable="@drawable/ic_launcher_foreground" />
+</adaptive-icon>
```
robertwu1, line 5: Add newlines to end of files

robertwu1 (approved):

mmoczkowski (approved):

samples/minimaloboe/src/main/res/values/colors.xml
philburk: This was default code provided by Android Studio.

samples/minimaloboe/src/main/res/mipmap-anydpi-v26/ic_launcher.xml
philburk: More default files provided by Android Studio. Should be fixed in Android Studio.

philburk:
This was merged manually to avoid merge conflicts.
I followed the instructions given by GitHub, which unfortunately did not involve a squash. So it was a messy merge.

PR #1636, opened 2022-10-04
Refactor minimal app to move UI state into AudioPlayer

dturner:
Hey @philburk :) 

I refactored your code sample to move the playback state into the `AudioPlayer` and made it a singleton. This way `AudioPlayer` becomes the source of truth for your UI. 

`AudioPlayer` implements `DefaultLifecycleObserver` which means it can listen for lifecycle events. When it receives an `onStop` event from `ProcessLifecycleOwner` it indicates that the app has gone into the background so we stop playback. 

Note that playback continues if you just rotate the device. I figured this was the desired behaviour.

LMK if you have any questions.


philburk:
@donturner - Thanks for sharing your expertise! This is exactly what I was hoping for.  I like Kotlin/Compose but the state management is a bit arcane.

philburk (approved):

PR #1652, opened 2022-11-11
OboeTester: Log whether an added device is a sink or source in TestPlugLatencyActivity

robertwu1:
I tested with a 3.5mm headset and this now shows one sink and one source


philburk (approved):

philburk:
Thanks for the quick response.

Issue #1657, opened 2022-11-22
Can't enable asan on OboeTester


xzhan96:
Android version(s): Android 12
Android device(s): Mi 9
Oboe version: master source code
App name used for testing: OboeTester
(Please try to reproduce the issue using the OboeTester or an Oboe sample.)
 
**Short description**
Follow the guide of https://developer.android.com/ndk/guides/asan#cmake to enable asan on OboeTester, but not works when setup a crash case.

**Steps to reproduce**
1. follow the step here https://developer.android.com/ndk/guides/asan#cmake, enable asan
2. create a case that must crash on ASAN version apk
![image](https://user-images.githubusercontent.com/13726257/203255071-084ed3b5-4153-4de5-add6-f87882e0edc6.png)
3. build the OboeTester
4. Run Oboe tester --> Test Output -->Open & Start
you will see the above log print, but the crash not happens.

philburk:
@xzhan96 - Thanks for reporting this. I think we should always be running with asan. So let's get this to work.

I am trying to get this to work.

Did you add:
    android:debuggable="true"
to the AndroidManifest.xml in the <application> tag?

Have you tried running under the debugger?

Interesting discussion at:
https://stackoverflow.com/questions/2952140/android-how-to-mark-my-app-as-debuggable

I find the instructions in the NDK to be a bit vague. I filed a bug for that.

I also uploaded my attempt at enabling ASAN in #1662 . How does it compare with your attempt?

philburk:
I  was able to get everything to build.
I even added your "test[3]=5" code.
It does not trigger ASAN!

philburk:
I asked the ASAN folks and they said that the optimizing compiler probably eliminated the buffer overflow. They recommended using a volatile to trick the compiler.
So I used this code to trigger the ASAN assertion!

````
volatile int* gSmallArray = nullptr;

JNIEXPORT jint JNICALL
Java_com_mobileer_oboetester_TestAudioActivity_startNative(JNIEnv *env, jobject) {
    // Test ASAN
    static int callCounter = 0;
    LOGE("%s FIXME - try to CRASH!!", __func__);
    if (gSmallArray == nullptr) {
        gSmallArray = new int[3]{0};
    }
    gSmallArray[3] = ++callCounter;
    LOGE("gSmallArray[3] = %d", *(gSmallArray + 3));
    return (jint) engine.getCurrentActivity()->start();
}
````

Updated code in an experimental PR at:
[...]

PR #1662, opened 2022-11-24
DO NOT MERGE - enable asan

philburk:
Try to enable hardware address sanitizer.
Do not merge this because it does not work
and it adds lots of overhead. This is just for testing.

See #1657


philburk:
Close now but keep as an example of how to turn on ASAN.

PR #1665, opened 2022-11-29
OboeTester: fix overflow in array of PeakDetectors.

philburk:
This caused crashes or other undefined behavior
when trying to record more than 8 channels in
the RECORD AND PLAY test.

For b/260169236

### DIFF
--- a/apps/OboeTester/app/src/main/cpp/InputStreamCallbackAnalyzer.h
+++ b/apps/OboeTester/app/src/main/cpp/InputStreamCallbackAnalyzer.h
@@ -29,14 +29,12 @@
 #include "MultiChannelRecording.h"
 #include "OboeTesterStreamCallback.h"
 
-constexpr int kMaxInputChannels = 8;
-
 class InputStreamCallbackAnalyzer : public OboeTesterStreamCallback {
 public:
 
     void reset() {
-        for (auto detector : mPeakDetectors) {
-            detector.reset();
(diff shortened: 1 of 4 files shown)


flamme (commented):
Do we have documentation about the maximum channel limitation?

apps/OboeTester/app/src/main/cpp/InputStreamCallbackAnalyzer.h
```diff
@@ -75,11 +75,11 @@ class InputStreamCallbackAnalyzer : public OboeTesterStreamCallback {
     }
 
 public:
-    PeakDetector            mPeakDetectors[kMaxInputChannels];
+    int32_t mNumChannels = 0;
+    std::unique_ptr<PeakDetector[]> mPeakDetectors;
```
robertwu1, line 79: There can be cases where getPeakLevel crashes as a certain channel is non-existent or the the unique_ptr is null.

This may be fine as this is a test app.

philburk: Yeah, it should only be called after setup().
But I should add error checks and return -1.0 on error.

philburk:
> Do we have documentation about the maximum channel limitation?

No. There is no fixed limit now that I fixed that code.
This code is internal to OboeTester. It is designed to possibly exceed the capabilities of AAudio.

robertwu1 (approved):

PR #1669, opened 2022-12-04
Add BackgroundRunner to HelloOboe

### LINKED ISSUE #960: Add example of calling AAudio from a background thread
If multiple threads are opening and closing stream then there can be race conditions that result in a crash. #953

The use of a single threaded audio "manager" can be used to avoid this.
The Java code can send start/stop messages to the thread when the app goes through its lifecycle, or in response to user input.
Error callbacks can occur when a device is disconnected. The callback can send a command to this thread.
The thread can have delays so that the app has time to respond to device changes from both Java and the error callback and then decide what to do.

Maybe we could put this in LiveEffect or Hello Oboe.

### LINKED ISSUE #1538: Use background thread when calling Oboe from examples
Calls to AudioFlinger can hang for 5-10 seconds if opening, starting, stopping, closing or doing other things that hit the audio service.

That can result in an ANR if called from the nUI thread.
We should move those calls to an AsyncTask or command thread.

We also need to document this.

philburk:
Call audio in background to prevent ANRs.
If AudioServer has crashed it can cause the UI thread to time out.

Fixes #960
Fixes #1538

### DIFF
--- a/samples/hello-oboe/src/main/java/com/google/oboe/samples/hellooboe/MainActivity.java
+++ b/samples/hello-oboe/src/main/java/com/google/oboe/samples/hellooboe/MainActivity.java
@@ -25,6 +25,8 @@ import android.os.Bundle;
 
 import androidx.annotation.RequiresApi;
 
+import android.os.Message;
+import android.util.Log;
 import android.view.MotionEvent;
 import android.view.View;
 import android.widget.AdapterView;
@@ -46,7 +48,7 @@ import java.util.Timer;
 import java.util.TimerTask;
 
 public class MainActivity extends Activity {
-    private static final String TAG = MainActivity.class.getName();
+    private static final String TAG = "HelloOboe";
     private static final long UPDATE_LATENCY_EVERY_MILLIS = 1000;
     private static final Integer[] CHANNEL_COUNT_OPTIONS = {1, 2, 3, 4, 5, 6, 7, 8};
     // Default to Stereo (OPTIONS is zero-based array so index 1 = 2 channels)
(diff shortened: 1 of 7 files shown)


samples/hello-oboe/src/main/java/com/google/oboe/samples/hellooboe/MainActivity.java
```diff
@@ -65,6 +67,64 @@ public class MainActivity extends Activity {
...
 
+    /** Commands for background thread. */
+    private static final int WHAT_START = 100;
+    private static final int WHAT_STOP = 101;
+    private static final int WHAT_SET_DEVICE_ID = 102;
+    private static final int WHAT_SET_AUDIO_API = 103;
+    private static final int WHAT_SET_CHANNEL_COUNT = 104;
+
+    private class MyBackgroundRunner extends BackgroundRunner {
+        int audioApi;
```
robertwu1, line 78: Should we explicitly define default values for these?

robertwu1 (approved):
Looks good to me. I left one minor nit about defining some variables.

samples/hello-oboe/src/main/java/com/google/oboe/samples/hellooboe/MainActivity.java
philburk: Java always initializes variables to zero so we typically do not init them. But I will document that the default zero values correspond to the UNSPECIFIED values for the three parameters.

philburk:
Regarding whether to quit() the Looper thread, Christopher Tate says  not to:
https://groups.google.com/g/android-platform/c/_jsUqXi6DKM/m/9I_8U_0wtk0J

> Speaking very generally, never quit() your looper threads.  That method exists mostly for historical and testing
> reasons.  In Real Life™, I recommend that you continue to reuse the same looper thread(s) for the life of the
> process rather than creating/quitting them.

I will make this a Singleton so we only create one.

PR #1670, opened 2022-12-05
OboeTester: catch IOExceptions

philburk:
Was crashing if 13 channels selected.


robertwu1:
What about InputTestActivity and TestOutputActivity?

robertwu1 (approved):

philburk:
> What about InputTestActivity and TestOutputActivity?

public void openAudio(View view)
in TestInputActivity already catches IOException.
I tested them and they are not crashing.

PR #1674, opened 2022-12-09
OboeTester: fix phase error calculation

### LINKED ISSUE #1672: OboeTester: phase difference not accounting for wrap.
I am not accounting for the circular wraparound of the phase!

Phase is in radians from -PI to +PI.
Imagine two angles, -3.14 and +3.13.
Those are actually very close. But: diff = 6.27
The actual phaseError should be only 2PI - 6.27 = 0.0131

We need a new method:

    double calculatePhaseError(double p1, double p2) {
        double diff = Math.abs(p1 - p2);
        if (diff > Math.PI) {
           diff = (Math.PI * 2) - diff;
        }
        return diff;
    }

philburk:
Tune the Data Paths test.

Fixes #1672

### DIFF
--- a/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestDataPathsActivity.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestDataPathsActivity.java
@@ -26,6 +26,7 @@ import android.os.Bundle;
 import android.os.Handler;
 import android.os.Looper;
 import android.support.annotation.NonNull;
+import android.util.Log;
 import android.widget.CheckBox;
 
 import com.mobileer.audio_device.AudioDeviceInfoConverter;
@@ -66,11 +67,7 @@ public class TestDataPathsActivity  extends BaseAutoGlitchActivity {
     private final static int TYPICAL_SAMPLE_RATE = 48000;
     private final static double FRAMES_PER_CYCLE = TYPICAL_SAMPLE_RATE / MAX_SINE_FREQUENCY;
     private final static double PHASE_PER_BIN = 2.0 * Math.PI / FRAMES_PER_CYCLE;
-    private final static double MAX_ALLOWED_JITTER = 0.5 * PHASE_PER_BIN;
-    // Start by failing then let good results drive us into a pass value.
-    private final static double INITIAL_JITTER = 2.0 * MAX_ALLOWED_JITTER;
-    // A coefficient of 0.0 is no filtering. 0.9999 is extreme low pass.
-    private final static double JITTER_FILTER_COEFFICIENT = 0.8;
+    private final static double MAX_ALLOWED_JITTER = 2.0 * PHASE_PER_BIN;
     private final static String MAGNITUDE_FORMAT = "%7.5f";
 
     final int TYPE_BUILTIN_SPEAKER_SAFE = 0x18; // API 30
@@ -79,7 +76,8 @@ public class TestDataPathsActivity  extends BaseAutoGlitchActivity {
     private double mMaxMagnitude;
     private int    mPhaseCount;
     private double mPhase;
-    private double mPhaseJitter;
+    private double mPhaseErrorSum;
+    private double mPhaseErrorCount;
 
     AudioManager   mAudioManager;
     private CheckBox mCheckBoxInputPresets;
@@ -112,6 +110,14 @@ public class TestDataPathsActivity  extends BaseAutoGlitchActivity {
         }
     }
 
+    public static double calculatePhaseError(double p1, double p2) {
+        double diff = Math.abs(p1 - p2);
+        if (diff > Math.PI) {
+            diff = (Math.PI * 2) - diff;
+        }
+        return diff;
+    }
+
     // Periodically query for magnitude and phase from the native detector.
     protected class DataPathSniffer extends NativeSniffer {
 
@@ -125,7 +131,8 @@ public class TestDataPathsActivity  extends BaseAutoGlitchActivity {
             mMaxMagnitude = -1.0;
             mPhaseCount = 0;
             mPhase = 0.0;
-            mPhaseJitter = INITIAL_JITTER;
+            mPhaseErrorSum = 0.0;
+            mPhaseErrorCount = 0;
             super.startSniffer();
         }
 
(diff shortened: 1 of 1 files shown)


apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestDataPathsActivity.java
```diff
@@ -112,6 +110,14 @@ public static String comparePassedField(String prefix, Object failed, Object pas
         }
     }
 
+    public static double calculatePhaseError(double p1, double p2) {
```
robertwu1, line 113: The diff would be negative if the diff is over 2PI. Should we do some math wrapping around if it exceeds 2PI?

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestDataPathsActivity.java
```diff
@@ -267,16 +280,25 @@ public String didTestFail() {
...
         }
         return why;
     }
 
+    private double getAveragePhaseError() {
+        return (mPhaseErrorCount > 0) ? (mPhaseErrorSum / mPhaseErrorCount) : MAX_ALLOWED_JITTER;
+    }
+
+    private boolean isPhaseJitterValid() {
+        return mPhaseErrorCount > 4;
```
robertwu1, line 299: Where does this 4 come from? Can we define this as a constant?

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestDataPathsActivity.java
```diff
@@ -267,16 +280,25 @@ public String didTestFail() {
...
+        if (!isPhaseJitterValid()) {
+            why += ", jitterUnknown";
+        } else if (getAveragePhaseError() > MAX_ALLOWED_JITTER) {
+            why += ", jitterHigh";
         }
         return why;
     }
 
+    private double getAveragePhaseError() {
+        return (mPhaseErrorCount > 0) ? (mPhaseErrorSum / mPhaseErrorCount) : MAX_ALLOWED_JITTER;
```
robertwu1, line 295: Returning MAX_ALLOWED_JITTER if there is no error feels strange

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestDataPathsActivity.java
```diff
@@ -125,22 +131,27 @@ public void startSniffer() {
...
         @Override
         public void run() {
             mMagnitude = getMagnitude();
             mMaxMagnitude = getMaxMagnitude();
+            Log.d(TAG, String.format("magnitude = %7.4f, maxMagnitude = %7.4f",
+                    mMagnitude, mMaxMagnitude));
             // Only look at the phase if we have a signal.
             if (mMagnitude >= MIN_REQUIRED_MAGNITUDE) {
                 double phase = getPhase();
                 if (mPhaseCount > 3) {
```
robertwu1, line 148: Where does this 3 come from?

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestDataPathsActivity.java
```diff
@@ -267,16 +280,25 @@ public String didTestFail() {
...
-        boolean passed = true;
         if (mMaxMagnitude <= MIN_REQUIRED_MAGNITUDE) {
             why += ", mag";
         }
-        if (mPhaseJitter > MAX_ALLOWED_JITTER) {
-            why += ", jitter";
+        if (!isPhaseJitterValid()) {
+            why += ", jitterUnknown";
+        } else if (getAveragePhaseError() > MAX_ALLOWED_JITTER) {
+            why += ", jitterHigh";
```
robertwu1, line 289: Do we want to print the average phase error here?

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestDataPathsActivity.java
```diff
@@ -267,16 +280,25 @@ public String didTestFail() {
...
         StreamConfiguration actualOutConfig = mAudioOutTester.actualConfiguration;
-        boolean passed = true;
         if (mMaxMagnitude <= MIN_REQUIRED_MAGNITUDE) {
             why += ", mag";
         }
-        if (mPhaseJitter > MAX_ALLOWED_JITTER) {
-            why += ", jitter";
+        if (!isPhaseJitterValid()) {
+            why += ", jitterUnknown";
+        } else if (getAveragePhaseError() > MAX_ALLOWED_JITTER) {
```
robertwu1, line 288: This statement is using getAveragePhaseError() > MAX_ALLOWED_JITTER while is finished early is using (getAveragePhaseError() < MAX_ALLOWED_JITTER). One of these be should <= or >= so the statements are opposite of one another.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestDataPathsActivity.java
philburk: Thanks. The atan2() function will return angles in a reasonable range. But this function does not know where the angles came from and should be more robust.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestDataPathsActivity.java
philburk: Done

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestDataPathsActivity.java
philburk: It is not that there is "no error". We just have no error measurements. I will bump up the value and comment it better.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestDataPathsActivity.java
philburk: I added an explanation.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestDataPathsActivity.java
philburk: I already print that on the same line. This code is just to print the reason the test failed.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestDataPathsActivity.java
philburk: The statements are not opposite of each other. One asks whether we fail. The other code asks whether we can finish early because the results look so good.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestDataPathsActivity.java
```diff
@@ -79,7 +76,8 @@ public class TestDataPathsActivity  extends BaseAutoGlitchActivity {
     private double mMaxMagnitude;
     private int    mPhaseCount;
     private double mPhase;
-    private double mPhaseJitter;
+    private double mPhaseErrorSum;
+    private double mPhaseErrorCount;
```
robertwu1, line 80: nit: mPhaseErrorCount is a confusing variable name as it makes you think this is the count of the number of phase errors. I can't think of a much better name though

robertwu1 (approved):

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestDataPathsActivity.java
philburk: I agree that Error has bad connotations.
And I also cannot think of a better name.

PR #1702, opened 2023-02-02
oboe: clip waitForAvailableFrames

### LINKED ISSUE #1701: OboeTester recording fails when channels >= 3 and Input Margin >= 3
Android version(s): all
Android device(s): all
Oboe version: 1.7.0 and earlier
App name used for testing:  OboeTester
 
**Short description**
Input Margin can be set larger than the capacity. Then recording is corrupted.

**Steps to reproduce**

Launch OboeTester.
Press RECORD AND PLAY
Select 3 channels.
Set Input Margin to 4.
RECORD
PLAY

**Expected behavior**
good recording

**Actual behavior**
no recording, frame counters barely advance

When setting channels >= 3 the stream switches from MMAP to Legacy.
Multi-channel legacy streams have larger bursts and smaller capacity.
On Pixel 7 Pro I had a burst size of 960 frames and a capacity of 2880. 
So the capacity is only 3 bursts. 
When the Margin is set to 3 or 4, OboeTester was waiting for the buffer to fill up to capacity, or even above capacity.
It can't fill up that much so the callback is timing out.
The solution is to clip the margin to the (bufferCapacity-burstSize) in Oboe.

This matches internal bug b/265719561

philburk:
Clip against BufferCapacity minus BurstSize.

Also bump Oboe to 1.7.1 and OboeTester to  2.3.6.

Fixes #1701

### DIFF
--- a/src/common/AudioStream.cpp
+++ b/src/common/AudioStream.cpp
@@ -166,6 +166,12 @@ ResultWithValue<int32_t> AudioStream::waitForAvailableFrames(int32_t numFrames,
     if (numFrames == 0) return Result::OK;
     if (numFrames < 0) return Result::ErrorOutOfRange;
 
+    // Make sure we don't try to wait for more frames than the buffer can hold.
+    // Subtract framesPerBurst because this is often called from a callback
+    // and we don't want to be sleeping if the buffer is close to overflowing.
+    const int32_t maxAvailableFrames = getBufferCapacityInFrames() - getFramesPerBurst();
+    numFrames = std::min(numFrames, maxAvailableFrames);
+
     int64_t framesAvailable = 0;
     int64_t burstInNanos = getFramesPerBurst() * kNanosPerSecond / getSampleRate();
     bool ready = false;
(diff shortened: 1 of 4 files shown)


src/common/AudioStream.cpp
```diff
@@ -166,6 +166,12 @@ ResultWithValue<int32_t> AudioStream::waitForAvailableFrames(int32_t numFrames,
     if (numFrames == 0) return Result::OK;
     if (numFrames < 0) return Result::ErrorOutOfRange;
 
+    // Make sure we don't try to wait for more frames than the buffer can hold.
+    // Subtract framesPerBurst because this is often called from a callback
+    // and we don't want to be sleeping if the buffer is close to overflowing.
+    const int32_t maxAvailableFrames = getBufferCapacityInFrames() - getFramesPerBurst();
```
robertwu1, line 172: Is there ever a case where getFramesPerBurst is equal to getBufferCapacityInFrames?

philburk: In general the capacity would always be at least one burst.
But now that you mention it the code seems unsafe. I will clip against zero or one burst.

robertwu1 (approved):

PR #1711, opened 2023-02-21
Coerce positive error from open() to negative code

philburk:
AAudio could sometimes return a positive number (1) as an error code. That was confusing to apps like OboeTester, which often look for (result < 0).

The fix is to coerce these bogus error codes to ErrorInternal.

Bump Oboe to 1.7.2 and OboeTester to 2.3.7.

### DIFF
--- a/src/aaudio/AudioStreamAAudio.cpp
+++ b/src/aaudio/AudioStreamAAudio.cpp
@@ -332,17 +332,17 @@ Result AudioStreamAAudio::open() {
 
 error2:
     mLibLoader->builder_delete(aaudioBuilder);
-    LOGD("AudioStreamAAudio.open: AAudioStream_Open() returned %s = %d",
-         mLibLoader->convertResultToText(static_cast<aaudio_result_t>(result)),
-         static_cast<int>(result));
     if (static_cast<int>(result) > 0) {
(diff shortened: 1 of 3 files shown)


src/aaudio/AudioStreamAAudio.cpp
```diff
@@ -332,8 +332,18 @@ Result AudioStreamAAudio::open() {
...
-         mLibLoader->convertResultToText(static_cast<aaudio_result_t>(result)));
+    LOGD("AudioStreamAAudio.open: AAudioStream_Open() returned %s = %d",
+         mLibLoader->convertResultToText(static_cast<aaudio_result_t>(result)),
+         static_cast<int>(result));
+    if (static_cast<int>(result) > 0) {
+        // Possibly due to b/267531411
+        if (OboeGlobals::areWorkaroundsEnabled()) {
+            result = Result::ErrorInternal; // Coerce to negative error.
+        } else {
+            LOGW("AudioStreamAAudio.open: AAudioStream_Open() returned positive error = %d",
```
robertwu1, line 343: We should add this warning message for both cases for easier debugging

robertwu1 (approved):

PR #1716, opened 2023-02-24
OboeTester: catch null from getChannelMapping()

### LINKED ISSUE #1715: OboeTester: DeviceReport crashing on null mapping
micInfo.getChannelMapping()
is supposed to return an "Empty list when the MicrophoneInfo is returned by AudioManager.getMicrophones()."
See [b/270537789](http://b/270537789) | P4 | [Oboe Tester] : Oboe tester app crashed when click on DEVICE REPORT

Detect the null and show a Toast.

philburk:
Fixes #1715

### DIFF
--- a/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/DeviceReportActivity.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/DeviceReportActivity.java
@@ -25,7 +25,9 @@ import android.media.AudioManager;
 import android.media.MicrophoneInfo;
 import android.os.Build;
 import android.os.Bundle;
+import android.util.Log;
 import android.widget.TextView;
+import android.widget.Toast;
 
 import com.mobileer.audio_device.AudioDeviceInfoConverter;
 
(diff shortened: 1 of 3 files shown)


apps/OboeTester/app/src/main/java/com/mobileer/oboetester/DeviceReportActivity.java
```diff
@@ -127,8 +129,12 @@ public String reportAllMicrophones() {
                     report.append(micItem);
                 }
             } catch (IOException e) {
-                e.printStackTrace();
+                Log.e(TestAudioActivity.TAG, "Caught ", e);
```
robertwu1, line 132: Should we show error toast for IOExceptions?

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/MicrophoneInfoConverter.java
```diff
@@ -60,6 +60,9 @@ public static String reportMicrophoneInfo(MicrophoneInfo micInfo) {
         sb.append("\nType       : " + micInfo.getType());
 
         List<Pair<Integer, Integer>> mapping = micInfo.getChannelMapping();
+        if (mapping == null) {
```
robertwu1, line 63: Are we sure null is an error? What about if this is empty?

The documentation for the function seems a bit weird https://developer.android.com/reference/android/media/MicrophoneInfo#getChannelMapping()

robertwu1 (changes requested):

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/MicrophoneInfoConverter.java
philburk: Jiabin says yes, it is a regression.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/DeviceReportActivity.java
philburk: We did not do it before and I assume there was a reason, which I don't recall.

PR #1729, opened 2023-03-03
Do not submit: testing Doxyfile

robertwu1:


robertwu1:
Abandoning because there is no real value in updating the index file at https://google.github.io/oboe/index.html. You can update https://github.com/google/oboe/blob/main/include/oboe/Oboe.h#L23 if you want the section to be longer but the blurb there seems good enough as is.

PR #1733, opened 2023-03-06
OboeTester: Fix and add more properties in DeviceReport

### LINKED ISSUE #1732: OboeTester DeviceReport is reading properties incorrectly
Instead of using AudioManager.getParameters(), it should AudioManager.getProperty()

robertwu1:
Fixes #1732


philburk (approved):
Thanks!

PR #1739, opened 2023-03-08
OboeTester: Use AndroidX

### LINKED ISSUE #1740: Update OboeTester to use AndroidX
AndroidX is now recommended as the best practice

robertwu1:
Fixes #1740

### DIFF
--- a/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/AnalyzerActivity.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/AnalyzerActivity.java
@@ -16,16 +16,15 @@
 
 package com.mobileer.oboetester;
 
-import androidx.annotation.NonNull;
-import androidx.core.app.ActivityCompat;
-import androidx.core.content.ContextCompat;
-
 import android.Manifest;
 import android.content.pm.PackageInfo;
 import android.content.pm.PackageManager;
 import android.os.Build;
 import android.os.Bundle;
 import android.widget.Toast;
+import androidx.annotation.NonNull;
+import androidx.core.app.ActivityCompat;
+import androidx.core.content.ContextCompat;
 
 import java.io.File;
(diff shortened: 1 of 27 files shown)


apps/OboeTester/app/src/main/java/com/mobileer/oboetester/AnalyzerActivity.java
```diff
@@ -78,27 +79,27 @@ protected String getCommonTestReport() {
         report.append("build.fingerprint = " + Build.FINGERPRINT + "\n");
         try {
             PackageInfo pinfo = getPackageManager().getPackageInfo(getPackageName(), 0);
-            report.append(String.format("test.version = %s\n", pinfo.versionName));
-            report.append(String.format("test.version.code = %d\n", pinfo.versionCode));
+            report.append(String.format(Locale.getDefault(), "test.version = %s\n", pinfo.versionName));
```
philburk, line 82: You don't need to pass Locale.getDefault(). That is the default.
https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#format(java.lang.String,%20java.lang.Object...)

philburk (changes requested):

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/AnalyzerActivity.java
robertwu1: You get the warning `
Implicitly using the default locale is a common source of bugs: Use String.format(Locale, ...) instead` if you don't do this

robertwu1: Updated to Locale.US

philburk (approved):
This makes the code look messy without adding any functionality.
But it gets rid of an annoying Android Studio warning.
And it helps Google3.
OK.

PR #1751, opened 2023-03-14
Add privacy sensitive mode for API 30+

### LINKED ISSUE #1093: Support overriding input stream privacy sensitivity
Android 11 added [AAudioStreamBuilder_setPrivacySensitive()](https://developer.android.com/ndk/reference/group/audio#aaudiostreambuilder_setprivacysensitive) for AAudio. This API allows to override default privacy sensitivity of input stream which is otherwise based on input preset.

Oboe's [AudioStreamBuilder](https://github.com/google/oboe/blob/master/include/oboe/AudioStreamBuilder.h) should have the corresponding property.

When specified, it shall override default privacy sensitivity. When not specified, it shall not invoke the above AAudio API and thus lets the system to decide the privacy sensitivity.

robertwu1:
setPrivacySensitive/isPrivacySensitive are APIs in AAudio for overriding the default input presets. We are adding them to Oboe here.

Instead of using a boolean like in AAudio, Oboe is using a tristate of Unspecified/Enabled/Disabled. This allows a default implementation to be used and a fallback pattern for OpenSLES. Under the hood, AAudio also uses a tristate for this.

Fixes #1093


src/aaudio/AAudioLoader.h
```diff
@@ -141,7 +150,7 @@ class AAudioLoader {
 
     typedef int32_t (*signature_I_PSKPLPL)(AAudioStream *, clockid_t, int64_t *, int64_t *);
 
-    typedef bool    (*signature_B_PS)(AAudioStream *);
+    typedef bool    (*signature_O_PS)(AAudioStream *);
```
philburk, line 153: Thanks for spotting this.

tests/testStreamOpen.cpp
```diff
@@ -460,3 +460,69 @@ TEST_F(StreamOpenOutput, OboeExtensions){
         ASSERT_TRUE(OboeExtensions::isMMapEnabled());
     }
 }
+
+TEST_F(StreamOpenInput, AAudioInputSetPrivacySensitiveModeUnspecifiedMedia){
```
philburk, line 464: Should name say Unprocessed?

tests/testStreamOpen.cpp
```diff
@@ -460,3 +460,69 @@ TEST_F(StreamOpenOutput, OboeExtensions){
...
+    if (getSdkVersion() >= __ANDROID_API_R__){
```
philburk, line 465: You could run these tests on <R and then conditionally choose different expected results. Then we can check if it causes crashes.

philburk (changes requested):

tests/testStreamOpen.cpp
robertwu1: Added a separate test for <R below. Adding more if statements makes the test confusing

tests/testStreamOpen.cpp
```diff
@@ -526,3 +526,13 @@ TEST_F(StreamOpenInput, OpenSLESInputSetPrivacySensitiveModeGetsUnspecified){
     ASSERT_EQ(mStream->getPrivacySensitiveMode(), PrivacySensitiveMode::Unspecified);
     ASSERT_TRUE(closeStream());
 }
+
+TEST_F(StreamOpenInput, OldAndroidVersionInputSetPrivacySensitiveModeGetsUnspecified){
+    if (getSdkVersion() < __ANDROID_API_R__) {
+        mBuilder.setDirection(Direction::Input);
+        mBuilder.setInputPreset(InputPreset::VoiceCommunication);
```
philburk, line 533: If you try to set privacy when <R does it crash?

philburk (changes requested):

tests/testStreamOpen.cpp
robertwu1: No, it does not crash. It simply ignores the parameter.

philburk (changes requested):

tests/testStreamOpen.cpp
robertwu1: Oops. Nice catch!

PR #1755, opened 2023-03-17
Add support for AAudioStream_release

### LINKED ISSUE #1754: Add support for AAudioStream_release
[AAudioStream_release](https://developer.android.com/ndk/reference/group/audio#aaudiostream_release) was added in Android R. Should be added to Oboe as well.

robertwu1:
Fixes #1754

### DIFF
--- a/src/aaudio/AudioStreamAAudio.cpp
+++ b/src/aaudio/AudioStreamAAudio.cpp
@@ -362,6 +362,12 @@ Result AudioStreamAAudio::release() {
         return Result::ErrorUnimplemented;
     }
 
+    // AAudioStream_release() is buggy on Android R.
+    if (OboeGlobals::areWorkaroundsEnabled() && getSdkVersion() == __ANDROID_API_R__) {
+        LOGW("Skipping release() on Android R");
+        return Result::ErrorUnimplemented;
+    }
+
     std::lock_guard<std::mutex> lock(mLock);
     AAudioStream *stream = mAAudioStream.load();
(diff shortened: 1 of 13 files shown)


src/aaudio/AudioStreamAAudio.cpp
```diff
@@ -357,6 +357,26 @@ Result AudioStreamAAudio::open() {
...
+    }
+
+    std::lock_guard<std::mutex> lock(mLock);
+    AAudioStream *stream = mAAudioStream.load();
+    if (stream != nullptr) {
+        if (OboeGlobals::areWorkaroundsEnabled()) {
+            // Make sure we are really stopped. Do it under mLock
+            // so another thread cannot call requestStart() right before the close.
+            requestStop_l(stream);
+            sleepBeforeClose();
```
philburk, line 372: We should not have to sleep here because we are not deleting the stream objects. The autodoc for release() says the callbacks may still be active.

philburk (changes requested):
Nice! Thanks for adding this.

We should check the bug log. I vaguely recall that the first version of AAudioStream_release() may not have been fully implemented.

robertwu1:
> Nice! Thanks for adding this.
> 
> We should check the bug log. I vaguely recall that the first version of AAudioStream_release() may not have been fully implemented.

Seems like b/163165126 was fixed in a R QPR. I added a workaround regardless in case some phones do not have the QPR. Thanks!

PR #1758, opened 2023-03-21
Add IEC61937 for API 34 

### LINKED ISSUE #1757: Add IEC61937 to Oboe for API 34
Added in AAudio. Should add in Oboe too

robertwu1:
Fixes #1757

### DIFF
--- a/apps/OboeTester/app/src/main/cpp/FormatConverterBox.cpp
+++ b/apps/OboeTester/app/src/main/cpp/FormatConverterBox.cpp
@@ -28,6 +28,7 @@ FormatConverterBox::FormatConverterBox(int32_t numSamples,
     mSource.reset();
     switch (mInputFormat) {
         case oboe::AudioFormat::I16:
+        case oboe::AudioFormat::IEC61937:
             mSource = std::make_unique<oboe::flowgraph::SourceI16>(1);
             break;
         case oboe::AudioFormat::I24:
@@ -46,6 +47,7 @@ FormatConverterBox::FormatConverterBox(int32_t numSamples,
     mSink.reset();
     switch (mOutputFormat) {
         case oboe::AudioFormat::I16:
+        case oboe::AudioFormat::IEC61937:
             mSink = std::make_unique<oboe::flowgraph::SinkI16>(1);
             break;
         case oboe::AudioFormat::I24:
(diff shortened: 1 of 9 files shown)


apps/OboeTester/app/src/main/java/com/mobileer/oboetester/StreamConfiguration.java
```diff
@@ -493,6 +494,8 @@ public static String convertFormatToText(int format) {
                 return "I32";
             case AUDIO_FORMAT_PCM_FLOAT:
                 return "Float";
+            case AUDIO_FORMAT_IEC61937:
+                return "Float";
```
flamme, line 498: iec61937

flamme (approved):
The PR looks good to me. Thanks for making the change!

apps/OboeTester/app/src/main/cpp/FormatConverterBox.cpp
```diff
@@ -28,6 +28,7 @@ FormatConverterBox::FormatConverterBox(int32_t numSamples,
     mSource.reset();
     switch (mInputFormat) {
         case oboe::AudioFormat::I16:
+        case oboe::AudioFormat::IEC61937:
```
philburk, line 31: IEC61937 cannot go through the flowgraph because it will get corrupted. It must be "bit perfect".

robertwu1: Yup! See QuickManager.cpp. This is for OboeTester so I think this is ok as is.

philburk:
Can this be tested?

robertwu1:
> Can this be tested?

You can test opening the stream with OboeTester.

By looking at the device report, if encoding [13](https://developer.android.com/reference/android/media/AudioFormat#ENCODING_IEC61937) is supported, then this should open for API 34+.

philburk:
You can test using OboeTester by selecting IEC61937
then select SampleRate=32000
then select SRC Quality::Medium
Oboe should fail on Open.

robertwu1:
> You can test using OboeTester by selecting IEC61937 then select SampleRate=32000 then select SRC Quality::Medium Oboe should fail on Open.

Good test! This works for SampleRate=32000 but fails for SampleRate=96000 on the TV I tested as 32000 is supported on the TV but not 96000.

PR #1760, opened 2023-03-24
oboetester: try to repro routing crash

philburk:
This test tries to trigger a routing change
while long duration callbacks are playing.
This can cause the callback to write into a dead buffer and cause a SIGSEGV.

See b/274815060

### DIFF
--- a/apps/OboeTester/app/src/main/cpp/TestRoutingCrash.cpp
+++ b/apps/OboeTester/app/src/main/cpp/TestRoutingCrash.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022 The Android Open Source Project
+ * Copyright 2023 The Android Open Source Project
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/apps/OboeTester/app/src/main/cpp/TestRoutingCrash.cpp
+++ b/apps/OboeTester/app/src/main/cpp/TestRoutingCrash.cpp
@@ -1,0 +1,90 @@
+/*
+ * Copyright 2022 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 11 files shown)


apps/OboeTester/app/src/main/cpp/TestRoutingCrash.cpp
```diff
@@ -0,0 +1,90 @@
+/*
+ * Copyright 2022 The Android Open Source Project
```
flamme, line 2: 2023?

apps/OboeTester/app/src/main/cpp/TestRoutingCrash.h
```diff
@@ -0,0 +1,67 @@
+/*
+ * Copyright 2022 The Android Open Source Project
```
flamme, line 2: 2023?

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestRouteDuringCallbackActivity.java
```diff
@@ -0,0 +1,141 @@
+/*
+ * Copyright 2022 The Android Open Source Project
```
flamme, line 2: 2023?

flamme (commented):
Thanks Phil for the PR!

The restoring the track seems pretty relies on the timing to making a crash. Is it possible to have a call from JNI to JAVA when receiving a data callback to force routing change?

apps/OboeTester/app/src/main/res/layout/activity_routing_crash.xml
```diff
@@ -0,0 +1,46 @@
...
+        android:layout_height="wrap_content"
+        android:orientation="vertical">
+
+        <TextView
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:fontFamily="monospace"
+            android:gravity="bottom"
+            android:text="
+Try to cause a crash by changing the\n
```
robertwu1, line 21: Put into strings.xml?

apps/OboeTester/app/src/main/cpp/TestRoutingCrash.cpp
philburk: Done

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestRouteDuringCallbackActivity.java
philburk: Done

apps/OboeTester/app/src/main/res/layout/activity_routing_crash.xml
philburk: Done

PR #1776, opened 2023-04-12
Add spatialization functions to Oboe

### LINKED ISSUE #1612: Support spatialization behavior in oboe
Spatialization behavior is added in API 32 in AAudio. Oboe need to support this API so that the apps can sets the behavior affecting whether spatialization will be used.

What's more, the framework will put spatialization output at a higher priority than low latency when it is possible. It may worth to discuss if oboe want to disable spatialization on stereo stream by default.

robertwu1:
Fixes #1612

### DIFF
--- a/include/oboe/AudioStreamBuilder.h
+++ b/include/oboe/AudioStreamBuilder.h
@@ -374,6 +374,38 @@ public:
     }
 
     /**
+     * Specifies whether the audio data of this output stream has already been processed for spatialization.
+     *
+     * If the stream has been processed for spatialization, setting this to true will prevent issues such as
+     * double-processing on platforms that will spatialize audio data.
+     *
+     * Available since API level 32.
+     *
+     * @param isContentSpatialized whether the content is already spatialized
+     * @return pointer to the builder so calls can be chained
+     */
+    AudioStreamBuilder *setIsContentSpatialized(bool isContentSpatialized) {
+        mIsContentSpatialized = isContentSpatialized;
+        return this;
+    }
+
+    /**
+     * Sets the behavior affecting whether spatialization will be used.
+     *
+     * The AAudio system will use this information to select whether the stream will go through a
+     * spatializer effect or not when the effect is supported and enabled.
+     *
+     * Available since API level 32.
+     *
+     * @param spatializationBehavior the desired spatialization behavior
+     * @return pointer to the builder so calls can be chained
+     */
+    AudioStreamBuilder *setSpatializationBehavior(SpatializationBehavior spatializationBehavior) {
+        mSpatializationBehavior = spatializationBehavior;
+        return this;
+    }
+
+    /**
      * Specifies an object to handle data related callbacks from the underlying API.
      *
      * <strong>Important: See AudioStreamCallback for restrictions on what may be called
--- a/src/aaudio/AudioStreamAAudio.cpp
+++ b/src/aaudio/AudioStreamAAudio.cpp
@@ -269,6 +269,15 @@ Result AudioStreamAAudio::open() {
                 mPrivacySensitiveMode == PrivacySensitiveMode::Enabled);
     }
 
+    if (mLibLoader->builder_setIsContentSpatialized != nullptr) {
+        mLibLoader->builder_setIsContentSpatialized(aaudioBuilder, mIsContentSpatialized);
+    }
+
+    if (mLibLoader->builder_setSpatializationBehavior != nullptr) {
+        mLibLoader->builder_setSpatializationBehavior(aaudioBuilder,
+                static_cast<aaudio_spatialization_behavior_t>(mSpatializationBehavior));
+    }
+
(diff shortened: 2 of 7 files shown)


include/oboe/AudioStreamBuilder.h
```diff
@@ -373,6 +373,38 @@ class AudioStreamBuilder : public AudioStreamBase {
...
+     * @param isContentSpatialized whether the content is already spatialized
+     * @return pointer to the builder so calls can be chained
+     */
+    AudioStreamBuilder *setIsContentSpatialized(bool isContentSpatialized) {
+        mIsContentSpatialized = isContentSpatialized;
+        return this;
+    }
+
+    /**
+     * Sets the behavior affecting whether spatialization will be used.
```
philburk, line 393: Document the default if this is not called.

include/oboe/AudioStreamBuilder.h
```diff
@@ -373,6 +373,38 @@ class AudioStreamBuilder : public AudioStreamBase {
...
 
+    /**
+     * Specifies whether the audio data of this output stream has already been processed for spatialization.
+     *
+     * If the stream has been processed for spatialization, setting this to true will prevent issues such as
+     * double-processing on platforms that will spatialize audio data.
+     *
+     * Available since API level 32.
+     *
+     * @param isContentSpatialized whether the content is already spatialized
```
philburk, line 384: Document the default if this is not called.

tests/testStreamOpen.cpp
```diff
@@ -437,6 +441,50 @@ TEST_F(StreamOpenInput, AAudioInputSetAttributionTag){
...
+}
+
+TEST_F(StreamOpenOutput, AAudioOutputSetSpatializationBehaviorUnspecified) {
+    mBuilder.setDirection(Direction::Output);
+    mBuilder.setSpatializationBehavior(SpatializationBehavior::Unspecified);
+    ASSERT_TRUE(openStream());
+    if (getSdkVersion() >= __ANDROID_API_S_V2__){
+        ASSERT_EQ(mStream->getSpatializationBehavior(), SpatializationBehavior::Never);
+    } else {
+        ASSERT_EQ(mStream->getSpatializationBehavior(), SpatializationBehavior::Unspecified);
```
philburk, line 467: Maybe this should be Never because you will never get spatialization before SC_V2.

tests/testStreamOpen.cpp
```diff
@@ -437,6 +441,50 @@ TEST_F(StreamOpenInput, AAudioInputSetAttributionTag){
...
+    mBuilder.setDirection(Direction::Output);
+    mBuilder.setSpatializationBehavior(SpatializationBehavior::Never);
+    ASSERT_TRUE(openStream());
+    ASSERT_EQ(mStream->getSpatializationBehavior(), SpatializationBehavior::Never);
+    ASSERT_TRUE(closeStream());
+}
+
+TEST_F(StreamOpenOutput, AAudioOutputSetSpatializationBehaviorUnspecified) {
+    mBuilder.setDirection(Direction::Output);
+    mBuilder.setSpatializationBehavior(SpatializationBehavior::Unspecified);
```
philburk, line 462: Add In/Out tests that does not call setSpatializationBehavior() to make sure we get the same Never.

tests/testStreamOpen.cpp
```diff
@@ -437,6 +441,50 @@ TEST_F(StreamOpenInput, AAudioInputSetAttributionTag){
...
+        ASSERT_EQ(mStream->getSpatializationBehavior(), SpatializationBehavior::Never);
+    } else {
+        ASSERT_EQ(mStream->getSpatializationBehavior(), SpatializationBehavior::Unspecified);
+    }
+    ASSERT_TRUE(closeStream());
+}
+
+TEST_F(StreamOpenInput, AAudioInputSetIsContentSpatialized) {
+    mBuilder.setDirection(Direction::Input);
+    mBuilder.setIsContentSpatialized(true);
```
philburk, line 474: Add In/Out tests that does not call setIsContentSpatialized() to make sure we get false.

philburk (changes requested):

src/aaudio/AudioStreamAAudio.cpp
```diff
@@ -271,6 +271,8 @@ Result AudioStreamAAudio::open() {
 
     if (mLibLoader->builder_setIsContentSpatialized != nullptr) {
         mLibLoader->builder_setIsContentSpatialized(aaudioBuilder, mIsContentSpatialized);
+    } else {
```
philburk, line 274: isContentSpatialized refers to the content itself, not a feature of Android. You can play spatialized audio on old versions of Android if the content is already spatialized.  So maybe it it is OK to just set mIsContentSpatialized to whatever the app says. On old versions it will be ignored. On new versions of Android AAudio will care about the flag.

PR #1785, opened 2023-04-20
OboeTester: check for callback size in InputAnalyzer

robertwu1:
InputAnalyzer creates a buffer based on the burst size for callbacks.

This did not account for the fact that the user can manually override the callback size.

Fixes b/278942008

### DIFF
--- a/apps/OboeTester/app/src/main/cpp/NativeAudioContext.h
+++ b/apps/OboeTester/app/src/main/cpp/NativeAudioContext.h
@@ -347,7 +347,7 @@ protected:
 
     oboe::Result startStreams() override {
         mInputAnalyzer.reset();
-        mInputAnalyzer.setup(getInputStream()->getFramesPerBurst(),
+        mInputAnalyzer.setup(std::max(getInputStream()->getFramesPerBurst(), callbackSize),
                              getInputStream()->getChannelCount(),
                              getInputStream()->getFormat());
         return getInputStream()->requestStart();


apps/OboeTester/app/src/main/cpp/NativeAudioContext.h
```diff
@@ -347,7 +347,7 @@ class ActivityTestInput : public ActivityContext {
 
     oboe::Result startStreams() override {
         mInputAnalyzer.reset();
-        mInputAnalyzer.setup(getInputStream()->getFramesPerBurst(),
+        mInputAnalyzer.setup(std::max(getInputStream()->getFramesPerBurst(), callbackSize),
```
philburk, line 350: We also have ActivityContext.getFramesPerBlock().
I wonder if that would also work.

robertwu1: I don't like ActivityContext.getFramesPerBlock(). This does not differentiate the input and output stream.

philburk: It seems to work.
We could also bounds check in FormatConverterBox. I am trying an assert.

philburk: Right. I am looking at NativeAudioContext.cpp:244
It may not handle streams with different burst sizes properly when there is both INPUT and OUTPUT.

philburk (approved):

PR #1788, opened 2023-04-21
Dynamic CPU Load test

philburk:
Measure glitches when the CPU load changes.

Try to use ADPF to reduce the glitching.


apps/OboeTester/app/src/main/cpp/jni-bridge.cpp
```diff
@@ -213,6 +226,15 @@ Java_com_mobileer_oboetester_OboeAudioStream_getBufferSizeInFrames(
     return result;
 }
 
+JNIEXPORT void JNICALL
+Java_com_mobileer_oboetester_OboeAudioStream_setPerformanceHintEnabled(
+        JNIEnv *env, jobject, jint streamIndex, jboolean enabled) {
+    std::shared_ptr<oboe::AudioStream> oboeStream = engine.getCurrentActivity()->getStream(streamIndex);
+    if (oboeStream != nullptr) {
+        oboeStream->setPerformanceHintEnabled(enabled);
```
robertwu1, line 234: If the user checks the checkbox before the stream is opened, would it be ignored? Should we either disable the checkbox before the stream is actually opened?

philburk: Good idea. I disabled it.

apps/OboeTester/app/src/main/cpp/synth/SynthTools.h
```diff
@@ -0,0 +1,173 @@
...
+                                       synth_float_t *output,
+                                       int32_t numSamples) {
+        for (int i = 0; i < numSamples; i++) {
+            *output++ = *input1++ * *input2;
+        }
+    }
+
+    static double convertTimeToExponentialScaler(synth_float_t duration, synth_float_t sampleRate) {
+        // Calculate scaler so that scaler^frames = target/source
+        double numFrames = duration * sampleRate;
```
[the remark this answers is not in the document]
philburk: Fixed.

src/opensles/AudioStreamOpenSLES.cpp
```diff
@@ -383,6 +383,9 @@ Result AudioStreamOpenSLES::close_l() {
     EngineOpenSLES::getInstance().close();
 
     setState(StreamState::Closed);
+
+    closePerformanceHint();
```
robertwu1, line 387: Where is this defined for OpenSL ES? I only see a definition in AAudio

src/common/AdpfWrapper.cpp
```diff
@@ -0,0 +1,123 @@
...
+    int result = loadAphFunctions();
+    if (result < 0) return result;
+
+    // This is a singleton.
+    APerformanceHintManager* manager = gAPH_getManagerFn();
+
+    int32_t thread32 = threadId;
+    if (sUseAlternativeHack) {
+        // FIXME hack because setprop not working!
+        targetDurationNanos = (targetDurationNanos & ~0xFF) | 0xA5;
```
robertwu1, line 85: Can you add a comment of what 0xA5 represents?

src/aaudio/AudioStreamAAudio.cpp
```diff
@@ -146,6 +146,39 @@ void AudioStreamAAudio::internalErrorCallback(
...
+}
+
+void AudioStreamAAudio::endPerformanceHintInCallback(int32_t numFrames) {
+    if (mAdpfWrapper.isOpen()) {
+        double durationScaler = static_cast<double>(mFramesPerBurst) / numFrames;
+        // Skip this callback if the numFrames is too small.
+        // This can happen when buffers wrap around, particularly when doing sample rate conversion.
+        // If we have a short callback then just scale the duration as if it was a full burst.
+        if (durationScaler < 2.0) {
+            mAdpfWrapper.onEndCallback(durationScaler);
```
robertwu1, line 177: Should we still use durationScaler when doing the scaling math here?

src/common/AdpfWrapper.cpp
```diff
@@ -0,0 +1,123 @@
...
+    }
+}
+
+void AdpfWrapper::onBeginCallback() {
+    if (isOpen()) {
+        mBeginCallbackNanos = oboe::AudioClock::getNanoseconds(CLOCK_REALTIME);
+    }
+}
+
+void AdpfWrapper::onEndCallback(double durationScaler) {
```
robertwu1, line 116: This seems to be only called when durationScaler < 2.0. Is this expected?

src/common/AudioStream.cpp
```diff
@@ -213,4 +219,5 @@ void AudioStream::calculateDefaultDelayBeforeCloseMillis() {
          static_cast<int>(mDelayBeforeCloseMillis));
 }
 
+
```
robertwu1, line 222: Extra line

src/opensles/AudioStreamOpenSLES.cpp
philburk: I removed the call. It is not supported in OpenSL ES.

src/aaudio/AudioStreamAAudio.cpp
philburk: Yes, we are calculating the scaling factor here and passing it to the ADPF wrapper which applies the scaler.
And also deciding whether we have enough data to report.

src/common/AudioStream.cpp
philburk: Done

src/common/AdpfWrapper.cpp
philburk: It is OK to skip this if we have a short callback to reduce error. ADPF still works on sparse data.

robertwu1 (approved):

PR #1804, opened 2023-05-16
DrumThumper: Retry stream open/start

### LINKED ISSUE #1628: DrumThumper is not restarting a stream when start() fails.
Android version(s): master
Android device(s): Pixel 6
Oboe version: main

**Steps to reproduce**

I hacked AAudioStream_requestStart() so that it returns AAUDIO_ERROR_DISCONNECTED the first time it is called.
This simulates a rare condition where a device is unplugged right when it is started. It happens.
I then started DrumThumper and played drums.

**Expected behavior**
DrumThumper should close the old stream and reopen a new stream if the start call fails.
Drums should then make sound.

**Actual behavior**

No sound from the drums.

robertwu1:
Fixes #1628

### DIFF
--- a/samples/iolib/src/main/cpp/player/SimpleMultiPlayer.cpp
+++ b/samples/iolib/src/main/cpp/player/SimpleMultiPlayer.cpp
@@ -116,16 +116,30 @@ bool SimpleMultiPlayer::openStream() {
 }
 
 bool SimpleMultiPlayer::startStream() {
-    Result result = mAudioStream->requestStart();
-    if (result != Result::OK){
-        __android_log_print(
-                ANDROID_LOG_ERROR,
-                TAG,
-                "requestStart failed. Error: %s", convertToText(result));
-        return false;
+    int tryCount = 0;
+    while (tryCount < 3) {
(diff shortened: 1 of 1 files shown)


samples/iolib/src/main/cpp/player/SimpleMultiPlayer.cpp
```diff
@@ -116,16 +116,30 @@ bool SimpleMultiPlayer::openStream() {
...
-                ANDROID_LOG_ERROR,
-                TAG,
-                "requestStart failed. Error: %s", convertToText(result));
-        return false;
+    int tryCount = 0;
+    while (tryCount < 3) {
+        bool wasOpenSuccessful = true;
+        if (tryCount > 0) {
+            wasOpenSuccessful = openStream(); // Try to open the stream again after the first try.
+            usleep(20 * 1000); // Sleep between tries to give the system time to settle.
```
philburk, line 124: Sleep before the open so that there is a gap between the close() and the open().

philburk (changes requested):

philburk (approved):

PR #1811, opened 2023-05-23
OboeTester: consolidate CommDevice UI

philburk:
Move SCO dialog to CommDeviceView

To help debug b/282831177

### DIFF
--- a/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/CommunicationDeviceView.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/CommunicationDeviceView.java
@@ -105,9 +105,7 @@ public class CommunicationDeviceView extends LinearLayout {
                 AudioDeviceInfo[] commDeviceArray = mDeviceSpinner.getCommunicationsDevices();
                 if (commDeviceArray != null) {
                     if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
-                        if (position == CommunicationDeviceSpinner.POS_NOOP) {
-                            // do nothing
-                        } else if (position == CommunicationDeviceSpinner.POS_CLEAR) {
+                        if (position == CommunicationDeviceSpinner.POS_CLEAR) {
                             mAudioManager.clearCommunicationDevice();
                         } else {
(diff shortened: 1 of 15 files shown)


apps/OboeTester/app/src/main/java/com/mobileer/oboetester/CommunicationDeviceView.java
```diff
@@ -122,14 +151,41 @@ private void setSpeakerPhoneOn(boolean enabled) {
...
             if (commDeviceInfo != null) {
-                text += ", CommDev=" + commDeviceInfo.getId();
+                text += ", #" + commDeviceInfo.getId();
             }
         }
-        mIsSpeakerphoneText.setText(" => " + text);
+        mSpeakerStatusView.setText(text + ",");
+
+        if (mScoState == AudioManager.SCO_AUDIO_STATE_CONNECTING) {
+            text = ":wait";
```
robertwu1, line 164: nit: wait is in a different font as CON and DISCO

apps/OboeTester/app/src/main/res/layout/activity_main.xml
```diff
@@ -182,42 +182,15 @@
         app:layout_constraintEnd_toEndOf="parent"
         app:layout_constraintStart_toEndOf="@+id/textView" />
 
-    <LinearLayout
-        android:id="@+id/layoutBluetoothSco"
```
robertwu1, line 186: We may want to just test swapping between SCO and A2DP outputs. We won't be able to test this if we move this into input activities.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/CommunicationDeviceView.java
philburk: Also change to DISCON

apps/OboeTester/app/src/main/res/layout/activity_main.xml
philburk: TEST OUTPUT does have the UI.
We should add it to TapToTone, etc.

robertwu1 (approved):

PR #1813, opened 2023-05-27
resampler: do not filter when upsampling

### LINKED ISSUE #1625: Resampler has very low output when converting from 11025 to 88200
Enable the unit test in testResampler.cpp to reproduce this.

There is an initial hump and then the output is shaped correctly but with very low amplitude.

### LINKED ISSUE #1673: Unexpected low input bandwidth with Oboe vs OpenSL in some configurations
Android version(s): 10 and 11, possibly others
Android device(s):  Oppo and Moto devices, possibly others
Oboe version: 1.6.1
App name used for testing: Sing!

 
**Short description**
This may be more of a question about how best to configure our stream than a proper bug report, but we're receiving unexpected results with Oboe/AAudio versus our prior OpenSL implementation.

We've received several reports of "muffled voice" recordings from our Oboe implementation. Inspecting the frequency content of the uploads it appears these vocal files had a capture sample rate of about 16k. Users perceive this as a downgrade from our Open SL implementation, where we can see with the same input device the rate was apparently more like 32k.

For Open SL we usually set sample rate according to `AudioManager.PROPERTY_OUTPUT_SAMPLE_RATE` to have the best chance of getting the low latency path.

For Oboe we pass a sample rate (again from `AudioManager.PROPERTY_OUTPUT_SAMPLE_RATE`) and request conversion so the underlying stream sees `oboe::Unspecified`. I recently noticed this comment in `QuirksManager.cpp` about this behavior:

```
    // If a SAMPLE RATE is specified for low latency, let the native code choose an optimal rate.
    // This isn't really a workaround. It is an Oboe feature that is convenient to place here.
    // TODO There may be a problem if the devices supports low latency
    //      at a higher rate than the default.
```

This reminded me of these user reports and I wonder if it's pointing in the general direction of the problem. I have also realized our requirements may be a little different from what we're requesting, and there may be a point where we would favor higher input sample rate over a low latency stream if we can't have both.

Questions:

* Will AAudio sometimes select a lower sample rate than necessary for input? Is there a way to ensure this doesn't occur?
* Is there a way to ensure we get the highest sample rate for a given configuration with the current set of APIs, perhaps by determining the set of input sample rates that would result in a low latency stream so we can select the highest?

**Steps to reproduce**
We have been unable to reproduce this in-house. We see this primarily for wired headset mics.

**Expected behavior**
AAudio provides the highest possible input sample rate for low latency input streams, or a way to specify this behavior.

**Actual behavior**
We sometimes unexpectedly see a low capture sample rate.

**Device**

moto g(8) power lite, CPH1931 (Oppo  A5), others

**Any additional context**

We can compile a more complete list of devices and OS versions if that would be helpful.

philburk:
The sound was muffled when upsampling for no reason.

Fixes #1673
Fixes #1625

### DIFF
--- a/src/flowgraph/resampler/MultiChannelResampler.cpp
+++ b/src/flowgraph/resampler/MultiChannelResampler.cpp
@@ -135,10 +135,9 @@ void MultiChannelResampler::generateCoefficients(int32_t inputRate,
     int coefficientIndex = 0;
     double phase = 0.0; // ranges from 0.0 to 1.0, fraction between samples
     // Stretch the sinc function for low pass filtering.
-    const float cutoffScaler = normalizedCutoff *
-            ((outputRate < inputRate)
-             ? ((float)outputRate / inputRate)
-             : ((float)inputRate / outputRate));
+    const float cutoffScaler = (outputRate < inputRate)
+             ? (normalizedCutoff * (float)outputRate / inputRate)
+             : 1.0f; // Do not filter when upsampling.
     const int numTapsHalf = getNumTaps() / 2; // numTaps must be even.
     const float numTapsHalfInverse = 1.0f / numTapsHalf;
     for (int i = 0; i < numRows; i++) {
--- a/src/flowgraph/resampler/MultiChannelResampler.h
+++ b/src/flowgraph/resampler/MultiChannelResampler.h
@@ -111,6 +111,9 @@ public:
          * Set lower to reduce aliasing.
          * Default is 0.70.
          *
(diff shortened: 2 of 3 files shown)


robertwu1:
Can you enable the commented out test in #1625 and add a couple more?

philburk:
I reenabled the unit test for 11025 to 88200. It passes.

I will add some more tests and also make some tests improvements in a later CL.

Note that running tests now seems to be dependent on fixing #1815

robertwu1 (approved):

PR #1818, opened 2023-05-31
OboeTester: stop CPU Load sniffer

### LINKED ISSUE #1817: OboeTester workload varying in TEST OUTPUT
STEPS
Tap TEST OUTPUT
Increase workload to 60 (voices)

EXPECT - %CPU to be fairly steady

ACTUAL - drops down to 2% then back up to 80%

The varying workload is from the "Dynamic CPU Load" test. Should be disabled for TEST OUPUT.

philburk:
Stop sniffer when onStop() called for Activity.
The sniffer was being left running the background. That was causing the workload to go up and down.

Fixes #1817

### DIFF
--- a/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/DynamicWorkloadActivity.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/DynamicWorkloadActivity.java
@@ -345,9 +345,13 @@ public class DynamicWorkloadActivity extends TestOutputActivityBase {
     }
 
     public void stopTest(View view) {
+        onStopTest();
+    }
+
+    @Override
+    public void onStopTest() {
         mUpdateThread.stop();
-        stopAudio();
-        closeAudio();
         updateButtons(false);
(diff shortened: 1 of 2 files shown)


apps/OboeTester/app/src/main/java/com/mobileer/oboetester/DynamicWorkloadActivity.java
```diff
@@ -345,9 +345,13 @@ public void startTest(View view) {
     }
 
     public void stopTest(View view) {
+        onStopTest();
+    }
+
+    @Override
+    public void onStopTest() {
         mUpdateThread.stop();
```
robertwu1, line 353: we should still stop and close audio right?

philburk: Yep!
super.onStopTest() calls stopAudio() and closeAudio().
Thanks for asking.

robertwu1 (approved):

PR #1823, opened 2023-06-09
OboeTester: add out_usage to automated testing

philburk:
Also change where the results files is stores.
No longer requires permissions but the path name is longer!

See apps/OboeTester/docs/AutomatedTesting.md

### DIFF
--- a/apps/OboeTester/app/build.gradle
+++ b/apps/OboeTester/app/build.gradle
@@ -6,8 +6,8 @@ android {
         applicationId = "com.mobileer.oboetester"
         minSdkVersion 23
         targetSdkVersion 33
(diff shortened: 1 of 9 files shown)


robertwu1:
Can you test this on some phones running older Android builds to see if there are no regressions? Especially before Android O.

philburk:
I found a Nexus 5 running 8.1 (O MR1).
It hung when I tried this test but not because of this PR.
It also hangs on the old code. See #1826
After that is fixed then this code works.

Note that the minSdkVersion is 23 or 6.0 (M).

robertwu1 (approved):

PR #1830, opened 2023-06-16
OboeTester: Add input device selector to Tap To Tone

### LINKED ISSUE #470: OboeTester: select built-in mic for Tap-To-Tone
This will help when measuring latency for BT devices that have a microphone.

robertwu1:
Fixes #470

### DIFF
--- a/apps/OboeTester/app/src/main/java/com/mobileer/audio_device/AudioDeviceListEntry.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/audio_device/AudioDeviceListEntry.java
@@ -34,6 +34,10 @@ public class AudioDeviceListEntry {
 
     private AudioDeviceInfo mDeviceInfo;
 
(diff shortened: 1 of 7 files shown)


philburk (commented):
Thanks for doing this. The intention was to just find the built-in mic and select it.
Is there an advantage in giving the user an option to pick the input from a menu?
Does it default to Built-in Mic type?

robertwu1:
> Thanks for doing this. The intention was to just find the built-in mic and select it. Is there an advantage in giving the user an option to pick the input from a menu? Does it default to Built-in Mic type?

I was testing the PR on a Pixel 6 and it has two built-in MICs. That's why I wanted to surface all audio sources.

philburk (changes requested):

philburk:
> I was testing the PR on a Pixel 6 and it has two built-in MICs. That's why I wanted to surface all audio sources.

OK.

philburk (approved):

PR #1838, opened 2023-06-26
remove unecessary streamContext.tester assignement

eronnen:
`streamContext.tester = AudioInputTester.getInstance();` is already executed at the start of the function


philburk (approved):

philburk:
@eronnen - Thanks for spotting this.

PR #1841, opened 2023-06-29
OboeTester: Add keep screen on

### LINKED ISSUE #1840: Keep Data Paths tests screen on
Tests like Data Path tests would benefit from us not having to constantly tap the screen

robertwu1:
Fixes #1840


philburk (commented):
This PR adds a checkbox for keeping the screen on.

What are the advantages or  disadvantages to just using FLAG_KEEP_SCREEN_ON whenever a long running test like DataPaths or Disconnect or AutoGlitch are active?

robertwu1:
> What are the advantages or disadvantages to just using FLAG_KEEP_SCREEN_ON whenever a long running test like DataPaths or Disconnect or AutoGlitch are active?

I enabled this by default for all audio activities.

Note that both this (keep screen on) and background enabled don't do anything to activities that don't inherit from TestAudioActivity like tap to tone. Is this fine?

robertwu1:
Seems to already exist with #467

PR #1850, opened 2023-07-06
OboeTester: add flush button

### LINKED ISSUE #1849: OboeTester should have a flush button
Will help with understanding where glitches comes from

robertwu1:
Fixes #1849

### DIFF
--- a/apps/OboeTester/app/src/main/cpp/NativeAudioContext.cpp
+++ b/apps/OboeTester/app/src/main/cpp/NativeAudioContext.cpp
@@ -301,7 +301,6 @@ oboe::Result ActivityContext::start() {
 
 oboe::Result ActivityContext::flush() {
     oboe::Result result = oboe::Result::OK;
-    stopBlockingIOThread();
     for (auto entry : mOboeStreams) {
         std::shared_ptr<oboe::AudioStream> oboeStream = entry.second;
         result = oboeStream->requestFlush();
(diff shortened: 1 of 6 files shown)


apps/OboeTester/app/src/main/cpp/NativeAudioContext.cpp
```diff
@@ -299,6 +299,16 @@ oboe::Result ActivityContext::start() {
     return result;
 }
 
+oboe::Result ActivityContext::flush() {
+    oboe::Result result = oboe::Result::OK;
+    stopBlockingIOThread();
```
philburk, line 304: This is a nice protection in a safe app.
But OboeTester should allow users to shoot themselves in the foot.
What happens if you flush a stream while another thread is writing to it?
Can we remove this stopBLockingIOThread()?
If it leads to a crash then we would like to understand why.

philburk (changes requested):

apps/OboeTester/app/src/main/cpp/NativeAudioContext.cpp
robertwu1: I removed it. Flush returns AAUDIO_ERROR_INVALID_STATE when the stream is running for AAudio so this is fine. Somehow, OpenSL ES output streams just stop but don't crash. Oboe doesn't implement flush for OpenSL ES input streams.

philburk (approved):
Looks great! Thanks for reordering the other methods and constants. Good attention to detail!

PR #1882, opened 2023-07-31
Add a limiter to Oboe flowgraph

### LINKED ISSUE #1881: Oboe should have a Limiter
An limiter was added to AAudio in Android U. We should use the same limiter for Oboe.

robertwu1:
Fixes #1881


src/common/DataConversionFlowGraph.cpp
```diff
@@ -181,6 +181,14 @@ Result DataConversionFlowGraph::configure(AudioStream *sourceStream, AudioStream
         }
     }
 
+    // For a pure float graph, there is chance that the data range may be very large.
```
philburk, line 184: "is a chance"

philburk (commented):
The DataConversionFlowgraph is only used when needed, not on all streams.
So some streams will get the limiter and some will not. That is inconsistent.
Also AAudio has its own Limiter. I think it is better to only have one limiter at the very last step, in the OS, where it is needed.

PR #1898, opened 2023-08-14
oboe: copy more parameters in FilterAudioStream

### LINKED ISSUE #1897: OboeTester TEST DISCONNECT confused about sharing mode
TEST DISCONNECT appears to get EXCLUSIVE mode on Mokey at 44100 Hz.
Mokey does not support MMAP so that is not possible.

REPRO STEPS
Launch OboeTester on Mokey or other device that does not support AAudio MMAP.
Press TEST DISCONNECT
Press RUN
Plug in headset.

EXPECT - actual sharing mode should be "SH" not "EX"

ACTUAL - got "EX"

Later a test is skipped

philburk:
sharingMode and the hardware formats were not getting copied.

It was noticed in OboeTester.
The test specs for TEST DISCONNECT were made more rational.

Also a unit test was added to catch the oboe bug.

Fixes #1897

### DIFF
--- a/tests/testStreamOpen.cpp
+++ b/tests/testStreamOpen.cpp
@@ -15,9 +15,11 @@
  */
 
 #include <gtest/gtest.h>
-#include <oboe/Oboe.h>
-#include <android/api-level.h>
 
+#include <aaudio/AAudioExtensions.h>
+#include <oboe/Oboe.h>
+
+#include <android/api-level.h>
 #ifndef __ANDROID_API_S__
 #define __ANDROID_API_S__ 31
 #endif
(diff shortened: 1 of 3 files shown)


tests/testStreamOpen.cpp
```diff
@@ -389,6 +391,27 @@ TEST_F(StreamOpenOutput, LowLatencyStreamHasSmallBufferSize){
...
+TEST_F(StreamOpenOutput, AAudioOutputSampleRate44100FilterConfiguration) {
+    if (mBuilder.isAAudioRecommended()) {
+        mBuilder.setDirection(Direction::Output);
+        mBuilder.setPerformanceMode(PerformanceMode::LowLatency);
+        mBuilder.setSharingMode(SharingMode::Exclusive);
+        // Try to force the use of a FilterAudioStream by requesting conversion.
+        mBuilder.setSampleRate(44100);
+        mBuilder.setSampleRateConversionQuality(SampleRateConversionQuality::Medium);
+        ASSERT_TRUE(openStream());
+        ASSERT_LT(0, mStream->getHardwareSampleRate());
```
robertwu1, line 404: getHardware* functions are only supported in Android API 34+. Please wrap this test around an SDK level check.

PR #1899, opened 2023-08-15
OboeTester: Stop using get_nprocs()

### LINKED ISSUE #1891: get_nprocs() not supported on older versions on Android
[get_nprocs()](https://linux.die.net/man/3/get_nprocs) was supported starting in [Android 5.1](https://www.gnu.org/software/gnulib/manual/html_node/get_005fnprocs.html). We should add the appropriate checks in Oboe

robertwu1:
Fixes #1891


philburk (approved):
I tested this with the CPU LOAD test and it worked fine.

PR #1901, opened 2023-08-17
OboeTester: improve CPU LOAD test

philburk:
Show CPU migrations.
Fix clearing of CPU affinity.
Disable drawing during the benchmarking phase so it will converge.

### DIFF
--- a/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/DynamicWorkloadActivity.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/DynamicWorkloadActivity.java
@@ -17,6 +17,7 @@
 package com.mobileer.oboetester;
 
 import android.graphics.Color;
+import android.graphics.Typeface;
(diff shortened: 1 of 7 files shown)


apps/OboeTester/app/src/main/java/com/mobileer/oboetester/DynamicWorkloadActivity.java
```diff
@@ -213,6 +217,31 @@ private void stop() {
 
     }
 
+
+    /**
+     * This text will look best in a monospace font.
+     * @param cpuMask CPU core bit mask
+     * @return a text display of the selected cores like "--2-45-6-"
```
robertwu1, line 224: No dash between 5 and 6?

philburk: Good catch. Done.

robertwu1 (approved):

PR #1903, opened 2023-08-18
OboeTester: AUTO GLITCH improvements

### LINKED ISSUE #1902: OboeTester AUTO GLITCH needs various improvements
Review FIXME for setTolerance(0.3)
Display instructions for plugging in a USB device then setting Volume to around 80%.
Print peak.amplitude and sine.amplitude in report.
Warn if peak.amplitude too low.
Don't compare with other passed tests if none passed.

philburk:
* Reduce tolerance for glitches.
* Ask user to plugin loopback adapter and turn up volume. 
* Print peak.amplitude and sine.amplitude in report. Warn if peak.amplitude too low.
* Don't compare with other passed tests if none passed. 
* Fix SNR calculation. Was always 0.0.

Fixes #1902

### DIFF
--- a/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/BaseAutoGlitchActivity.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/BaseAutoGlitchActivity.java
@@ -288,14 +288,14 @@ public class BaseAutoGlitchActivity extends GlitchActivity {
                 getOutputChannel()
         );
 
-        // The test would only be worth running if we got the configuration we requested on input or output.
+        // The test will only be worth running if we got the configuration we requested on input or output.
         String skipReason = shouldTestBeSkipped();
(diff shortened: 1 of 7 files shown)


apps/OboeTester/app/src/main/java/com/mobileer/oboetester/BaseAutoGlitchActivity.java
```diff
@@ -408,7 +409,18 @@ void logAnalysis(String text) {
...
+        for (TestResult other : mTestResults) {
+            if (other.passed()) {
+                numPassed++;
+            }
+        }
+        return numPassed;
+    }
+
     protected void analyzeTestResults() {
+        if (countPassingTests() == 0) return;
```
robertwu1, line 423: Is there any indication that zero tests passed? Should we print something?

robertwu1 (approved):

PR #1913, opened 2023-09-06
OboeTester: Test downmixing with data paths test 

### LINKED ISSUE #1912: OboeTester Data Paths test should test more output channel masks
All output channel masks less that FCC_LIMIT (12) should be tested by the data paths test

robertwu1:
Currently, the data paths test only plays output channel masks explicitly declared in AudioDeviceInfo.

AudioFlinger handles downmixing for extra channels. Thus, we can actually test that downmixing works for channel masks that the device does not explicitly support.

This PR changes output configs to an arbitrary set of channel masks (mono, stereo, 2.1, 7.1.4) with an checkbox to test all supported channel masks.

Fixes #1912

### DIFF
--- a/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestDataPathsActivity.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestDataPathsActivity.java
@@ -756,7 +756,8 @@ public class TestDataPathsActivity  extends BaseAutoGlitchActivity {
                     }
                 }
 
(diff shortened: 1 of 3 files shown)


philburk (commented):
The title sounds like a bug. It should describe the change.
Also the commit message should have more than the Fixes line.
What changed in the code?

robertwu1:
> The title sounds like a bug. It should describe the change. Also the commit message should have more than the Fixes line. What changed in the code?

Updated title and description above.

philburk (approved):

PR #1922, opened 2023-09-20
Samples: Remove some fixed size constants

### LINKED ISSUE #1921: Potential buffer overflows with some hard coded constants in samples
In samples/shared/Mixer.h:

~~~
constexpr int32_t kBufferSize = 192*10;  // Temporary buffer is used for mixing
constexpr uint8_t kMaxTracks = 100;
~~~

robertwu1:
This PR fixes some issues with potential overflow of buffers.
1. Mixer buffer in Mixer.h could be too small if frame size is large. Fix here is to use std::unique_ptr<float[]> instead of a raw array. Mixer.h is used by MegaDrone, Soundboard, and RhythmGame.
2. If the number of tracks goes over 100 in Mixer.h, there will be a buffer overflow. Fix here is to use std::vector<IRenderableAudio*> instead of an std::array.
3. If the number of tracks goes over 100 in Synth.h, there will be a buffer overflow. Fix here is to use std::unique_ptr<SynthSound[]> instead of an std::array.

Fixes #1921

### DIFF
--- a/samples/shared/Mixer.h
+++ b/samples/shared/Mixer.h
@@ -31,19 +31,19 @@ class Mixer : public IRenderableAudio {
 public:
     void renderAudio(float *audioData, int32_t numFrames) {
 
-        int numBytes = numFrames * mChannelCount;
-        if (numBytes > mBufferSize) {
-            mMixingBuffer = std::make_unique<float[]>(numBytes);
-            mBufferSize = numBytes;
+        int numSamples = numFrames * mChannelCount;
+        if (numSamples > mBufferSize) {
+            mMixingBuffer = std::make_unique<float[]>(numSamples);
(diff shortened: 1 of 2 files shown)


samples/shared/Mixer.h
```diff
@@ -34,35 +31,38 @@ class Mixer : public IRenderableAudio {
 public:
     void renderAudio(float *audioData, int32_t numFrames) {
 
+        int numBytes = numFrames * mChannelCount;
```
philburk, line 34: numBytes should be "numSamples" or "numFloats"

robertwu1: Oops... Thanks! I just fixed it

philburk (approved):

PR #1924, opened 2023-09-21
OboeTester: Add checkbox for setStreamControlByAttributes

### LINKED ISSUE #1923: OboeTester: need checkbox to disable call to setStreamControlByAttributes()
This was added in #1886
Just needed in TEST OUTPUT.
Maybe fit it next to "Signal:" menu.

robertwu1:
Fixes #1923

### DIFF
--- a/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
@@ -604,7 +604,7 @@ abstract class TestAudioActivity extends Activity {
      * @param usage usage for the stream
      * @param contentType tupe of the stream
      */
-    private void setStreamControlByAttributes(int usage, int contentType) {
+    protected void setStreamControlByAttributes(int usage, int contentType) {
         AudioAttributes attributes = new AudioAttributes.Builder().setUsage(usage)
                 .setContentType(contentType).build();
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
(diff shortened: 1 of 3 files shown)


apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestOutputActivity.java
```diff
@@ -179,6 +184,13 @@ public void onChannelBoxClicked(View view) {
         mAudioOutTester.setChannelEnabled(channelIndex, checkBox.isChecked());
     }
 
+    @Override
+    protected void setStreamControlByAttributes(int usage, int contentType) {
+        if (mShouldSetStreamControlByAttributes.isChecked()) {
+            super.setStreamControlByAttributes(usage, contentType);
```
philburk, line 190: A set() method should set().
We should condition the call to the set() method.
The method that is conditional might be called "maybeSet.....".

philburk (changes requested):

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestOutputActivity.java
robertwu1: Thanks! Done!

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
```diff
@@ -604,7 +604,7 @@ public void openAudio() throws IOException {
      * @param usage usage for the stream
      * @param contentType tupe of the stream
      */
-    protected void setStreamControlByAttributes(int usage, int contentType) {
+    protected void maybeSetStreamControlByAttributes(int usage, int contentType) {
```
philburk, line 607: This is the method that sets it (assuming the API is implemented).

philburk (changes requested):
This is not what I meant.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
robertwu1: See newest iteration. I added a helper function

philburk (approved):

Issue #1925, opened 2023-09-26
StreamDataCallback onAudioReady audioData pointer SegFault


JP-Lisn:
Android version(s): API 33
Android device(s): Pixel 6 Pro
Oboe version: v1.7.0
App name used for testing: 
I wasn't able to find a suitable example to reproduce this within the oboeTest / library. 
 
**Short description**
I am trying to utilize `gtest` to create a test suite for the library im creating which utilizes oboe. I have setup a gtest fixture to set up my library and consequently oboe but upon running the tests, my first call of `StreamDataCallback::onAudioReady` the test suite application experiences a segmentation fault when trying to initialize the passed `void *audioData` pointer to 0. This is most certainly an issue in my code, but I'm at a loss as to what I've done wrong here and hoping someone can point out my error.

Any insights would be greatly appreciated! 

**Expected behavior**
That the `StreamDataCallback::onAudioReady` callback passed `audioData` is valid pointer which can be initialized to 0.  

**Actual behavior**
The pointer appears to be invalid and results in a segmentation fault. When debuging address sanitizer provides these insights: 
```
15139-15276 com.lisn...testapp com.lisn.framework_testapp      I  =================================================================
15139-15276 com.lisn...testapp com.lisn.framework_testapp      I  ==15139==ERROR: AddressSanitizer: unknown-crash on address 0xb400007c4cc87330 at pc 0x007ac3f24318 bp 0x007abd69b0e0 sp 0x007abd69a8d0
15139-15276 com.lisn...testapp com.lisn.framework_testapp      I  WRITE of size 768 at 0xb400007c4cc87330 thread T-1
15139-15139 AAudio             com.lisn.framework_testapp      D  AAudioStream_requestStart(s#1) returned 0 ---------
15139-15269 AudioStreamLegacy  com.lisn.framework_testapp      D  onAudioDeviceUpdate(deviceId = 22)
15139-15139 LisnFrameworkTests com.lisn.framework_testapp      I  ** player state change! ** (2) true
15139-15139 playFunction       com.lisn.framework_testapp      I  play
15139-15269 AudioStreamLegacy  com.lisn.framework_testapp      D  onAudioDeviceUpdate(deviceId = 22)
[...]

PR #1926, opened 2023-10-03
Added logging to deprecated functions

JP-Lisn:
As discussed in https://github.com/google/oboe/issues/1925, logging when using deprecated functions has been added. I did waffle on if these should be warnings or debug messages... I landed on warnings. I Will change if a different log level is preferred. 

FYI - Just signed the CLA.


philburk (approved):
Thanks for doing this!

philburk:
Warnings are good because use of the functions can lead to crashes.

JP-Lisn:
Whoops, let me double check this.

include/oboe/AudioStreamBuilder.h
```diff
@@ -515,6 +518,8 @@ class AudioStreamBuilder : public AudioStreamBase {
      * @return pointer to the builder so calls can be chained
      */
     AudioStreamBuilder *setCallback(AudioStreamCallback *streamCallback) {
+        LOGW("`setCallback` is deprecated, Call setDataCallback(std::shared_ptr<AudioStreamDataCallback>) and
```
philburk, line 521: Needs a " at the end of 521 and a " and space at the beginning of line 521

philburk (changes requested):
CI caught build errors. Did you get build errors when you tested this?

You need to:

#include "common/OboeDebug.h"

But that might interfere with making the pre-built.
We normally only do logging in .cpp files.

JP-Lisn:
Sorry about that, a little embarrassing 😞 I did run the build script, but I must have accidentally done so from the wrong source directory.

JP-Lisn:
Interesting, 

Both the `build_all_android.sh` and `prefab_build.sh` complete successfully when run locally.

philburk (changes requested):
Moving the OboeDebug.h file is an API change that could break existing application builds.
We will need to discuss this internally and decide how to move forward.
Please hold off on any more changes. Thanks.

JP-Lisn:
Totally understand. Initially, I thought this change was all good until I built the sample applications. Tests and build scripts all worked. The only way I could solve this without code repetition was to expose that header publicly so the linker could find it. I have run all the samples and tests and found nothing that broke thus far with this change.  

I didn't think this would cause issues with existing builds, as the header was not exposed before. I assumed only internal library references would be at issue, although I realize I might be missing something here. 

If Internally you decide on a different path, I would be happy to assist in making that happen :) 
Thanks again for all your help.

JP-Lisn:
Closing, for now replaced with: https://github.com/google/oboe/pull/1930

PR #1930, opened 2023-10-11
Log deprecated fix

JP-Lisn:
Removed deprecated function logging from all API inlined functions.


robertwu1:
Thanks for the changes!

PR #1937, opened 2023-11-21
OboeTester: Add sample rate tests to data paths test

### LINKED ISSUE #1936: OboeTester should have an option to test sample rates in the data paths test

robertwu1:
Fixes #1936

Note that these tests will be off by default.


philburk (approved):
Thanks!

PR #1940, opened 2023-11-23
OboeTester: speed up latency analysis by >100X

### LINKED ISSUE #1939: OboeTester latency analysis is too slow
See also b/261723697 | P2 | CTS Verifier latency analysis is too slow

Ideas for speedup:
* rectify and peak follow the noise bursts
* downsample
* look for positive edges with similar timing as original
* after finding approximate latency, follow up with frame-by-frame correlation on a small window

philburk:
The latncy analysis was too slow on lower end devices. This use a rough pass to find an approximate latency.

Then it does a fine grained analysis in a small window around the approximate peak to find the exact value.

Fixes #1939

### DIFF
--- a/apps/OboeTester/app/src/main/cpp/analyzer/LatencyAnalyzer.h
+++ b/apps/OboeTester/app/src/main/cpp/analyzer/LatencyAnalyzer.h
@@ -259,7 +259,6 @@ static int measureLatencyFromPulsePartial(AudioRecording &recorded,
                                                            stride);
         correlations[i] = correlation;
     }
-    
     // Find highest peak in correlation array.
     float peakCorrelation = 0.0;
     int32_t peakIndex = -1;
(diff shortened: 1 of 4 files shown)


apps/OboeTester/app/src/main/cpp/analyzer/LatencyAnalyzer.h
```diff
@@ -220,31 +224,46 @@ class AudioRecording
...
+        const int32_t index = (i * stride) + recordedOffset;
+        float correlation = calculateNormalizedCorrelation(&recorded.getData()[index],
                                                            &pulse.getData()[0],
-                                                           pulse.size());
+                                                           pulse.size(),
+                                                           stride);
         correlations[i] = correlation;
     }
-
+
```
philburk, line 262: white space

robertwu1 (approved):

apps/OboeTester/app/src/main/cpp/analyzer/LatencyAnalyzer.h
philburk:
```suggestion

```

PR #1956, opened 2024-02-05
Improve glitch tests

philburk:
* Track slowly changing sample rates
* Improve display of glitches, add cursor
* Add "Auto draw" checkbox
* Add "Force glitch" checkbox

### DIFF
--- a/apps/OboeTester/app/src/main/cpp/jni-bridge.cpp
+++ b/apps/OboeTester/app/src/main/cpp/jni-bridge.cpp
@@ -757,6 +757,12 @@ Java_com_mobileer_oboetester_GlitchActivity_getGlitchCount(JNIEnv *env,
     return engine.mActivityGlitches.getGlitchAnalyzer()->getGlitchCount();
 }
 
+JNIEXPORT jint JNICALL
+Java_com_mobileer_oboetester_GlitchActivity_getGlitchLength(JNIEnv *env,
+                                                           jobject instance) {
+    return engine.mActivityGlitches.getGlitchAnalyzer()->getGlitchLength();
+}
+
 JNIEXPORT double JNICALL
 Java_com_mobileer_oboetester_GlitchActivity_getPhase(JNIEnv *env,
                                                            jobject instance) {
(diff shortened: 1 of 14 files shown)


apps/OboeTester/app/src/main/cpp/jni-bridge.cpp
```diff
@@ -794,11 +812,11 @@ Java_com_mobileer_oboetester_TestDataPathsActivity_getMaxMagnitude(JNIEnv *env,
     return engine.mActivityDataPath.getDataPathAnalyzer()->getMaxMagnitude();
 }
 
-JNIEXPORT jdouble JNICALL
-Java_com_mobileer_oboetester_TestDataPathsActivity_getPhase(JNIEnv *env,
-                                                                          jobject instance) {
-    return engine.mActivityDataPath.getDataPathAnalyzer()->getPhaseOffset();
-}
+//JNIEXPORT jdouble JNICALL
```
philburk, line 815: Remove dead code.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/ManualGlitchActivity.java
```diff
@@ -19,27 +19,36 @@
...
+import android.util.Log;
+import android.view.View;
+import android.widget.CheckBox;
 import android.widget.SeekBar;
 import android.widget.TextView;
 
 import java.io.IOException;
 import java.util.Locale;
 
+import kotlin.random.Random;
```
philburk, line 31: Why kotlin?

philburk: removed line

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/ManualGlitchActivity.java
```diff
@@ -166,19 +176,45 @@ public void onTestBegan() {
...
+            mWaveformView.setCursorData(cursors);
+            Log.i(TAG,"onGlitchDetected: glitch, numSamples = " + numSamples);
             mWaveformView.postInvalidate();
         }
     }
-
-    private float[] getGlitchWaveform() {
-        return mWaveform;
+    @Override
+    protected void maybeDisplayWaveform() {
```
robertwu1, line 202: Should we hide the old waveform in some cases or should we leave it as is? If we leave it as is, is there any indicator that the waveform isn't the most recent?

robertwu1 (approved):

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/ManualGlitchActivity.java
philburk: I want to leave the glitch waveform on the screen so we can study it and maybe take a screenshot.

If you enable the "Auto draw" then it will update about twice per second.

> is there any indicator that the waveform isn't the most recent?

Only that the display is not changing for a long time.

Issue #1960, opened 2024-02-07
TEST DISCONNECT doesn't work for USB device


flamme:
The `TEST DISCONNECT` test will listen to headset plugin event to synchronize the test with the device plugging state. That works well for a USB headset and wired headset. However, if the plugged in device is a USB DAC, it will be recognized as a USB device and will not trigger headset plugin event. In that case, the test will stuck at waiting for syncing the plugging state.

PR #1961, opened 2024-02-08
Listen to USB device de/attach event.

flamme:
With listening to USB device de/attach event, disconnect test can well sync on the peripherals plug state.


philburk:
When fixing a bug add this magic text to the commit message. It will causes the bug to be closed automatically.

Fixes #1960

philburk (approved):
Nice! Thanks.

PR #1964, opened 2024-02-16
OboeTester: Remove sample rate conversion in test disconnect

robertwu1:
Oboe sample rate conversion should not be used in test disconnect as it defeats the purpose of setting a sample rate explicitly.

### DIFF
--- a/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestDisconnectActivity.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestDisconnectActivity.java
@@ -334,9 +334,6 @@ public class TestDisconnectActivity extends TestAudioActivity {
         requestedConfig.setPerformanceMode(perfMode);
         requestedConfig.setSharingMode(sharingMode);
         requestedConfig.setSampleRate(sampleRate);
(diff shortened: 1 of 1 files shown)


philburk (changes requested):
I believe this test was added because sample rate conversion in oboe makes use of a kind of proxy filter object. At one point the error callback was not being propagated properly from the proxy to the main application object. So disconnect was failing. This tests a feature of oboe. So we should keep it.

But if we want to check to make sure that sample rate conversion in aaudio does not break the error callbacks, then we should make another sample rate conversion test that uses the oboe sample rate conversion quality of none. Then AAudio will do the conversion.

robertwu1:
> I believe this test was added because sample rate conversion in oboe makes use of a kind of proxy filter object. At one point the error callback was not being propagated properly from the proxy to the main application object. So disconnect was failing. This tests a feature of oboe. So we should keep it.
> 
> But if we want to check to make sure that sample rate conversion in aaudio does not break the error callbacks, then we should make another sample rate conversion test that uses the oboe sample rate conversion quality of none. Then AAudio will do the conversion.

Let me know if the newest iteration looks ok to you. It does both now.

philburk (approved):

PR #1968, opened 2024-02-27
Fix isMMapEnabled()

### LINKED ISSUE #1967: isMMapEnabled() always returns false if setMMapEnabled(f) never called
Oboe version: 1.8.1 and earlier
 
**Short description**
Jiabin found an error in isMMapEnabled().
It returns a true if MMAP policy is set for MMAP on.
But if the policy is UNSPECIFIED then it returns false, even if MMAP is enabled in the system.

**Steps to reproduce**
I added logging to OboeTester and then ran this test before setMMapEnabled(f) was called.

* print isMMapSupported()
* print isMMapEnabled()

**Expected behavior**
If MMspSupported() then MMapEnabled() should also be true.

**Actual behavior**
isMMapEnabled() is false.

philburk:
If policy is Unspecified then return isMMapSupported.

Fixes #1967

### DIFF
--- a/src/aaudio/AAudioExtensions.h
+++ b/src/aaudio/AAudioExtensions.h
@@ -89,7 +89,7 @@ public:
         if (loadSymbols()) return false;
         if (mAAudio_getMMapPolicy == nullptr) return false;
         int32_t policy = mAAudio_getMMapPolicy();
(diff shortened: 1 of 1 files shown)


flamme (approved):
Thanks for the fix!

PR #1978, opened 2024-03-12
OboeTester: Fix problem with phaseJitter=0 in DataPaths

philburk:
Set phase calculation as invalid when magnitude is low. Some tests has phaseJitter=0.
Also some tests had very high phaseJitter and
should have failed. So I lower the max allowable jitter.

### DIFF
--- a/apps/OboeTester/app/src/main/cpp/analyzer/BaseSineAnalyzer.h
+++ b/apps/OboeTester/app/src/main/cpp/analyzer/BaseSineAnalyzer.h
@@ -211,8 +211,16 @@ protected:
     double  mPhaseIncrement = 0.0;
     double  mOutputPhase = 0.0;
     double  mOutputAmplitude = 0.75;
+    // This is the phase offset between the output sine wave and the recorded
+    // signal at the tuned frequency.
     // If this jumps around then we are probably just hearing noise.
+    // Noise can cause the magnitude to be high but mPhaseOffset will be pretty random.
+    // If we are tracking a sine wave then mPhaseOffset should be consistent.
     double  mPhaseOffset = 0.0;
+    // kPhaseInvalid indicates that the phase measurement cannot be used.
+    // We were seeing times when a magnitude of zero was causing atan2(s,c) to
+    // return a phase of zero, which looked valid to Java. This is a way of passing
+    // an error code back to Java as a single value to avoid race conditions.
     static constexpr double kPhaseInvalid = -999.0;
     double  mMagnitude = 0.0;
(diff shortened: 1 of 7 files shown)


apps/OboeTester/app/src/main/cpp/analyzer/BaseSineAnalyzer.h
```diff
@@ -195,7 +213,9 @@ class BaseSineAnalyzer : public LoopbackProcessor {
     double  mOutputAmplitude = 0.75;
     // If this jumps around then we are probably just hearing noise.
     double  mPhaseOffset = 0.0;
+    static constexpr double kPhaseInvalid = -999.0;
```
robertwu1, line 216: Can you add some comments on how this is used and when to use this?

apps/OboeTester/app/src/main/cpp/analyzer/BaseSineAnalyzer.h
```diff
@@ -195,7 +213,9 @@ class BaseSineAnalyzer : public LoopbackProcessor {
...
     double  mMagnitude = 0.0;
+    static constexpr double kMinValidMagnitude = 2.0 / (1 << 16);
```
robertwu1, line 218: Why do we need both kMinValidMagnitude and kPhaseInvalid?

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/WaveformView.java
```diff
@@ -78,7 +78,8 @@ protected void onSizeChanged(int w, int h, int oldw, int oldh) {
         mCurrentWidth = w;
         mCurrentHeight = h;
         mOffsetY = 0.5f * h;
```
robertwu1, line 80: Should this be "mOffsetY = 0.475f * h"?

philburk: No, because we want the scaled plot to still be centered between the top and bottom of the viewport.

apps/OboeTester/app/src/main/cpp/analyzer/BaseSineAnalyzer.h
philburk: I added a comment.

apps/OboeTester/app/src/main/cpp/analyzer/BaseSineAnalyzer.h
philburk: Done.

robertwu1 (approved):

PR #1221, opened 2021-03-12
Float samples

android-usb:

PR #1993, opened 2024-04-04
Samples: Add 24 bit and 32 bit support for parsing wav files

robertwu1:
24 bit and 32 bit support was added in #1221. I tested that they work with DrumThumper.

This PR now declares that those audio encodings are valid.

### DIFF
--- a/samples/parselib/src/main/cpp/wav/WavStreamReader.cpp
+++ b/samples/parselib/src/main/cpp/wav/WavStreamReader.cpp
@@ -54,8 +54,11 @@ int WavStreamReader::getSampleEncoding() {
             case 24:
                 return AudioEncoding::PCM_24;
 
-            default:
+            case 32:
                 return AudioEncoding::PCM_32;
+
+            default:
+                return AudioEncoding::INVALID;
         }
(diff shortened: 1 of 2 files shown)


samples/parselib/src/main/cpp/wav/WavStreamReader.cpp
```diff
@@ -52,11 +52,10 @@ int WavStreamReader::getSampleEncoding() {
                 return AudioEncoding::PCM_16;
 
             case 24:
-                // TODO - Support 24-bit WAV data
-                return AudioEncoding::INVALID; // for now
+                return AudioEncoding::PCM_24;
 
             default:
-                return AudioEncoding::INVALID;
+                return AudioEncoding::PCM_32;
```
philburk, line 58: Could add a "case 32" for PCM_32.
If mSampleSize is something weird, like 64, then we could return INVALID for default case.

robertwu1: Thanks! Done!

philburk (approved):

android-usb (approved):

PR #2020, opened 2024-05-10
OboeTester: add missing super call

philburk:


apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
```diff
@@ -202,6 +202,7 @@ protected void onCreate(Bundle savedInstanceState) {
 
     @Override
```
robertwu1, line 203: Should we do the same for MainActivity.java?

robertwu1: I'm pushing these changes to test this in google3.

robertwu1 (approved):

PR #2031, opened 2024-06-10
oboe: log warning if start times out

philburk:
This would have helped us catch a service bug sooner.


robertwu1 (approved):
Can log the timeout as well. LGTM!

PR #2036, opened 2024-06-12
Use a shared_ptr when creating a child stream

### LINKED ISSUE #2035: Possible crash on disconnect with streams using Oboe sample rate conversion
If a stream was opened using setSampleRate() and the setSampleRateConversionQuality() was not not None,
then if a headset was plugged in or unplugged,
and another thread or the user closed the stream,
then it was possible to get a crash in oboe_aaudio_error_thread_proc().

Note that this bug could also occur if any of these other conversion functions were enabled:

    builder.setChannelConversionAllowed(true);
    builder.setFormatConversionAllowed(true);

To avoid this bug, use Oboe v1.9.0 or later. (not yet released, as of 2024-06-11)

Note that the bug was due to not using shared_ptr when calling builder->openStream().
Because it is so dangerous, we are disabling the openStream(AudioStream **streamPP)
method. It was deprecated a long time ago. Now if you call it you will get an error and no stream.

The root cause was because the AudioStreamBuilder.openStream() was not using a shared_ptr
when creating a child stream. [buggy code](https://github.com/google/oboe/blob/da4de45cf07986aa270cd0757f9687c983e63da9/src/common/AudioStreamBuilder.cpp#L119)

Using a shared_ptr is necessary to prevent an error callback from using a raw pointer and referencing a deleted stream.
It also prevents the stream from being deleted while a data callback is active.

philburk:
This is used when Oboe does sample rate conversion. Or format or channel count conversion.

It used to use a raw pointer, which could cause crashes.

Fixes #2035

### DIFF
--- a/include/oboe/AudioStream.h
+++ b/include/oboe/AudioStream.h
@@ -54,7 +54,7 @@ public:
      */
     explicit AudioStream(const AudioStreamBuilder &builder);
 
-    virtual ~AudioStream() = default;
+    virtual ~AudioStream();
 
     /**
      * Open a stream based on the current settings.
--- a/include/oboe/AudioStreamBuilder.h
+++ b/include/oboe/AudioStreamBuilder.h
@@ -609,14 +609,13 @@ public:
     }
 
     /**
-     * Create and open a stream object based on the current settings.
-     *
-     * The caller owns the pointer to the AudioStream object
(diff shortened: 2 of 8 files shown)


philburk:
Closing. Use #2037 instead.

PR #2048, opened 2024-06-26
OboeTester: Change action bar to light

robertwu1:
OboeTester had always used a light action bar until #2019. That PR changed the action bar to a dark theme. This PR changes action bars back to light.


philburk (changes requested):
We should stick to portrait because the tests do not handle rotation properly.

philburk (approved):

Issue #2062, opened 2024-07-09
Glitch in background, exclusive mode


simdax:
Hello.

I'm working on a plugin that should be open when my application runs in background.

I use a fullduplex stream, as i filter the input from the default microphone. The idea is to keep the app alive while recording the surrounding sound.

For now, all is ok when I run my plugin from the app. But when going in background the stream becomes glitchy.

I'm not sure also to really be able to be able to start a stream with exclusive mode. Do I need to do something special?

Thank you!

robertwu1:
Try the LiveEffect sample. It currently does not run in the background so I created a Pull Request (#2063). It seems to work fine in the background with the minor change with that PR.

simdax:
As I mentionned it here https://github.com/google/oboe/pull/2063#issuecomment-2218522159
It seems that passing in background make some few glitches.

philburk:
If your callback function does a lot of processing then there may be a problem running in the background.
That is because the CPU gives lower priority to background tasks and lowers the CPU frequency.

What device are you on?
OS version?
etcetera

robertwu1:
See https://developer.android.com/develop/background-work/services/foreground-services. Starting for Android U, foreground services are required.

simdax:
It would be nice to have a background example, with the foreground service.
I'm not sure how to keep enough CPU for my plugin processing

robertwu1:
I filed issue #2065

PR #2063, opened 2024-07-09
LiveEffect: Run in background

robertwu1:
Since duplex streams are often used in the background, we should let LiveEffect run in the background.

Some users have trouble with FullDuplexStream in the background (#2062) so we should have a sample to point them to.


simdax:
yes it works indeed

simdax:
To be honnest after more testing some clicks may appear especially if I shutdown the screen

And at the end, the app can just stop itself after some time.

I think we would need something to keep the process really at the front somehow.

Issue #2059, opened 2024-07-08
OpenSL can hang if callback returns DataCallbackResult::Stop


philburk:
Oboe version: 1.9.0 or earlier
App name used for testing: OboeTester
 
**Short description**
The callback handler called requestStop() which tried to acquire mLock.
But the close already held mLock and was waiting for the callback to finish.
So there was a deadlock.

This would normally be a rare event. But the RAPID CYCLE test uses usleep()s and 
specially timed return of DataCallbackResult::Stop to trigger the bug.

**Steps to reproduce**
Run Oboetester
Press EXTRAS...
Press RAPID CYCLE
Press START

**Expected behavior**
Run forever.

**Actual behavior**
Stops after a few cycles.

philburk:
This is tracked by internal bug b/348615156

PR #2064, opened 2024-07-10
Test whether a stream stop can deadlock

philburk:
Try to reproduce the bug in #2059.

### DIFF
--- a/src/common/AudioStreamBuilder.cpp
+++ b/src/common/AudioStreamBuilder.cpp
@@ -94,19 +94,16 @@ Result AudioStreamBuilder::openStream(AudioStream **streamPP) {
 }
 
 Result AudioStreamBuilder::openStreamInternal(AudioStream **streamPP) {
-    static bool versionLogged = false;
-
     auto result = isValidConfig();
     if (result != Result::OK) {
         LOGW("%s() invalid config. Error %s", __func__, oboe::convertToText(result));
         return result;
     }
 
-    if (!versionLogged) {
-        LOGI("%s() %s -------- %s --------",
-             __func__, getDirection() == Direction::Input ? "INPUT" : "OUTPUT", getVersionText());
-         versionLogged = true;
-    }
+#ifndef OBOE_SUPPRESS_LOG_SPAM
+    LOGI("%s() %s -------- %s --------",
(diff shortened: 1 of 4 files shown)


src/common/AudioStreamBuilder.cpp
```diff
@@ -94,14 +94,19 @@ Result AudioStreamBuilder::openStream(AudioStream **streamPP) {
 }
 
 Result AudioStreamBuilder::openStreamInternal(AudioStream **streamPP) {
+    static bool versionLogged = false;
```
robertwu1, line 97: Do we need this? I'm not too big of a fan of static variables like this and openStream() shouldn't realistically be called that often. Also, this is a LOGI and not something like LOGE

src/opensles/AudioStreamOpenSLES.cpp
```diff
@@ -66,9 +66,6 @@ SLuint32 AudioStreamOpenSLES::getDefaultByteOrder() {
 }
 
 Result AudioStreamOpenSLES::open() {
-
-    LOGI("AudioStreamOpenSLES::open() chans=%d, rate=%d", mChannelCount, mSampleRate);
```
robertwu1, line 70: I really having a log where the parameters are logged somewhere. If this is not logged anywhere else, I prefer leaving this here.

src/common/AudioStreamBuilder.cpp
philburk: When I run the unit tests it tries to trigger a deadlock by opening and closing rapidly many times. These version logs spew into the test output and can be very spammy. Maybe I could add a compiler flag that disables this log. Then we could just disable it for the tests. That seems cleaner.  Something like:

    #ifndef OBOE_DISABLE_VERSION_LOG
        LOGI(blah blah);
    #endif

src/opensles/AudioStreamOpenSLES.cpp
philburk: This only logs 2 out of a dozen parameters. Maybe we should log more.

This spews into the logs when I am testing.  Maybe:

    #ifndef OBOE_SUPPRESS_LOG_SPAM

src/common/AudioStreamBuilder.cpp
philburk: Done

robertwu1 (approved):

PR #2070, opened 2024-07-11
LiveEffect: Add foreground service

### LINKED ISSUE #2062: Glitch in background, exclusive mode

robertwu1:
Fixes #2062

### DIFF
--- a/samples/LiveEffect/src/main/java/com/google/oboe/samples/liveEffect/DuplexStreamForegroundService.java
+++ b/samples/LiveEffect/src/main/java/com/google/oboe/samples/liveEffect/DuplexStreamForegroundService.java
@@ -28,20 +28,6 @@ public class DuplexStreamForegroundService extends Service {
     public static final String ACTION_START = "ACTION_START";
     public static final String ACTION_STOP = "ACTION_STOP";
 
-    public void startForegroundService1() {
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
-            startForeground(1, buildNotification(),
-                                ServiceInfo.FOREGROUND_SERVICE_TYPE_MEDIA_PLAYBACK
-                                        | ServiceInfo.FOREGROUND_SERVICE_TYPE_MICROPHONE);
-        }
-    }
-
-    public void stopForegroundService() {
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
-            stopForeground(STOP_FOREGROUND_REMOVE);
-        }
-    }
-
     @Override
     public IBinder onBind(Intent intent) {
         // We don't provide binding, so return null
--- a/samples/LiveEffect/src/main/java/com/google/oboe/samples/liveEffect/DuplexStreamForegroundService.java
+++ b/samples/LiveEffect/src/main/java/com/google/oboe/samples/liveEffect/DuplexStreamForegroundService.java
@@ -1,0 +1,90 @@
+package com.google.oboe.samples.liveEffect;
+
+import android.app.ForegroundServiceStartNotAllowedException;
+import android.app.Notification;
+import android.app.NotificationChannel;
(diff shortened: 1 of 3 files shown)


samples/LiveEffect/src/main/java/com/google/oboe/samples/liveEffect/DuplexStreamForegroundService.java
```diff
@@ -0,0 +1,76 @@
...
+            case ACTION_START:
+                Log.i(TAG, "Receive ACTION_START" + intent.getExtras());
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
+                    startForeground(1, buildNotification(),
+                            ServiceInfo.FOREGROUND_SERVICE_TYPE_MEDIA_PLAYBACK
+                                    | ServiceInfo.FOREGROUND_SERVICE_TYPE_MICROPHONE);
+                }
+                break;
+            case ACTION_STOP:
+                Log.i(TAG, "Receive STOP_RECORD" + intent.getExtras());
```
philburk: Should STOP_RECORD be ACTION_STOP?

samples/LiveEffect/src/main/java/com/google/oboe/samples/liveEffect/DuplexStreamForegroundService.java
```diff
@@ -0,0 +1,76 @@
...
+        switch (intent.getAction()) {
+            case ACTION_START:
+                Log.i(TAG, "Receive ACTION_START" + intent.getExtras());
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
+                    startForeground(1, buildNotification(),
+                            ServiceInfo.FOREGROUND_SERVICE_TYPE_MEDIA_PLAYBACK
+                                    | ServiceInfo.FOREGROUND_SERVICE_TYPE_MICROPHONE);
+                }
+                break;
+            case ACTION_STOP:
```
philburk, line 83: It's funny that the ACTION_STOP is handled by onStartCommand().

samples/LiveEffect/src/main/java/com/google/oboe/samples/liveEffect/MainActivity.java
```diff
@@ -163,6 +172,12 @@ protected void onPause() {
     @Override
     protected void onDestroy() {
         onStopTest();
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
+            Intent serviceIntent = new Intent(ACTION_START, null, this,
```
philburk: Should this be ACTION_STOP?

philburk (commented):
Does the service stop while it is in the background?
Can the audio keep running?

samples/LiveEffect/src/main/java/com/google/oboe/samples/liveEffect/MainActivity.java
robertwu1: Fixed in a newer patch

samples/LiveEffect/src/main/java/com/google/oboe/samples/liveEffect/DuplexStreamForegroundService.java
robertwu1: Yea, it's odd but onStartCommand() handles all commands like how a command queue would work.

philburk (approved):

PR #2071, opened 2024-07-11
OboeTester: Create foreground service

### LINKED ISSUE #2065: Add a foreground service to OboeTester
Starting in Android U, [foreground services](https://developer.android.com/develop/background-work/services/foreground-services) are enforced for Microphone access when the app runs in the background.

We should add a foreground service to OboeTester so the audio can run in the background.

robertwu1:
Fixes #2065

### DIFF
--- a/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/AudioForegroundService.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/AudioForegroundService.java
@@ -48,7 +48,7 @@ public class AudioForegroundService extends Service {
                     NotificationManager.IMPORTANCE_NONE));
 
             return new Notification.Builder(this, "all")
-                    .setContentTitle("Playing/recoding audio")
+                    .setContentTitle("Playing/recording audio")
                     .setContentText("playing/recording...")
                     .setSmallIcon(R.drawable.ic_notification)
                     .build();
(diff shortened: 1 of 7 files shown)


atneya (approved):
lgtm

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/MainActivity.java
```diff
@@ -16,6 +16,8 @@
 
 package com.mobileer.oboetester;
 
+import static com.mobileer.oboetester.AudioForegroundService.ACTION_START;
```
philburk: Are these used?

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/AudioForegroundService.java
```diff
@@ -0,0 +1,80 @@
...
+import android.app.Service;
+import android.content.Context;
+import android.content.Intent;
+import android.content.pm.ServiceInfo;
+import android.os.Build;
+import android.os.IBinder;
+import android.util.Log;
+
+public class AudioForegroundService extends Service {
+    private static final String TAG = "AudioForegroundService";
```
philburk, line 31: I have been using OboeTester as the tag so I can enable all OboeTester logs.

philburk (changes requested):
For testing and debugging, we need to retain the ability to test what happens if you do NOT start a ForegroundService. So please add another checkbox on MainActivity screen below "enable background operation" labelled "use foregroundService".

philburk (approved):
What notification does the user see?

robertwu1:
Screenshot_20240712-110941~2
Screenshot_20240712-110951

PR #2074, opened 2024-07-13
OboeTester Data Paths skip remaining if unplugged

philburk:
If a device is unplugged then there is no point in running any remaining tests.

### DIFF
--- a/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestDataPathsActivity.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestDataPathsActivity.java
@@ -143,6 +143,8 @@ public class TestDataPathsActivity  extends BaseAutoGlitchActivity {
     private double mPhaseErrorSum;
     private double mPhaseErrorCount;
 
+    private boolean skipRemainingTests;
+
     private CheckBox mCheckBoxInputPresets;
     private CheckBox mCheckBoxAllChannels;
     private CheckBox mCheckBoxInputChannelMasks;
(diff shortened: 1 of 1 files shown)


apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestDataPathsActivity.java
```diff
@@ -403,11 +405,13 @@ protected String whyShouldTestBeSkipped() {
...
+            skipRemainingTests = true; // the device must have been unplugged
         }
         if (requestedOutConfig.getDeviceId() != 0
                 && (requestedOutConfig.getDeviceId() != actualOutConfig.getDeviceId())) {
             why += ", outDev(" + requestedOutConfig.getDeviceId()
                     + "!=" + actualOutConfig.getDeviceId() + "),";
+            skipRemainingTests = true; // the device must have been unplugged
         }
         if ((requestedInConfig.getInputPreset() != actualInConfig.getInputPreset())) {
             why += ", inPre(" + requestedInConfig.getInputPreset()
```
robertwu1, line 417: When would the requested input preset be different than the actual input preset? Should we add a similar check here in another PR?

philburk: The QuirksManager may substitute the InputPreset if it is broken on a device.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestDataPathsActivity.java
```diff
@@ -143,6 +143,8 @@ public class TestDataPathsActivity  extends BaseAutoGlitchActivity {
     private double mPhaseErrorSum;
     private double mPhaseErrorCount;
 
+    private boolean skipRemainingTests;
```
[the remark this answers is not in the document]
philburk: I think of it as an imperative command, not a "should".
I can't think of another data type that fits the name. It could only be a boolean.
I will add the m prefix.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestDataPathsActivity.java
philburk: Also VoicePerformance will get coerced to VoiceRecognition on older devices.
The key idea here is that if we are trying to test one Preset and got a different one then we should skip the test.

> Should we add a similar check here in another PR?

What should we add a check for?

robertwu1: Thanks for the explanation. I don't think we need another check.

robertwu1 (approved):

PR #2079, opened 2024-07-18
Unit Tests: Test no sample rate conversion in TestStreamFrameProcessed

robertwu1:
It seems that we were missing some tests where no sample rate conversion is used for sample rates like 8000. This PR adds another param to TestStreamFrameProcessed to help test this case.

### DIFF
--- a/tests/testStreamFramesProcessed.cpp
+++ b/tests/testStreamFramesProcessed.cpp
@@ -56,7 +56,7 @@ TEST_P(StreamFramesProcessed, VerifyFramesProcessed) {
     mBuilder.setDirection(direction)
             ->setFormat(AudioFormat::I16)
             ->setSampleRate(sampleRate)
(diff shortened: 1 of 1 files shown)


tests/testStreamFramesProcessed.cpp
```diff
@@ -51,12 +51,16 @@ void StreamFramesProcessed::TearDown() {
 TEST_P(StreamFramesProcessed, VerifyFramesProcessed) {
     const Direction direction = std::get<0>(GetParam());
     const int32_t sampleRate = std::get<1>(GetParam());
+    const bool useSampleRateConversion = std::get<2>(GetParam());
```
philburk, line 54: useOboeSampleRateConversion

Sample Rate Conversion will happen somewhere. The question is whether it is in Oboe or lower down.

philburk (changes requested):

philburk (approved):

PR #2083, opened 2024-08-05
OboeTester Data Paths handle mix of USB

philburk:
Handle any combination of USB Headset or Device.


robertwu1 (commented):
See [findCompatibleInputDevice](https://github.com/google/oboe/blob/8ebc25217ea855ee2f3e87bc70307c7939bf3ddc/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/BaseAutoGlitchActivity.java#L485).

If there is a usb device AND a USB headset exposed, should we prefer the usb headset  input for the usb headset output and the usb device output for the usb headset input? 

Are there any USB device that we know of that exposes both a USB device and a USB headset endpoint?

philburk:
Interesting idea. I changed it to supported preferred order.

> Are there any USB device that we know of that exposes both a USB device and a USB headset endpoint?

The Rode NT USB Mini is the only one I know of.
This code was written to fix:
[b/356818652](http://b/356818652) | P2 | Oboe tester failed to execute data paths tests with Rode NT USB Mini

PR #2089, opened 2024-08-15
WIP: Log device id changes in OboeTester

robertwu1:


philburk (commented):
This looks like a lot of refactoring that may be a great change.
How does it all relate to the logging of device changes.

Issue #2090, opened 2024-08-16
OboeTester needs a way to measure latency when AEC is active


philburk:
When measuring latency with VoiceComm active the device often adds an AEC.
This will cause some or all of the loopback signal to be cancelled.
In some cases a brief chirp can get through before the AEC adaptive filter converges.

![aec_latency_chirp](https://github.com/user-attachments/assets/e0cb1406-5ed7-4b28-ab6d-da67d002cd05)

So we should add an option to use alternative signals in place of the pulse train including
* brief chirp
* short white noise pulse
* short FM noise burst using a C/M ratio of sqrt(2)

philburk:
It would also be nice to display waveforms of the output and input recordings.

And we should show the measured latency even if the confidence is very low.

philburk:
Similar to #1291

philburk:
I found that I can defeat the AEC just by enabling the Session checkbox and leaving the AEC checkbox off.

JimG777:
> I found that I can defeat the AEC just by enabling the Session checkbox and leaving the AEC checkbox off.

That seems to be the case on many devices that I have tested with, but not all:
- Samsung S22 (with Exynos chipset) w/ Android 14
  - Session checkbox disabled: AEC is defeated (so no AEC here!)
  - Session checkbox enabled, AEC checked or unchecked: AEC applied
- Samsung S22 (with Qualcomm chipset) w/ Android 14
  - AEC can't be defeated
- Samsung S21 w/ Android 13
  - No speaker: AEC can't be defeated
  - Using Speaker: Session checkbox enabled, AEC unchecked = AEC is defeated
- Samsung S10 (with Exynos chipset) w/ Android 12
  - AEC can't be defeated

Other devices varied. Most of my Pixel devices had the same behavior as yours (except the Pixel 1).

Why is there so much variation here?

Also, what is the expected behavior? I thought that if using the VoiceComm preset, the "/vendor/etc/audio_effects.xml" would get applied automatically. That generally seems to be the case, but why does providing the SessionId to Oboe override that?

_My tests use these OboeTester settings for input:_
- _Callback size: 480_
- _Enable Oboe workarounds_
- _Mode: IN_COMMUNICATION_
- _48000/1/PCM_I16/LOW_LATENCY/VoiceComm_

PR #2091, opened 2024-08-19
OboeTester: show default AEC status

philburk:
Also add NoiseSuppressor

For #2090

### DIFF
--- a/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/StreamConfigurationView.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/StreamConfigurationView.java
@@ -590,6 +590,7 @@ public class StreamConfigurationView extends LinearLayout {
                     Log.e(TAG, String.format(Locale.getDefault(), "Could not create AcousticEchoCanceler"));
                 }
             }
+
             // If AGC is not available, the checkbox will be disabled in initializeViews().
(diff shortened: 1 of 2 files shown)


apps/OboeTester/app/src/main/java/com/mobileer/oboetester/StreamConfigurationView.java
```diff
@@ -575,11 +594,26 @@ public void setupEffects(int sessionId) {
...
                 if (mAutomaticGainControl != null) {
+                    boolean wasOn = mAutomaticGainControl.getEnabled();
+                    String text = mAutomaticGainControlText + "(" + (wasOn ? "Y" : "N") + ")";
+                    mAutomaticGainControlCheckBox.setText(text);
                     mAutomaticGainControl.setEnabled(mAutomaticGainControlCheckBox.isChecked());
                 } else {
                     Log.e(TAG, String.format(Locale.getDefault(), "Could not create AutomaticGainControl"));
                 }
             }
+            // If AEC is not available, the checkbox will be disabled in initializeViews().
```
robertwu1, line 605: should this be noise suppressor?

philburk: Yes! Good catch.

PR #2097, opened 2024-09-04
LiveEffect: Request permissions before starting foreground service

### LINKED ISSUE #2095: LiveEffect not opening and has glitched audio on Android 14 but runs fine on Android 13.
**Issues:**
(1)The app doesn't open at first because "FOREGROUND_SERVICE_MICROPHONE" has "RECORD_AUDIO" as runtime prerequisites in Android 14. The app runs by requesting audio recording permission before running foreground services.

(2) Using AAudio as API, the audio doesn't work as intended and becomes noisy and has a constant sound every time it runs. This doesn't occur while using OpenSL ES as API.  Can this be a device specific issue?

I tested it on Android 13, the app has no issue opening as well as running. 

**Devices:**
Android version: 14
Android device: Motorola g14

Android version: 13
Android device: Motorola g52

robertwu1:
Fixes #2095

### DIFF
--- a/samples/LiveEffect/src/main/java/com/google/oboe/samples/liveEffect/MainActivity.java
+++ b/samples/LiveEffect/src/main/java/com/google/oboe/samples/liveEffect/MainActivity.java
@@ -130,6 +130,10 @@ public class MainActivity extends Activity
         LiveEffectEngine.setDefaultStreamValues(this);
         setVolumeControlStream(AudioManager.STREAM_MUSIC);
 
+        if (!isRecordPermissionGranted()){
+            requestRecordPermission();
+        }
+
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
             Intent serviceIntent = new Intent(ACTION_START, null, this,
                     DuplexStreamForegroundService.class);
--- a/samples/LiveEffect/src/main/java/com/google/oboe/samples/liveEffect/MainActivity.java
+++ b/samples/LiveEffect/src/main/java/com/google/oboe/samples/liveEffect/MainActivity.java
@@ -132,6 +132,8 @@ public class MainActivity extends Activity
 
         if (!isRecordPermissionGranted()){
             requestRecordPermission();
+        } else {
+            startForegroundService();
         }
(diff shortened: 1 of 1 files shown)


samples/LiveEffect/src/main/java/com/google/oboe/samples/liveEffect/MainActivity.java
```diff
@@ -130,6 +130,10 @@ public void onClick(View v) {
         LiveEffectEngine.setDefaultStreamValues(this);
         setVolumeControlStream(AudioManager.STREAM_MUSIC);
 
+        if (!isRecordPermissionGranted()){
+            requestRecordPermission();
```
philburk, line 134: Does this block?
If YES then it may cause an ANR.
If NO then you may need code to start the foreground service and test in onRequestPermissionResult().

robertwu1: Thanks for bringing this up! I fixed this in the recent commit.

samples/LiveEffect/src/main/java/com/google/oboe/samples/liveEffect/MainActivity.java
```diff
@@ -130,10 +130,8 @@ public void onClick(View v) {
         LiveEffectEngine.setDefaultStreamValues(this);
         setVolumeControlStream(AudioManager.STREAM_MUSIC);
 
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
-            Intent serviceIntent = new Intent(ACTION_START, null, this,
-                    DuplexStreamForegroundService.class);
-            startForegroundService(serviceIntent);
+        if (!isRecordPermissionGranted()){
+            requestRecordPermission();
```
philburk, line 134: If we already have record permission from an earlier run, then will onRequestPermissionsResult() get called?
If not then we won't get a foreground service the second time we run the app.
We may want to add:

    } else { /* start FGS */ }

philburk (approved):

PR #2107, opened 2024-10-05
OboeTester: measure latency with varying buffer sizes to detect MMAP errors

philburk:
This can be used to detect errors in the DSP MMAP position timestamps
used by the AAudio IsochronousClockModel.

If the DSP position is off then it can cause a discontinuity in round-trip latency as a function of buffer size.
So we change the buffer size and look for the discontinuity.

### DIFF
--- a/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/RoundTripLatencyActivity.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/RoundTripLatencyActivity.java
@@ -227,6 +227,7 @@ public class RoundTripLatencyActivity extends AnalyzerActivity {
         private int mState = STATE_MEASURE_LOW;
         private int mFramesPerBurst;
 
+        public static final int RESULT_CONTINUE= 1;
         public static final int RESULT_OK = 0;
         public static final int RESULT_DISCONTINUITY = -1; // DSP is reading from the wrong place.
         public static final int RESULT_ERROR = -2; // Could not measure latency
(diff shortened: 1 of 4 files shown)


apps/OboeTester/app/src/main/java/com/mobileer/oboetester/RoundTripLatencyActivity.java
```diff
@@ -165,30 +201,287 @@ public void start() {
...
+                                mLowBufferLatency,
+                                mMiddleBufferBursts,
+                                mMiddleBufferLatency);
+                        double deviationHigh = measureLatencyLinearity(
+                                mMiddleBufferBursts,
+                                mMiddleBufferLatency,
+                                mHighBufferBursts,
+                                mHighBufferLatency);
+                        if (deviationLow >= deviationHigh) {
+                            // bottom half was bad so subdivide it
```
robertwu1, line 310: Should we continue to subdivide this for a couple more iterations to hone in where the problem is at?

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/RoundTripLatencyActivity.java
```diff
@@ -165,30 +201,287 @@ public void start() {
...
+                                mLowBufferBursts,
+                                mLowBufferLatency,
+                                mMiddleBufferBursts,
+                                mMiddleBufferLatency);
+                        double deviationHigh = measureLatencyLinearity(
+                                mMiddleBufferBursts,
+                                mMiddleBufferLatency,
+                                mHighBufferBursts,
+                                mHighBufferLatency);
+                        if (deviationLow >= deviationHigh) {
```
robertwu1, line 309: What about the case where both sides are bad? Do we have any indication of this?

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/RoundTripLatencyActivity.java
robertwu1: It seems like nextBursts is not defined for the STATE_MEASURE_MIDDLE branch

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/RoundTripLatencyActivity.java
```diff
@@ -165,30 +201,287 @@ public void start() {
...
+                if (nextBursts < 0) {
+                    if ((mHighBufferBursts - mLowBufferBursts) <= 1) {
+                        result = RESULT_DISCONTINUITY;
+                        mMessage = "ERROR - DSP position error between "
+                                + mLowBufferBursts + " and "
+                                + mHighBufferBursts + " bursts!";
+                    } else {
+                        // Subdivide the remaining search space.
+                        mMiddleBufferBursts = (mHighBufferBursts + mLowBufferBursts) / 2;
+                        result = mMiddleBufferBursts;
```
robertwu1, line 336: I'm a bit confused about what we are returning here. Can you add some comments/rename some variables?

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/RoundTripLatencyActivity.java
robertwu1: Ohhh, it seems like ScanLatencyTestRunner is the one actually subdividing this.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/RoundTripLatencyActivity.java
```diff
@@ -165,30 +201,287 @@ public void start() {
...
+                                mOutputBufferCapacityInBursts,
+                                mOutputIsMMapExclusive);
+                        mBufferBursts = result;
+
+                        break;
+                    case STATE_SCANNING_INPUT:
+                        inputFinder.setFramesPerBurst(mInputFramesPerBurst);
+                        result = inputFinder.onAnalyserDone(getMeasuredLatency(),
+                                getMeasuredConfidence(),
+                                mInputMarginBursts,
```
robertwu1, line 415: For my understanding, why does input use input margin instead of buffer bursts? What is the difference?

robertwu1 (approved):

PR #2116, opened 2024-11-15
Oboe: add reportWorkload()

philburk:
Give ADPF advance warning when the application workload changes. This can reduce glitching when the workload increases suddenly. It works by estimating what the callback duration will be and then reporting that to ADPF as if it had already happened.

### DIFF
--- a/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/DynamicWorkloadActivity.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/DynamicWorkloadActivity.java
@@ -53,6 +53,8 @@ public class DynamicWorkloadActivity extends TestOutputActivityBase {
 
     public static final String KEY_USE_ADPF = "use_adpf";
     public static final boolean VALUE_DEFAULT_USE_ADPF = false;
+    public static final String KEY_USE_WORKLOAD = "use_workload";
+    public static final boolean VALUE_DEFAULT_USE_WORKLOAD = false;
     public static final String KEY_SCROLL_GRAPHICS = "scroll_graphics";
     public static final boolean VALUE_DEFAULT_SCROLL_GRAPHICS = false;
 
@@ -73,6 +75,7 @@ public class DynamicWorkloadActivity extends TestOutputActivityBase {
     private CheckBox mDrawAlwaysBox;
     private int mCpuCount;
     private boolean mShouldUseADPF;
+    private boolean mShouldUseWorkloadReporting;
 
     private static final int WORKLOAD_LOW = 1;
     private int mWorkloadHigh; // this will get set later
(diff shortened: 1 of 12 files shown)


include/oboe/AudioStream.h
```diff
@@ -544,6 +542,27 @@ class AudioStream : public AudioStreamBase {
         return mPerformanceHintEnabled;
     }
 
+    /**
+     * If you have called setPerformanceHintEnabled(true) then you can give
+     * the performance manager more information about your workload.
+     * You can call this at the beginning of the callback when you figure
+     * out what your workload will be.
+     *
+     * The workload can be in your own units. If you are synthesizing music
```
robertwu1, line 551: Can you give a suggested range or some best practices? What happens if they use 1000 for each voice?

src/common/AdpfWrapper.cpp
```diff
@@ -120,5 +120,22 @@ void AdpfWrapper::onEndCallback(double durationScaler) {
...
+        if (appWorkload > mPreviousWorkload
+                && mPreviousWorkload > 0
+                && mPreviousDuration > 0) {
+            int64_t predictedDuration = appWorkload * mPreviousDuration / mPreviousWorkload;
+            reportActualDuration(predictedDuration);
+
+        }
+        mPreviousWorkload = appWorkload;
+    }
+
```
robertwu1, line 140: nit: extra space

src/common/AdpfWrapper.cpp
```diff
@@ -120,5 +120,22 @@ void AdpfWrapper::onEndCallback(double durationScaler) {
...
+void AdpfWrapper::reportWorkload(int32_t appWorkload) {
+    if (isOpen()) {
+        // Compare with previous workload. If we think we will need more
+        // time to render the callback then warn ADPF as early as possible.
+        if (appWorkload > mPreviousWorkload
+                && mPreviousWorkload > 0
+                && mPreviousDuration > 0) {
+            int64_t predictedDuration = appWorkload * mPreviousDuration / mPreviousWorkload;
+            reportActualDuration(predictedDuration);
```
robertwu1, line 135: Could there be a race condition between the reportActualDuration in this function and the one in onEndCallback()?

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/DynamicWorkloadActivity.java
```diff
@@ -407,6 +425,8 @@ public void startTestUsingBundle() {
             // Specific options.
             mShouldUseADPF = mBundleFromIntent.getBoolean(KEY_USE_ADPF,
                     VALUE_DEFAULT_USE_ADPF);
+            mShouldUseWorkloadReporting = mBundleFromIntent.getBoolean(KEY_USE_WORKLOAD,
```
robertwu1, line 428: Please update https://github.com/google/oboe/blob/main/apps/OboeTester/docs/AutomatedTesting.md with the new parameter.

include/oboe/AudioStream.h
```diff
@@ -544,6 +542,27 @@ class AudioStream : public AudioStreamBase {
...
+     * then the workload could be the number of active voices.
+     * If your app is a game then it could be the number of sound effects.
+     *
+     * The performance hint code can use this as an advance warning that the callback duration
```
robertwu1, line 555: How many different sources of truth will we have for the performance hint after this CL? What if multiple are actively used?

How would apps know that they are doing the right thing when they call this function? Is there some easy graphs or traces we can point them to?

src/common/AdpfWrapper.cpp
philburk: There is no race. ADPF will just have more information to work with. At the end of the callback, ADPF will get the actual duration, which may be affected by the CPU boost due to the last call.

include/oboe/AudioStream.h
philburk: ADPF works on a per-thread basis. Then the scheduler balances the requests from all the threads.

It is a bit fuzzy and indirect. If the glitches stop then it is probably working. They could also look at Perfetto traces.

robertwu1 (approved):

PR #2136, opened 2025-01-12
Update DuplexCallback.h in fxlab

dipak140:
In cases where the stream does not start immediately when starting the app, the audio stream is stopped immediately, which means the demo does not work. This if statement is unnecessary since any errors will be handled by ErrorCallback.

### DIFF
--- a/apps/fxlab/app/src/main/cpp/DuplexCallback.h
+++ b/apps/fxlab/app/src/main/cpp/DuplexCallback.h
@@ -43,10 +43,6 @@ public:
         std::fill(outputData, outputData + numFrames * outputChannelCount, 0);
         oboe::ResultWithValue<int32_t> result = inRef.read(inputBuffer.get(), numFrames, 0);
         int32_t framesRead = result.value();
-        if (!result) {
-            inRef.requestStop();
-            return oboe::DataCallbackResult::Stop;
-        }
         if (mSpinUpCallbacks > 0 && framesRead > 0) {
             mSpinUpCallbacks--;
             return oboe::DataCallbackResult::Continue;


philburk:
@dipak140  - Is the read() returning an error?!
What is the error code?
What device is this happening on?

Maybe we could start with a bug report. The fix may be needed in Oboe.

dipak140:
@philburk There was no error, the read just returned null at the beginning, which meant the stream was requested to close immediately, and the data callback stopped. This behaviour is unfortunately not replicable easily, and even i didn't catch it at first since it worked fine on my device (Samsung A12). It did not work on Realme, which was caught later on. 

Secondly, the if statement should be added before the read, if i am not wrong? 

I am also skeptical about the need of stopping the stream altogether if the read is unsuccessful, since the start request can take time and the onAudioReady callback would be called regardless. Please do correct me if this assumption is wrong. Will raise a bug report regardless. 

Additionally, the app uses managed stream which is deprecated from what i checked, so would you be open to a Patch to replace it with audioStream?

robertwu1:
The issue is that the ErrorCallback only is registered for the output stream currently in the sample. If remove the line of code in this PR, if the input fails, there will be no clues.

There is a [line](https://github.com/google/oboe/blob/b6ec9fdc959b5ac115bf92a2232393f5e1794a6e/apps/fxlab/app/src/main/cpp/DuplexEngine.cpp#L76) where we wait at most 500 seconds to start a stream. Can you try increasing this to a large number on the Realme? Thanks!

Perhaps the real change is to increase this timeout.

robertwu1:
We're open to the patch to remove the managed stream. Thanks!

philburk:
> There was no error, the read just returned null at the beginning,

If the result.value() is zero then that just mens there was no data available.
That can happen when the streams are stabilizing.

We should probably use the FullDuplexStream class in the fxlab example.

dipak140:
The LiveEffect sample uses FullDuplexStream, I can raise a PR.

philburk:
I think we all agree.
FXLab should use std::shared_ptr instead of ManagedStreams.
And it should use a FullDuplexStream.
See #2157 | fxlab should use std:shared_ptr and FullDuplexStream

This PR should not be used so I am closing it.

PR #2137, opened 2025-01-15
DO NOT MERGE: OboeTester: Add game mode for testing dynamic workload activity

robertwu1:
Note that GameMode and GameState are different.
[GameMode](https://developer.android.com/games/optimize/adpf/gamemode/gamemode-api) is the actual performance mode of the game (STANDARD, PERFORMANCE, BATTERY).
[GameState](GameState) takes in at least 2 parameters, isLoading and Mode (GAMEPLAY_INTERRUPTIBLE, GAMEPLAY_UNINTERRUPTIBLE, CONTENT), and is the app specified state and used for system optimization.

GameManager allows apps to set GameState and users to set the GameMode. To set the GameMode, you must switch the game mode manually with the [Game Dashboard](https://developer.android.com/games/optimize/adpf/gamemode/gamemode-api#switch_game_modes) outside of the game.

It seems like GameMode is an API that allows apps to handle each of the 3 modes (STANDARD, PERFORMANCE, BATTERY) differently.

If audio apps were to use the game mode API, they can consider changing different buffer sizes depending on the mode.

### DIFF
--- a/apps/OboeTester/app/src/main/AndroidManifest.xml
+++ b/apps/OboeTester/app/src/main/AndroidManifest.xml
@@ -30,7 +30,10 @@
         android:supportsRtl="true"
         android:theme="@style/AppTheme"
         android:requestLegacyExternalStorage="true"
-        android:banner="@mipmap/ic_launcher">
+        android:banner="@mipmap/ic_launcher"
+        android:appCategory="game">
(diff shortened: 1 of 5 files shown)


robertwu1:
Abandoning because I don't want to make OboeTester as a game

PR #2140, opened 2025-01-17
OboeTester: Fix race condition with setWorkload in Dynamic Workload Activity

robertwu1:
If the workload is set in the middle of onAudioReady(), the new workload can be used for the synth before audioStream->reportWorkload() is called for this workload.

The fix here is to use an atomic variable and the same workload throughout the function.


philburk (approved):
Good catch!
mNumWorkloadVoices is modified asynchronously by Java.

PR #2164, opened 2025-02-14
OboeTester: Add HW getters to data paths test

### LINKED ISSUE #2161: Add hardware configuration information
Some of the test result, such as data path test, they only report the requested configuration and the actual configuration. It will also be good to report the hardware configuration, which is available since API 34. With the hardware configuration, it may be possible for us to better identify if the failure is related to hardware configuration or not.

robertwu1:
Fixes #2161


philburk (commented):
LGTM

PR #2167, opened 2025-02-22
Use new AAudio API to query MMAP policy.

### LINKED ISSUE #2163: Support AAudio new APIs for querying mmap capabilities
From API 36, aaudio supports API to query mmap policy for particular device type. Oboe should also add the same support.

flamme:
Fixes #2163

### DIFF
--- a/include/oboe/Definitions.h
+++ b/include/oboe/Definitions.h
@@ -835,6 +835,204 @@ namespace oboe {
     };
 
     /**
+     * Audio device type.
+     *
+     * Note that these match the device types defined in android/media/AudioDeviceInfo.java.
+     *
+     * Added in API level 36 for AAudio.
+     */
+    enum class DeviceType : int32_t {
+        /**
+         * A device type describing the attached earphone speaker.
+         */
+        BuiltinEarpiece = 1,
+
+        /**
+         * A device type describing the speaker system (i.e. a mono speaker or stereo speakers)
+         * built in a device.
+         */
+        BuiltinSpeaker = 2,
+
+        /**
+         * A device type describing a headset, which is the combination of a headphones and
+         * microphone.
+         */
+        WiredHeadset = 3,
+
+        /**
+         * A device type describing a pair of wired headphones.
+         */
+        WiredHeadphones = 4,
+
+        /**
+         * A device type describing an analog line-level connection.
+         */
+        LineAnalog = 5,
+
+        /**
+         * A device type describing a digital line connection (e.g. SPDIF).
+         */
+        LineDigital = 6,
+
+        /**
+         * A device type describing a Bluetooth device typically used for telephony.
+         */
+        BluetoothSco = 7,
+
+        /**
+         * A device type describing a Bluetooth device supporting the A2DP profile.
+         */
+        BluetoothA2dp = 8,
+
+        /**
+         * A device type describing an HDMI connection .
+         */
+        Hdmi = 9,
+
+        /**
+         * A device type describing the Audio Return Channel of an HDMI connection.
+         */
+        HdmiArc = 10,
+
+        /**
+         * A device type describing a USB audio device.
+         */
+        UsbDevice = 11,
+
+        /**
+         * A device type describing a USB audio device in accessory mode.
+         */
+        UsbAccessory = 12,
+
+        /**
+         * A device type describing the audio device associated with a dock.
+         */
+        Dock = 13,
+
+        /**
+         * A device type associated with the transmission of audio signals over FM.
+         */
+        FM = 14,
+
+        /**
+         * A device type describing the microphone(s) built in a device.
+         */
+        BuiltinMic = 15,
+
+        /**
+         * A device type for accessing the audio content transmitted over FM.
+         */
+        FMTuner = 16,
+
+        /**
+         * A device type for accessing the audio content transmitted over the TV tuner system.
+         */
+        TVTuner = 17,
+
+        /**
+         * A device type describing the transmission of audio signals over the telephony network.
+         */
+        Telephony = 18,
+
+        /**
+         * A device type describing the auxiliary line-level connectors.
+         */
+        AuxLine = 19,
+
+        /**
+         * A device type connected over IP.
+         */
+        IP = 20,
+
+        /**
+         * A type-agnostic device used for communication with external audio systems.
+         */
+        Bus = 21,
+
+        /**
+         * A device type describing a USB audio headset.
+         */
+        UsbHeadset = 22,
+
+        /**
+         * A device type describing a Hearing Aid.
+         */
+        HearingAid = 23,
+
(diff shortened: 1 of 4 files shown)


include/oboe/Definitions.h
```diff
@@ -834,6 +834,204 @@ namespace oboe {
...
     };
 
+    /**
+     * Audio device type.
+     *
+     * Note that these match the device types defined in android/media/AudioDeviceInfo.java.
+     *
+     * Added in API level 36 for AAudio.
+     */
+    enum class DeviceType : int32_t {
```
robertwu1, line 844: Should we have something that's the equivalent of 0?

src/aaudio/AAudioExtensions.h
```diff
@@ -37,22 +38,80 @@ namespace oboe {
 
 typedef struct AAudioStreamStruct         AAudioStream;
 
+const static std::set<DeviceType> ALL_OUTPUT_DEVICE_TYPES = {
```
robertwu1, line 41: What happens when the device types expand over time? I don't think we'll have issues with expanding list but I just wanted to confirm

src/aaudio/AAudioExtensions.h
```diff
@@ -80,12 +142,22 @@ class AAudioExtensions {
      * @return 0 or a negative error code
      */
     int32_t setMMapEnabled(bool enabled) {
+        if (mLibLoader->aaudio_setMMapPolicy != nullptr) {
```
robertwu1, line 145: Can you add a comment for this file explaining why there are 2 cases, one when the APIs are officially supported and one when they arent?

include/oboe/Definitions.h
flamme: I'm thinking no as the device type must be something, it cannot be unspecified.

@philburk , could you also share your opinions here? Thanks,

src/aaudio/AAudioExtensions.h
flamme: Then the list here should also be updated. Added some comment

src/aaudio/AAudioExtensions.h
flamme: Done

include/oboe/Definitions.h
```diff
@@ -834,6 +834,204 @@ namespace oboe {
...
+         * will be delivered via audio framework data pipeline.
+         */
+        Auto,
+
+        /**
+         * AAudio MMAP must be used or fail.
+         */
+        Always
+    };
+    typedef int32_t aaudio_policy_t;
```
philburk, line 1033: Do we need this typedef?

src/aaudio/AAudioExtensions.h
```diff
@@ -80,12 +144,27 @@ class AAudioExtensions {
      * @return 0 or a negative error code
      */
     int32_t setMMapEnabled(bool enabled) {
+        // The API for setting mmap policy is on public after API level 36.
```
philburk, line 147: "is public"

src/aaudio/AAudioExtensions.h
```diff
@@ -80,12 +144,27 @@ class AAudioExtensions {
...
+        // When there is no public API, fallback to loading the symbol from hidden API.
         if (loadSymbols()) return AAUDIO_ERROR_UNAVAILABLE;
         if (mAAudio_setMMapPolicy == nullptr) return false;
-        return mAAudio_setMMapPolicy(enabled ? AAUDIO_POLICY_AUTO : AAUDIO_POLICY_NEVER);
+        return mAAudio_setMMapPolicy(
+                static_cast<int32_t>(enabled ? MMapPolicy::Auto : MMapPolicy::Never));
     }
 
     bool isMMapEnabled() {
+        // The API for getting mmap policy is on public after API level 36.
```
philburk, line 160: "is public"

src/aaudio/AAudioExtensions.h
```diff
@@ -66,6 +127,9 @@ class AAudioExtensions {
     }
 
     bool isMMapUsed(AAudioStream *aaudioStream) {
+        if (mLibLoader->stream_isMMapUsed != nullptr) {
```
philburk, line 130: On line 183 we check for (mLibLoader == nullptr)!
Is it Nullable?

src/aaudio/AAudioExtensions.h
```diff
@@ -80,12 +144,27 @@ class AAudioExtensions {
...
         if (mAAudio_setMMapPolicy == nullptr) return false;
-        return mAAudio_setMMapPolicy(enabled ? AAUDIO_POLICY_AUTO : AAUDIO_POLICY_NEVER);
+        return mAAudio_setMMapPolicy(
+                static_cast<int32_t>(enabled ? MMapPolicy::Auto : MMapPolicy::Never));
     }
 
     bool isMMapEnabled() {
+        // The API for getting mmap policy is on public after API level 36.
+        // Use it when it is available.
+        if (mLibLoader->aaudio_getMMapPolicy != nullptr) {
```
philburk, line 162: The loadsymbols() call is after this check.
And loadsymbols() sets mLibLoader->aaudio_getMMapPolicy.
So I think the first time this gets called then it will fall back to using mAAudio_getMMapPolicy().
The second call will find the symbol and use it.
Maybe loadsymbols() should come first.

But I may be tracing the code wrong...

philburk (changes requested):

include/oboe/Definitions.h
flamme: Removed

src/aaudio/AAudioExtensions.h
flamme: Oh, that is a good point! I should also modify `loadSymbols`. The LibLoader will be opened from the constructor of AAudioExtensions when calling `initMMapPolicy`. If it is opened successfully, `mLibLoader->aaudio_getMMapPolicy` will be set if it is on Android 16+. Otherwise, it fallbacks to the old logic.

robertwu1 (approved):

src/aaudio/AAudioExtensions.h
philburk: > The LibLoader will be opened from the constructor of AAudioExtensions when calling initMMapPolicy

In the new code:
AAudioExtensions() constructor calls initMMapPolicy():

```
AAudioExtensions() {
        mLibLoader = AAudioLoader::getInstance();
        if (!initMMapPolicy()) {
```

Since mLibLoader is non-null, the open() will get called:

```
    bool initMMapPolicy() {
        if (mLibLoader == nullptr || mLibLoader->open() != 0) {
            return false;
        }
```

The open() method sets aaudio_getPlatformMMapPolicy and aaudio_getMMapPolicy.

So I guess this is OK.

philburk (approved):

PR #2169, opened 2025-02-27
OboeTester: Add supported mixer attributes to device report

robertwu1:


apps/OboeTester/app/src/main/java/com/mobileer/audio_device/AudioDeviceInfoConverter.java
```diff
@@ -74,30 +76,37 @@ public static String toString(AudioDeviceInfo adi){
...
 
-        if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.S) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
             sb.append("\nAudio Profiles: ");
             List<AudioProfile> audioProfiles = adi.getAudioProfiles();
             sb.append(audioProfiles);
         }
 
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.UPSIDE_DOWN_CAKE) {
+            sb.append("\nSupported Mixer Attributes: ");
```
flamme, line 104: It may only be needed if it is a USB output device. For all other kinds of device, it will just be empty.

robertwu1: It's fine if we output an empty array on the other devices. It's better for debugging if we write [] over not writing anything

flamme (approved):

philburk (approved):

PR #2177, opened 2025-03-12
Redefine AAudio_DeviceType_t

robertwu1:
[AAudio_DeviceType](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/av/media/libaaudio/include/aaudio/AAudio.h;l=1?q=frameworks%2Fav%2Fmedia%2Flibaaudio%2Finclude%2Faaudio%2FAAudio.h&ss=android%2Fplatform%2Fsuperproject%2Fmain) is now defined in AOSP.

The issue is that this is a `typedef enum AAudio_DeviceType : int32_t`. We can't recreate it with `typedef int32_t AAudio_DeviceType;` like we do now as the types are different as the former is an enum type with a strict set of values. I propose that we create AAudio_DeviceType_t as an int32_t so at least we can cast to it.

Also, it seems like Android B is NDK 29 now that we have quarterly releases.

### DIFF
--- a/src/aaudio/AAudioLoader.h
+++ b/src/aaudio/AAudioLoader.h
@@ -96,11 +96,11 @@ typedef int32_t aaudio_spatialization_behavior_t;
 #define __ANDROID_API_B__ 36
 #endif
 
-#if __NDK_MAJOR__ < 28
-// These were defined in W
-typedef int32_t AAudio_DeviceType;
+#if __NDK_MAJOR__ < 29
+// These was defined in W
 typedef int32_t aaudio_policy_t;
 #endif
+typedef int32_t AAudio_DeviceType_t; // This is different because AAudio_DeviceType is an enum type
 
 namespace oboe {
 
--- a/src/aaudio/AAudioExtensions.h
+++ b/src/aaudio/AAudioExtensions.h
@@ -185,7 +185,7 @@ public:
             return MMapPolicy::Unspecified;
         }
         return static_cast<MMapPolicy>(mLibLoader->aaudio_getPlatformMMapPolicy(
-                static_cast<AAudio_DeviceType>(deviceType),
+                static_cast<AAudio_DeviceType_t>(deviceType),
                 static_cast<aaudio_direction_t>(direction)));
     }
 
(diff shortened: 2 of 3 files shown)


src/aaudio/AAudioLoader.h
```diff
@@ -96,11 +96,11 @@ typedef int32_t aaudio_spatialization_behavior_t;
...
 #endif
 
-#if __NDK_MAJOR__ < 28
-// These were defined in W
-typedef int32_t AAudio_DeviceType;
+#if __NDK_MAJOR__ < 29
+// These was defined in W
 typedef int32_t aaudio_policy_t;
 #endif
+typedef int32_t AAudio_DeviceType_t; // This is different because AAudio_DeviceType is an enum type
```
flamme, line 103: Just would like to confirm that is the way we want to go instead of adding a compiler option in external/oboe?

src/aaudio/AAudioLoader.h
```diff
@@ -96,11 +96,11 @@ typedef int32_t aaudio_spatialization_behavior_t;
 #define __ANDROID_API_B__ 36
 #endif
 
-#if __NDK_MAJOR__ < 28
-// These were defined in W
-typedef int32_t AAudio_DeviceType;
+#if __NDK_MAJOR__ < 29
```
flamme, line 99: Maybe all __NDK_MAJOR__ check should be moved to before we define __NDK_MAJOR__ in this file. Then it will looks like

```
#ifdef __NDK_MAJOR__
#if __NDK_MAJOR__  <29
...
```

philburk: This may be confusing to future programmers or AI.
So the solution should be self documenting. How about:

```
#ifdef __NDK_MAJOR__ 
    #define __NDK_MAJOR__ 0
    #define OBOE_USING_NDK 0
#else
    #define OBOE_USING_NDK 1
#endif
```
then later

```
#if  OBOE_USING_NDK && __NDK_MAJOR__ < 28
    typedef int32_t AAudio_DeviceType;
    typedef int32_t aaudio_policy_t;
#endif
```

src/aaudio/AAudioLoader.h
philburk: I think having the similar but slightly different symbol is confusing.

philburk (changes requested):

robertwu1:
Abandoning in favor of https://github.com/google/oboe/pull/2181

PR #2182, opened 2025-03-13
AudioStream: Add std::ignore for unused parameters

robertwu1:
These parameters are not used so certain compliers give warnings. We should not use something like /* appWorkload */  because these are virtual functions

### DIFF
--- a/include/oboe/AudioStream.h
+++ b/include/oboe/AudioStream.h
@@ -569,10 +569,13 @@ public:
      * @return OK or an error such as ErrorInvalidState if the PerformanceHint was not enabled.
      */
     virtual oboe::Result reportWorkload(int32_t appWorkload) {
+        std::ignore = appWorkload;
         return oboe::Result::ErrorUnimplemented;
     }
 
     virtual oboe::Result setOffloadDelayPadding(int32_t delayInFrames, int32_t paddingInFrames) {
+        std::ignore = delayInFrames;
+        std::ignore = paddingInFrames;
         return Result::ErrorUnimplemented;
     }
 


include/oboe/AudioStream.h
```diff
@@ -569,10 +569,13 @@ class AudioStream : public AudioStreamBase {
      * @return OK or an error such as ErrorInvalidState if the PerformanceHint was not enabled.
      */
     virtual oboe::Result reportWorkload(int32_t appWorkload) {
```
flamme, line 571: Surprised that it complained for header file. It is usually for cpp. BTW, using `__unused` will work here.

robertwu1: I don't like __unused here because it's used when its overriden

flamme: I have to say std::ignore is not the designed for unused parameters. See [doc](https://en.cppreference.com/w/cpp/utility/tuple/ignore). Could you share any code guide for using std::ignore?

> I don't like __unused here because it's used when its overriden

Using `__unused` isn't really indicate the parameters will be unused at all overridden functions. If you don't prefer the C style `__unused`, it is possible to use `[[maybe_unused]]` in c++. Or just comment it out.

Using std::ignore may cause build error if "<tuple>" or "<utility>" is not included...

robertwu1: I think it's fine honestly as maybe_unused is recent. From the doc:

A void expression or a volatile bit-field value cannot be assigned to std::ignore.

std::ignore is intended for use with [std::tie](https://en.cppreference.com/w/cpp/utility/tuple/tie) when unpacking a [std::tuple](https://en.cppreference.com/w/cpp/utility/tuple), as a placeholder for the arguments that are not used, but can be used for any unwanted assignment.

Some code guides recommend using std::ignore to avoid warnings from unused return values of [[[nodiscard](https://en.cppreference.com/w/cpp/language/attributes/nodiscard)]] functions, even though an assignment isn't required.

For ignoring values not requiring assignment, one may cast to void. For variables that have names, but whose value is unused, one may cast those to void or declare those variables with [[[maybe_unused](https://en.cppreference.com/w/cpp/language/attributes/maybe_unused)]].

flamme: As you mentioned, the doc suggest to use `[[maybe_unused]]` instead of `std::ignore`. 

Theoretically, using `std::ignore` will suppress the unused parameters warning now, but it is not preferred and it will always ask to include `<tuple>` header. See ISO c++ Q&A at https://isocpp.org/blog/2016/09/quick-q-stdignore-for-ignoring-unused-variable.

> I think it's fine honestly as maybe_unused is recent

Oboe is compiled with [c++17](https://github.com/google/oboe/blob/main/CMakeLists.txt#L86) since 4 years ago, which should support `[[maybe_unused]]`.

From our internal code guideline, it is suggested to comment out the name for unused parameters.

robertwu1: Created https://github.com/google/oboe/pull/2183

Issue #2180, opened 2025-03-13
Add compress offload example in OboeTester


flamme:
Offload support is added at https://github.com/google/oboe/issues/2154. Currently, OboeTester only test PCM offload. Need to add compress offload test.

PR #2188, opened 2025-03-21
Add compressed formats to valid formats list.

flamme:
This is part of the fix for #2180

### DIFF
--- a/include/oboe/AudioStream.h
+++ b/include/oboe/AudioStream.h
@@ -25,6 +25,7 @@
 #include "oboe/ResultWithValue.h"
 #include "oboe/AudioStreamBuilder.h"
 #include "oboe/AudioStreamBase.h"
+#include "oboe/Utilities.h"
 
 namespace oboe {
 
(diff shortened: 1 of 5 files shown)


robertwu1:
Can you do a search for code search for IEC61937? We should probably add similar code wherever IEC61937 is used

flamme:
> Can you do a search for code search for IEC61937? We should probably add similar code wherever IEC61937 is used

Thanks for the reminder! I added them to the oboe sources.

include/oboe/AudioStream.h
```diff
@@ -242,14 +243,20 @@ class AudioStream : public AudioStreamBase {
      * and the sample format. For example, a 2 channel floating point stream will have
      * 2 * 4 = 8 bytes per frame.
      *
+     * Note for compressed formats, bytes per frames is treated as 1 by convention.
+     *
      * @return number of bytes in each audio frame.
      */
-    int32_t getBytesPerFrame() const { return mChannelCount * getBytesPerSample(); }
+    int32_t getBytesPerFrame() const {
+        return isCompressedFormat(mFormat) ? 1 : mChannelCount * getBytesPerSample(); }
```
robertwu1, line 251: The messaging in getBytesPerFrame() and getBytesPerSample() aren't consistent. Should we return the same number for both?

flamme: Good catch! For compressed format, we consider bytes per sample as 0 due to not fixed size per sample. For bytes per frame, it is considered as 1 by convention. I updated the comments.

robertwu1 (approved):

PR #2199, opened 2025-04-15
OboeTester: Add AudioWorkloadTestActivity and AudioWorkloadTestRunnerActivity

robertwu1:
These activities use the new classes AudioWorkloadTest and AudioWorkloadTestRunner which can be used outside of OboeTester in isolation.


apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTest.h
```diff
@@ -0,0 +1,79 @@
...
+    struct CallbackStatus {
+        int32_t numVoices;
+        int64_t beginTimeNs;
+        int64_t finishTimeNs;
+        int32_t xRunCount;
+        int32_t cpuIndex;
+    };
+
+    AudioWorkloadTest();
+    int32_t open();
```
philburk, line 42: Add simple method docs and an example of calling the class.
That may increase adoption.

apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTestRunner.cpp
```diff
@@ -0,0 +1,118 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "AudioWorkloadTestRunner.h"
+#include <iostream>
+#include <iomanip>
+#include <sstream>
+
+AudioWorkloadTestRunner::AudioWorkloadTestRunner() : mAudioWorkloadTest(std::make_unique<AudioWorkloadTest>()) {}
```
philburk, line 22: Can you just statically allocate an AudioWorkloadTest here?

apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTestRunner.cpp
```diff
@@ -0,0 +1,118 @@
...
+        mIsDone = true;
+        mIsRunning = false;
+        mAudioWorkloadTest->close();
+        return -1;
+    }
+
+    return 0;
+}
+
+bool AudioWorkloadTestRunner::pollIsDone() {
```
philburk, line 77: Maybe "stopIfDone()"

apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTestRunner.cpp
```diff
@@ -0,0 +1,118 @@
...
+    }
+    return mIsDone;
+}
+
+std::string AudioWorkloadTestRunner::getStatus() const {
+    if (!mIsRunning) {
+        return mResultText;
+    }
+    int32_t callbacksCompleted = mAudioWorkloadTest->getCallbackCount();
+    return "Running: " + std::to_string(callbacksCompleted) + "/" + std::to_string(mNumCallbacks) + " callbacks completed. XRuns: " + std::to_string(mAudioWorkloadTest->getXRunCount());
```
philburk, line 89: long line

apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTestRunner.h
```diff
@@ -0,0 +1,55 @@
...
+#include <memory>
+#include "AudioWorkloadTest.h"
+
+class AudioWorkloadTestRunner {
+public:
+    AudioWorkloadTestRunner();
+    ~AudioWorkloadTestRunner();
+
+    int32_t start(
+            int32_t numCallbacks,
```
philburk, line 32: Using numCallbacks for the duration will make it hard to run a test for N seconds.
We may need that if a caller is trying to limit the time spent on each test.
Maybe pass a duration in seconds and calculate the numCallbacks after you get the sample Rate and framesPerBurst.

apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTestRunner.h
```diff
@@ -0,0 +1,55 @@
...
+            int32_t bufferSizeInBursts,
+            int32_t numVoices,
+            int32_t highNumVoices,
+            int32_t highLowPeriodMillis,
+            bool adpfEnabled,
+            bool sineEnabled);
+    bool pollIsDone();
+    std::string getStatus() const;
+    int32_t stop();
+    int32_t getResult() const;
```
philburk, line 42: 'at'return 0: not finished, 1: pass, -1: fail

apps/OboeTester/app/src/main/cpp/cpu/SynthWorkload.h
```diff
@@ -0,0 +1,41 @@
...
+public:
+    SynthWorkload();
+    SynthWorkload(int onFrames, int offFrames);
+    void onCallback(double workload);
+    void renderStereo(float *buffer, int numFrames);
+
+private:
+    marksynth::Synthesizer   mSynth;
+    static constexpr int     kDummyBufferSizeInFrames = 32;
+    std::unique_ptr<float[]> mDummyStereoBuffer = std::make_unique<float[]>(
```
philburk, line 32: If you are not going to replace this then you can use a statically allocated array.

    float     mDummyStereoBuffer[kDummyBufferSizeInFrames * 2];

apps/OboeTester/app/src/main/cpp/jni-bridge.cpp
```diff
@@ -1052,4 +1054,170 @@ Java_com_mobileer_oboetester_TestRapidCycleActivity_getCycleCount(JNIEnv *env, j
...
+    jmethodID arrayListAdd = env->GetMethodID(arrayListClass, "add", "(Ljava/lang/Object;)Z");
+    if (arrayListAdd == nullptr) {
+        std::cerr << "Error: Could not find 'add' method for ArrayList" << std::endl;
+        return nullptr;
+    }
+
+    jobject javaList = env->NewObject(arrayListClass, arrayListConstructor);
+
+    for (const auto& status : cppCallbackStats) {
+        jobject javaStatus = env->NewObject(
```
philburk, line 1164: Add comment that says this must match the constructor in  AudioWorkloadTestActivity.java

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/AudioWorkloadTestActivity.java
```diff
@@ -0,0 +1,280 @@
...
+    private UpdateThread mUpdateThread;
+
+    public static class CallbackStatus {
+        public int numVoices;
+        public long beginTimeNs;
+        public long finishTimeNs;
+        public int xRunCount;
+        public int cpuIndex;
+
+        public CallbackStatus(int numVoices, long beginTimeNs, long finishTimeNs, int xRunCount, int cpuIndex) {
```
philburk, line 70: Add comment that says this must match the NewObject call in jni-bridge.cpp

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/AudioWorkloadTestActivity.java
```diff
@@ -0,0 +1,280 @@
...
+            this.finishTimeNs = finishTimeNs;
+            this.xRunCount = xRunCount;
+            this.cpuIndex = cpuIndex;
+        }
+
+        @NonNull
+        @Override
+        public String toString() {
+            return "CallbackStatus{" +
+                    "numVoices=" + numVoices +
```
philburk, line 82: Could call toShortString() here.

philburk (changes requested):

apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTestRunner.h
```diff
@@ -23,33 +23,90 @@
...
+     * @brief Destructor. Ensures the test is stopped when the runner is destroyed.
+     */
     ~AudioWorkloadTestRunner();
 
+    /**
+     * @brief Starts the audio workload test with the specified parameters.
+     * @param targetDurationMs The desired duration of the test in milliseconds.
+     * @param bufferSizeInBursts The desired buffer size in terms of multiples of framesPerBurst.
+     * @param numVoices The primary number of synthesizer voices to simulate.
+     * @param highNumVoices An alternative number of voices for alternating workload.
```
philburk, line 56: I f I do not want to alternate, should I set this to zero or the same as numVoices?

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/AudioWorkloadTestActivity.java
```diff
@@ -0,0 +1,281 @@
...
+    }
+
+    public void stopTest(View view) {
+        stopTest();
+    }
+
+    private void stopTest() {
+        stop();
+        StringBuilder callbackStatisticsText = new StringBuilder();
+        List<CallbackStatus> callbackStatuses = getCallbackStatistics();
```
flamme, line 220: Should check the if the list is null or not given null can be returned.

apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTestRunner.h
```diff
@@ -0,0 +1,112 @@
...
+            int32_t numVoices,
+            int32_t highNumVoices,
+            int32_t highLowPeriodMillis,
+            bool adpfEnabled,
+            bool hearWorkload);
+
+    /**
+     * @brief Stops the test if it has completed its run (i.e., AudioWorkloadTest is no longer running).
+     * This is useful for polling and cleaning up once the underlying test finishes due to duration.
+     * @return True if the test is done (either stopped here or previously), false otherwise.
```
flamme, line 74: This looks wrong. The value is initialized as false. When the stream is not started, it returns false.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/AudioWorkloadTestRunnerActivity.java
```diff
@@ -0,0 +1,158 @@
...
+
+        int targetDurationMs = mTargetDurationMsSlider.getValue();
+        int bufferSizeInBursts = mBufferSizeInBurstsSlider.getValue();
+        int numVoices = mNumVoicesSlider.getValue();
+        int highNumVoices = mAlternateNumVoicesSlider.getValue();
+        int highLowPeriodMs = mAlternatingPeriodMsSlider.getValue();
+        boolean adpfEnabled = mEnableAdpfBox.isChecked();
+        boolean hearWorkload = mHearWorkloadBox.isChecked();
+
+        start(targetDurationMs, bufferSizeInBursts, numVoices, highNumVoices, highLowPeriodMs, adpfEnabled, hearWorkload);
```
flamme, line 115: Shouldn't the returned result should be validated?

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/AudioWorkloadTestActivity.java
```diff
@@ -0,0 +1,281 @@
...
+import androidx.annotation.NonNull;
+import androidx.appcompat.app.AppCompatActivity;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Audio Workload Test
+ */
+public class AudioWorkloadTestActivity extends AppCompatActivity {
```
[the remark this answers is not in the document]
robertwu1: Not using a common base class as we want the two tests to be isolated from each other. This way it's easy for users of the c++ libraries to understand how to use the code

apps/OboeTester/app/src/main/cpp/jni-bridge.cpp
```diff
@@ -1052,4 +1054,172 @@ Java_com_mobileer_oboetester_TestRapidCycleActivity_getCycleCount(JNIEnv *env, j
...
+JNIEXPORT jobject JNICALL
+Java_com_mobileer_oboetester_AudioWorkloadTestActivity_getCallbackStatistics(JNIEnv *env, jobject obj) {
+    std::vector<AudioWorkloadTest::CallbackStatus> cppCallbackStats = sAudioWorkload.getCallbackStatistics();
+    jclass callbackStatusClass = env->FindClass("com/mobileer/oboetester/AudioWorkloadTestActivity$CallbackStatus");
+    if (callbackStatusClass == nullptr) {
+        std::cerr << "Error: Could not find Java class CallbackStatus" << std::endl;
+        return nullptr;
+    }
+
+    jmethodID callbackStatusConstructor = env->GetMethodID(callbackStatusClass, "<init>", "(IJJII)V");
```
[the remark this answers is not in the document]
robertwu1: I tried caching this and it just crashes. Not caching for now.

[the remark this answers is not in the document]
robertwu1: Cached in latest commit

apps/OboeTester/app/src/main/cpp/jni-bridge.cpp
```diff
@@ -1126,54 +1126,159 @@ Java_com_mobileer_oboetester_AudioWorkloadTestActivity_close(JNIEnv *env, jobjec
...
-        env->DeleteLocalRef(javaStatus); // Important: Release local references
+        if (javaStatus == nullptr) {
+            std::cerr << "Error: Could not create new CallbackStatus object." << std::endl;
+            if (env->ExceptionCheck()) env->ExceptionDescribe();
+            env->DeleteLocalRef(javaList);
+            return nullptr;
+        }
+
+        env->CallBooleanMethod(javaList, g_arrayListAddMethod, javaStatus);
+        if (env->ExceptionCheck()) {
```
flamme, line 1274: Given it is an arraylist add, I'm not sure what exception may happen here. Could you explain more details about your concerns?

robertwu1: We use it quite liberally within our code in Android. I don't see any issues with extra exception checks as we could have easily typed the signature wrong for g_arrayListAddMethod. This makes it so when others copy paste this, they can make sure the type is right

flamme: It is not wrong to check the exception... But just if there is not exception to be thrown, why we bother to check exception...

If you type wrong signature of `g_arrayListAddMethod`, it should be caught when finding the method instead of here...

robertwu1: See https://www.w3schools.com/java/ref_arraylist_add.asp

ArrayList.add can still return IndexOutOfBoundsException although it should never return it in this case

flamme (approved):

apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTest.cpp
```diff
@@ -0,0 +1,208 @@
...
+        }
+        if (((timeMs - mStartTimeMs) % (2 * mAlternatingPeriodMs)) >= mAlternatingPeriodMs) {
+            currentVoices = mAlternateNumVoices;
+        }
+    }
+
+    auto floatData = static_cast<float *>(audioData);
+    int channelCount = audioStream->getChannelCount();
+
+    // Fill buffer with a sine wave.
```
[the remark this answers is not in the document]
philburk: This code is intended to be used in other projects like GTS-Interactive.
So we are trying to reduce the number of dependencies.

PR #2200, opened 2025-04-16
OboeTester: remove extra mSinkMemoryDirect nullptr check

robertwu1:
Not needed as mSinkMemoryDirect is created in this function


flamme (commented):
Should nullptr check also use at ActivityTestOutput::close given ActivityTapToTone doesn't override it?

flamme (approved):

robertwu1:
> Should nullptr check also use at ActivityTestOutput::close given ActivityTapToTone doesn't override it?

Not needed as close is for the shared_ptr and not for the underlying SinkMemoryDirect

PR #2206, opened 2025-04-29
OboeTester: Add package name and attribution tag

robertwu1:
This PR adds package names and attribution tags to OboeTester. These should be declared for audio streams or there may be permission issues with microphone permissions.

### DIFF
--- a/apps/OboeTester/app/src/main/cpp/jni-bridge.cpp
+++ b/apps/OboeTester/app/src/main/cpp/jni-bridge.cpp
@@ -525,7 +525,7 @@ Java_com_mobileer_oboetester_OboeAudioStream_getPackageName(
     if (oboeStream != nullptr) {
         return env->NewStringUTF(oboeStream->getPackageName().c_str());
     }
-    return (jstring) "";
+    return env->NewStringUTF("");
 }
 
 JNIEXPORT jstring JNICALL
@@ -535,7 +535,7 @@ Java_com_mobileer_oboetester_OboeAudioStream_getAttributionTag(
     if (oboeStream != nullptr) {
         return env->NewStringUTF(oboeStream->getAttributionTag().c_str());
     }
-    return (jstring) "";
+    return env->NewStringUTF("");
 }
 
 JNIEXPORT jlong JNICALL
--- a/apps/OboeTester/app/src/main/cpp/jni-bridge.cpp
+++ b/apps/OboeTester/app/src/main/cpp/jni-bridge.cpp
@@ -61,7 +61,9 @@ Java_com_mobileer_oboetester_OboeAudioStream_openNative(JNIEnv *env, jobject,
                                                        jint rateConversionQuality,
                                                        jboolean isMMap,
                                                        jboolean isInput,
-                                                       jint spatializationBehavior);
+                                                       jint spatializationBehavior,
+                                                       jstring packageName,
+                                                       jstring attributionTag);
 JNIEXPORT void JNICALL
 Java_com_mobileer_oboetester_OboeAudioStream_close(JNIEnv *env, jobject, jint);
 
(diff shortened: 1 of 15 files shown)


apps/OboeTester/app/src/main/cpp/jni-bridge.cpp
```diff
@@ -504,6 +518,26 @@ Java_com_mobileer_oboetester_OboeAudioStream_getSessionId(
...
 }
 
+JNIEXPORT jstring JNICALL
+Java_com_mobileer_oboetester_OboeAudioStream_getPackageName(
+        JNIEnv *env, jobject, jint streamIndex) {
+    std::shared_ptr<oboe::AudioStream> oboeStream = engine.getCurrentActivity()->getStream(streamIndex);
+    if (oboeStream != nullptr) {
+        return env->NewStringUTF(oboeStream->getPackageName().c_str());
+    }
+    return (jstring) "";
```
philburk, line 528: Is this enough or do you need to call NewStringUTF("")?

apps/OboeTester/app/src/main/cpp/jni-bridge.cpp
```diff
@@ -504,6 +518,26 @@ Java_com_mobileer_oboetester_OboeAudioStream_getSessionId(
...
+}
+
+JNIEXPORT jstring JNICALL
+Java_com_mobileer_oboetester_OboeAudioStream_getAttributionTag(
+        JNIEnv *env, jobject, jint streamIndex) {
+    std::shared_ptr<oboe::AudioStream> oboeStream = engine.getCurrentActivity()->getStream(streamIndex);
+    if (oboeStream != nullptr) {
+        return env->NewStringUTF(oboeStream->getAttributionTag().c_str());
+    }
+    return (jstring) "";
```
philburk, line 538: ditto

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/AudioStreamTester.java
```diff
@@ -28,6 +28,8 @@ AudioStreamBase getCurrentAudioStream() {
     }
 
     public void open() throws IOException {
+        requestedConfiguration.setPackageName(BuildConfig.APPLICATION_ID);
+        requestedConfiguration.setAttributionTag("AudioTag"); // Match attribution in AndroidManifest.xml
```
philburk, line 32: Do you need to read the value of "AudioTag" from the Manifest?

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/OboeAudioStream.java
```diff
@@ -258,6 +264,15 @@ public int getSessionId() {
     }
     private native int getSessionId(int streamIndex);
 
+    public String getPackageName() {
+        return getPackageName(streamIndex);
```
flamme, line 268: It is not related to this PR, but why `streamIndex` doesn't have prefix `m`.

I'm a little bit confused when I set the function here call `getPackageName(streamIndex)` while `streamIndex` is the parameter name for the bottom native method. And finally figure out `streamIndex` is a local member for the class.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/AudioStreamTester.java
flamme: +1 for reading the value from manifest.

apps/OboeTester/app/src/main/cpp/jni-bridge.cpp
flamme: This is dangerous as it directly to create a jstring via a C/C++ style string, it is likely to cause crash or undefined behavior when JAVA side receive it.

flamme:
BTW, from a test purpose, do we want to allow to random set package name and attribution tag when opening streams?

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/AudioStreamTester.java
robertwu1: There can be multiple attribution tags in the manifest and there's no way to query it. Thus, we just need to type it twice

robertwu1:
> BTW, from a test purpose, do we want to allow to random set package name and attribution tag when opening streams?

We can but tbh it'll cluster the UI as we almost never use this. I added an example in the unit tests

flamme:
> > BTW, from a test purpose, do we want to allow to random set package name and attribution tag when opening streams?
> 
> We can but tbh it'll cluster the UI as we almost never use this. I added an example in the unit tests

I was thinking it would be handy to test some permission issues.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/AudioStreamTester.java
flamme: There is API... https://developer.android.com/reference/android/content/pm/Attribution#getTag()

flamme (approved):

PR #2210, opened 2025-05-17
Allow sharing of the Trace API for Perfetto

philburk:
Use Singleton design pattern and early initialization

Add a "Workload" trace for OboeTester CPU LOAD test

### DIFF
--- a/apps/OboeTester/app/src/main/cpp/NativeAudioContext.h
+++ b/apps/OboeTester/app/src/main/cpp/NativeAudioContext.h
@@ -170,7 +170,10 @@ public:
 
     void setWorkload(int32_t workload) {
         oboeCallbackProxy.setWorkload(workload);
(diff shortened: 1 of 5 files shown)


apps/OboeTester/app/src/main/cpp/NativeAudioContext.h
```diff
@@ -169,6 +170,7 @@ class ActivityContext {
 
     void setWorkload(int32_t workload) {
         oboeCallbackProxy.setWorkload(workload);
+        oboe::Trace::getInstance().setCounter("Workload", workload);
```
robertwu1, line 173: Should we check isEnabled here?

philburk: Good idea.

robertwu1 (approved):

PR #2214, opened 2025-05-21
AdpfWrapper: Add notifyWorkloadIncrease/Spike/Reset

robertwu1:
Add new ADPF APIs introduced in Android API level 36 to Oboe. When testing with the new AudioWorkload test, this works great!

### DIFF
--- a/apps/OboeTester/app/src/main/cpp/jni-bridge.cpp
+++ b/apps/OboeTester/app/src/main/cpp/jni-bridge.cpp
@@ -627,14 +627,14 @@ Java_com_mobileer_oboetester_OboeAudioStream_setWorkload(
     engine.getCurrentActivity()->setWorkload(workload);
 }
 
-static const std::string sWorkloadDebugName = "OboeTester";
+static const char* sWorkloadDebugName = "OboeTester";
 
 JNIEXPORT jint JNICALL
 Java_com_mobileer_oboetester_OboeAudioStream_notifyWorkloadIncrease(
         JNIEnv *env, jobject, jint streamIndex, jboolean cpu, jboolean gpu) {
     std::shared_ptr<oboe::AudioStream> oboeStream = engine.getCurrentActivity()->getStream(streamIndex);
     if (oboeStream != nullptr) {
-        return (jint) oboeStream->notifyWorkloadIncrease(cpu, gpu, sWorkloadDebugName.c_str());
+        return (jint) oboeStream->notifyWorkloadIncrease(cpu, gpu, sWorkloadDebugName);
     }
     return -1;
 }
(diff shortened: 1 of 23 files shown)


apps/OboeTester/docs/AutomatedTesting.md
```diff
@@ -137,6 +137,8 @@ There are some optional parameters for just the "cpu_load" test:
     --ez use_adpf         {true, false} // if true, use work boost from performance hints. Default is false.
     --ez use_workload     {true, false} // if true and using ADPF then report workload changes. Default is false.
     --ez scroll_graphics  {true, false} // if true then continually update the power scope. Default is false.
+    --ez use_cpu_hint     {true, false} // if true and using ADPF, notify adpf for cpu usage increase. Default is false.
+    --ez use_Gpu_hint     {true, false} // if true and using ADPF, notify adpf for gpu usage increase. Default is false.
```
philburk, line 141: OboeTester is not using the GPU. So how would the usage increase?

include/oboe/AudioStream.h
```diff
@@ -579,6 +579,100 @@ class AudioStream : public AudioStreamBase {
         return oboe::Result::ErrorUnimplemented;
     }
 
+    /**
+    * Informs the framework of an upcoming increase in the workload of a graphics pipeline
```
philburk, line 583: This call is not really specific to graphics. They were just focussed on graphics when they were writing the doc. Maybe say "workload of an audio callback".

apps/OboeTester/app/src/main/cpp/jni-bridge.cpp
```diff
@@ -627,6 +627,28 @@ Java_com_mobileer_oboetester_OboeAudioStream_setWorkload(
     engine.getCurrentActivity()->setWorkload(workload);
 }
 
+static const std::string sWorkloadDebugName = "OboeTester";
```
philburk, line 630: Could just use char* since you have to pass a c_str anyway.

src/aaudio/AudioStreamAAudio.h
```diff
@@ -108,6 +108,27 @@ class AudioStreamAAudio : public AudioStream {
         return oboe::Result::OK;
     }
 
+    oboe::Result notifyWorkloadIncrease(bool cpu, bool gpu, const char* debugName) override {
+        if (!isPerformanceHintEnabled()) {
+            return oboe::Result::ErrorInvalidState;
+        }
+        return mAdpfWrapper.notifyWorkloadIncrease(cpu, gpu, debugName);
```
philburk, line 115: Does this have to be called from the callback thread?
Is the ADPF native API thread safe?

philburk (changes requested):

apps/OboeTester/docs/AutomatedTesting.md
robertwu1: Thanks! I changed the options to be much more readable.

philburk (approved):
I did not do a detailed analysis. But LGTM.

PR #2197, opened 2025-04-08
[PCM Offload] PowerPlay Sample App Implementation

madebymozart:
This is the powerplay sample app, which allows a user, through oboe, to choose a new performance mode that allows the CPU to sleep significantly more, thus aiding in it's ability to save power.

PR #2226, opened 2025-06-10
PowerPlay Sample App -> Standalone

madebymozart:
This is the powerplay sample app, which allows a user, through oboe, to choose a new performance mode that allows the CPU to sleep significantly more, thus aiding in it's ability to save power.

This app has been moved to a standalone state and avoid interfering with existing libraries


madebymozart:
In favor of https://github.com/google/oboe/pull/2197

PR #2230, opened 2025-06-19
Input stream not stopped in FullDuplexStream::stop

meteaurestudios:
Fix duplicate calls to `getOutputStream()->requestStop()`, resulting in never calling `getInputStream()->requestStop()`

Issue introduced in [this commit](https://github.com/google/oboe/commit/fae46d4c39a1918c5402cd40b84ef2cff671ff75#diff-93cf3a2be2c05a63390de19b8dc5b4104f4a3ad75d654982588d69d9b32a795a)


robertwu1:
Thank you so much!

PR #2233, opened 2025-06-24
OboeTester: Enable output effects

robertwu1:
There was a bug where OboeTester never enabled its output effects. This PR adds buttons for output effects in OboeTester

### DIFF
--- a/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/StreamConfigurationView.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/StreamConfigurationView.java
@@ -729,14 +729,14 @@ public class StreamConfigurationView extends LinearLayout {
     private void onBassBoostCheckBoxChanged(boolean isChecked) {
         if (mBassBoost != null) {
             mBassBoost.setEnabled(isChecked);
-            mBassBoostSeekBar.setEnabled(isChecked);
         }
+        mBassBoostSeekBar.setEnabled(isChecked);
     }
 
     private void onLoudnessEnhancerCheckBoxChanged(boolean isChecked) {
         if (mLoudnessEnhancer != null) {
             mLoudnessEnhancer.setEnabled(isChecked);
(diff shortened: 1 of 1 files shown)


apps/OboeTester/app/src/main/java/com/mobileer/oboetester/StreamConfigurationView.java
```diff
@@ -709,4 +725,18 @@ private void onNoiseSuppressorCheckBoxChanged(boolean isChecked) {
             mNoiseSuppressor.setEnabled(isChecked);
         }
     }
+
+    private void onBassBoostCheckBoxChanged(boolean isChecked) {
+        if (mBassBoost != null) {
+            mBassBoost.setEnabled(isChecked);
+            mBassBoostSeekBar.setEnabled(isChecked);
```
flamme: If `mBassBoost` is null, it is still better to set `mBassBoostSeekBar.setEnabled` for consistency.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/StreamConfigurationView.java
```diff
@@ -709,4 +725,18 @@ private void onNoiseSuppressorCheckBoxChanged(boolean isChecked) {
...
+        }
+    }
+
+    private void onLoudnessEnhancerCheckBoxChanged(boolean isChecked) {
+        if (mLoudnessEnhancer != null) {
+            mLoudnessEnhancer.setEnabled(isChecked);
+            mLoudnessEnhancerSeekBar.setEnabled(isChecked);
```
flamme: ditto

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/StreamConfigurationView.java
robertwu1: Done!

PR #2234, opened 2025-06-24
AudioWorkloadTest.h: Don't reset the stream pointer on close

robertwu1:
We can reset the stream pointer on open instead as callbacks may be using the pointer.


robertwu1:
abandoning in favor of a different PR

PR #2235, opened 2025-06-26
Workload Tests: Close the audio stream when the activity is closed

robertwu1:
Workload tests leave the stream open even when the activity is closed. This PR fixes that


flamme (commented):
It looks like if the activity is not destroyed, the stream will still be staled and affecting other tests.

robertwu1:
> It looks like if the activity is not destroyed, the stream will still be staled and affecting other tests.

changed to onStop(). Thanks!

flamme (approved):
Please don't forget to update the commit message, it is called from onStop instead of destroy

PR #2236, opened 2025-06-26
Fix mCallbackStatistics race condition in AudioWorkloadTest.h

robertwu1:
We should protect mCallbackStatistics with a mutex. The rest of the variables already are atomic

### DIFF
--- a/apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTest.h
+++ b/apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTest.h
@@ -26,6 +26,7 @@
 #include <thread>
 #include <vector>
 #include <unistd.h> // For CPU affinity
+#include <mutex> // Added for mutex
 #include "SynthWorkload.h"
 
 /**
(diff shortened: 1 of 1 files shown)


apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTest.h
```diff
@@ -148,7 +149,10 @@ class AudioWorkloadTest : oboe::AudioStreamDataCallback {
         mAlternateNumVoices = alternateNumVoices;
         mAlternatingPeriodMs = alternatingPeriodMs;
         mStartTimeMs = 0;
-        mCallbackStatistics.clear();
+        {
+            std::lock_guard<std::mutex> lock(mStatisticsMutex);
```
flamme, line 153: nit, just `std::lock_guard lock<mStatisticsMutex>` should work

flamme (commented):
Looks like some of the data can be protected by the same mutex instead of using atomic that may potentially cause wrong report.

Some other atomic values are not needed to be atomic, such as sample rate, which won't change after opening and only set when opening

apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTest.h
robertwu1: every single other mutex in oboe looks like this. Leaving as is

PR #2237, opened 2025-06-26
AudioWorkloadTest: Add locks around the stream

robertwu1:
Add locks around the stream in AudioWorkloadTest

### DIFF
--- a/apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTest.h
+++ b/apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTest.h
@@ -26,7 +26,7 @@
 #include <thread>
 #include <vector>
 #include <unistd.h> // For CPU affinity
-#include <mutex> // Added for mutex
+#include <mutex>
 #include "SynthWorkload.h"
 
 /**
@@ -393,7 +393,7 @@ private:
     std::atomic<bool> mHearWorkload{false};
     std::atomic<bool> mAdpfWorkloadIncreaseEnabled{false};
 
-    // Mutex to protect mCallbackStatistics
+    // Lock to protect mCallbackStatistics
     std::mutex mStatisticsLock;
     std::vector<CallbackStatus> mCallbackStatistics;
 


apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTest.h
```diff
@@ -78,6 +79,11 @@ class AudioWorkloadTest : oboe::AudioStreamDataCallback {
      * @return 0 on success, or a negative Oboe error code on failure.
      */
     int32_t open() {
+        std::lock_guard<std::mutex> lock(mStreamLock);
+        if (mStream) {
+            return 0; // Already open.
```
flamme: Q: why we are not using the oboe error code?

apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTest.h
```diff
@@ -143,6 +149,11 @@ class AudioWorkloadTest : oboe::AudioStreamDataCallback {
     int32_t start(int32_t targetDurationMillis, int32_t numBursts, int32_t numVoices,
                   int32_t alternateNumVoices, int32_t alternatingPeriodMs, bool adpfEnabled,
                   bool adpfWorkloadIncreaseEnabled, bool hearWorkload) {
+        bool expected = false;
+        if (!mRunning.compare_exchange_strong(expected, true)) {
```
flamme: I think the `mRunning` should be better protected by the mutex instead of using atomic so that there won't be race between two start request.

apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTest.h
```diff
@@ -156,17 +167,25 @@ class AudioWorkloadTest : oboe::AudioStreamDataCallback {
...
-        if (result != oboe::Result::OK) {
-            std::cerr << "Error starting stream: " << oboe::convertToText(result) << std::endl;
-            return static_cast<int32_t>(result);
+
+        {
+            std::lock_guard<std::mutex> lock(mStreamLock);
+            if (!mStream) {
+                mRunning = false;
+                return -EINVAL; // Stream not open
+            }
```
flamme: Shouldn't stream nullability be checked at the beginning of this method?

apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTest.h
```diff
@@ -369,6 +404,7 @@ class AudioWorkloadTest : oboe::AudioStreamDataCallback {
     const std::string kTestName = "AudioWorkloadTest";
 
     // Member variables
```
flamme: stale comment?

apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTest.h
```diff
@@ -357,7 +392,7 @@ class AudioWorkloadTest : oboe::AudioStreamDataCallback {
         int64_t currentTimeMs = std::chrono::duration_cast<std::chrono::milliseconds>(
                 std::chrono::high_resolution_clock::now().time_since_epoch()).count();
 
-        if (currentTimeMs - mStartTimeMs > mTargetDurationMs) {
+        if (mTargetDurationMs > 0 && (currentTimeMs - mStartTimeMs > mTargetDurationMs)) {
```
flamme: Should it be better to fail to start if target duration is smaller than 0?

apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTest.h
```diff
@@ -239,8 +258,15 @@ class AudioWorkloadTest : oboe::AudioStreamDataCallback {
      * @return 0 on success, or a negative Oboe error code on failure.
      */
     int32_t stop() {
-        if (mStream) {
-            oboe::Result result = mStream->stop();
+        mRunning = false; // Cooperatively stop the callback.
+        std::shared_ptr<oboe::AudioStream> stream;
```
flamme: What is the reason of copying `mStream` here?

apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTest.h
```diff
@@ -25,8 +25,9 @@
 #include <iostream>
 #include <thread>
 #include <vector>
+#include <cerrno>
```
flamme: There are already places using oboe error code. In that case, why not use oboe error code?

apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTest.h
```diff
@@ -251,12 +277,17 @@ class AudioWorkloadTest : oboe::AudioStreamDataCallback {
 
     /**
      * @brief Closes the audio stream.
-     * @return 0 on success, or a negative Oboe error code on failure.
-     */
+     * @return 0 on success, or a negative Oboe error code on failure. */
     int32_t close() {
-        if (mStream) {
-            oboe::Result result = mStream->close();
+        stop(); // Ensure stream is stopped before closing.
```
flamme: Isn't close will always stop? What is more, there is not lock to guaranteed the stream is stopped after calling `stop`. `start` can always be called after calling `stop`

robertwu1: Addressed all comments by adding a new lock that covers all functions that affect the stream

flamme (approved):

PR #2238, opened 2025-06-26
Remove mResultText in AudioWorkloadTestRunner.h

robertwu1:
Everything else in this file is atomic or locked in #2237

### DIFF
--- a/apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTestRunner.h
+++ b/apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTestRunner.h
@@ -33,7 +33,7 @@
  * Example usage:
  * start();
  * while(!stopIfDone()) { getStatus(); sleep(); }
- * getResultText();
+ * getResult(); getXRunCount();
  *
  */
 class AudioWorkloadTestRunner {
@@ -81,10 +81,6 @@ public:
         }
 
         if (mAudioWorkloadTest.open() != static_cast<int32_t>(oboe::Result::OK)) {
-            {
-                std::lock_guard<std::mutex> lock(mResultTextMutex);
-                mResultText = "Error opening audio stream.";
-            }
             mResult = -1;
             mIsDone = true;
             return -1;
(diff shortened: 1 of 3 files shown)


apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTestRunner.h
```diff
@@ -188,6 +202,8 @@ class AudioWorkloadTestRunner {
     std::atomic<bool> mIsRunning{false};
     std::atomic<bool> mIsDone{true};
     std::atomic<int32_t> mResult{0};
+    // Mutex to protect mResultText. Mutable to allow locking in const methods.
+    mutable std::mutex mResultTextMutex;
```
flamme: If you want to lock in a getter, it is more reasonable to make the getter not const

flamme (commented):
It looks like there are too many locks introduced here. Can we just have one lock to protect everything?

apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTestRunner.h
robertwu1: Done. Thanks!

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/AudioWorkloadTestRunnerActivity.java
```diff
@@ -60,13 +60,13 @@ public void run() {
...
+                int xRunCount = getXRunCount();
                 if (result == 1) {
-                    mResultTextView.setText("Result: PASS\n" + resultText);
+                    mResultTextView.setText("Result: PASS");
                 } else if (result == -1) {
-                    mResultTextView.setText("Result: FAIL\n" + resultText);
+                    mResultTextView.setText("Result: FAIL\n xRunCount: " + xRunCount);
                 } else {
-                    mResultTextView.setText("Result: UNKNOWN\n" + resultText);
+                    mResultTextView.setText("Result: UNKNOWN\n xRunCount: " + xRunCount);
```
flamme: Looks like line 62 to line 69 are the same as line 142 to 149. How about making a function for it?

robertwu1: Done! Nice catch

PR #2241, opened 2025-06-27
Drain offload stream when there is big amount of data written.

### LINKED ISSUE #2240: Drain the stream for TEST OUTPUT when it is offload and using blocking write
For offload, there can be big amount of data written in the buffer, there is no reason to keep writing and keep the buffer full for offload stream. It makes more sense to drain the stream for offload when there are already big enough data written.

flamme:
Fixes #2240.

### DIFF
--- a/apps/OboeTester/app/src/main/cpp/NativeAudioContext.cpp
+++ b/apps/OboeTester/app/src/main/cpp/NativeAudioContext.cpp
@@ -20,6 +20,7 @@
 // We could not trigger the race condition without adding these get calls and the sleeps.
 #define DEBUG_CLOSE_RACE 0
 
+#include <chrono>
 #include <fstream>
 #include <iostream>
 #if DEBUG_CLOSE_RACE
@@ -250,6 +251,7 @@ int ActivityContext::open(jint nativeApi,
         mChannelCount = oboeStream->getChannelCount(); // FIXME store per stream
         mFramesPerBurst = oboeStream->getFramesPerBurst();
         mSampleRate = oboeStream->getSampleRate();
+        mBufferSizeInFrames = oboeStream->getBufferSizeInFrames();
 
         createRecording();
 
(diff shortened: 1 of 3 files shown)


apps/OboeTester/app/src/main/cpp/NativeAudioContext.cpp
```diff
@@ -518,6 +547,21 @@ void ActivityTestOutput::runBlockingIO() {
             LOGE("%s() : write() wrote %d of %d\n", __func__, framesWritten, framesPerBlock);
             break;
         }
+
+        const int64_t bufferSizeInFrames = mBufferSizeInFrames.load();
+        if (oboeStream->getPerformanceMode() == PerformanceMode::POWER_SAVING_OFFLOADED &&
+            bufferSizeInFrames > mSampleRate) {
+            // If it is offload stream, the buffer size is more than 1 second and it is almost full,
```
robertwu1, line 554: where does the 1 second come from? It's not obvious from reading this code

apps/OboeTester/app/src/main/cpp/NativeAudioContext.cpp
```diff
@@ -358,6 +360,33 @@ double ActivityContext::getTimestampLatency(int32_t streamIndex) {
...
 }
 
+void ActivityContext::stopBlockingIOThread() {
+    if (dataThread != nullptr) {
+        // stop a thread that runs in place of the callback
+        threadEnabled.store(false); // ask thread to exit its loop
+        std::shared_ptr<oboe::AudioStream> oboeStream = getOutputStream();
+        if (oboeStream != nullptr &&
+            oboeStream->getPerformanceMode() == PerformanceMode::POWER_SAVING_OFFLOADED) {
+            std::lock_guard _l(threadLock);
```
robertwu1, line 370: Will we get into trouble with both threadEnabled and threadLock? Is there a better way solve this problem with less variables?

flamme: Could you share more context what troubles you are seeing? `threadEnabled` is an atomic value to indicate if the blocking IO thread should be running or not. `threadLock` is a mutex to be used with `threadWorkCV`, which is a condition variable to suspend the blocking IO thread to wait data draining. They are defined for different usage.

apps/OboeTester/app/src/main/cpp/NativeAudioContext.cpp
flamme: As I mentioned offline, this is a naive assumption that the OS won't suspend if there is any CPU work within one second. If the OS is not suspended, the audio framework won't suspend to drain data. There won't be issues to try to write more data. One other reason is that we use one second as the writing time out so that if we care about writing time out, at least one second of buffer size will be a reasonable minimum requirement.

robertwu1: Ohhh, bufferSizeInFrames > mSampleRate is one second...

robertwu1 (approved):

PR #2244, opened 2025-07-11
Update OboeTester and fxlab to Android 36

robertwu1:
This PR updates OboeTester to Android API 36. In Android API 35, there was a [breaking change](https://developer.android.com/about/versions/15/behavior-changes-15#ux) for apps where old apps with navigation bars are just broken in Android API 35. This PR is helping to refactor OboeTester and fxlab to accommodate this. Below are a summary of the changes.

1. In styles.xml, we need to disable the action bar `<style name="AppTheme" parent="Theme.AppCompat.Light.NoActionBar">` because old action bars just don't work correctly. Instead, we can roll out our own action bar.
2. To roll out our own action bar, we can add an linear layout that's around the size of an action bar and add a textbox manually with the activity's title.
3. In whatever top level container we use, we need to add `android:fitsSystemWindows="true"` so the notification icons and the time and the battery icons work correctly.
4. In order to keep colors consistent for Android 34- and Android 35+, we are using light backgrounds and black text for the system windows on all versions of the apps.
5. We need to wrap anything below the action bar in a linear layout or something similar so we can enforce an 16px on all 4 corners. Otherwise, the app looks too crowded. With Android API 35, top level paddings don't seem to work so we need to manually add a container below.
6. DeviceReportActivity has it's share icon changed as we're now rolling out our own action bar.
7. There was an one off raw array issue in NativeAudioContext.cpp that I couldn't compile on Android API 35 without fixing.

### DIFF
--- a/apps/OboeTester/app/src/main/res/layout/activity_dynamic_workload.xml
+++ b/apps/OboeTester/app/src/main/res/layout/activity_dynamic_workload.xml
@@ -1,123 +1,56 @@
-<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
     xmlns:tools="http://schemas.android.com/tools"
-
     android:layout_width="match_parent"
     android:layout_height="match_parent"
-    android:orientation="vertical"
-    android:paddingLeft="@dimen/activity_horizontal_margin"
-    android:paddingTop="@dimen/activity_vertical_margin"
-    android:paddingRight="@dimen/activity_horizontal_margin"
-    android:paddingBottom="@dimen/activity_vertical_margin"
-    tools:context="com.mobileer.oboetester.DynamicWorkloadActivity">
+    android:fitsSystemWindows="true"
+    tools:context=".DynamicWorkloadActivity">
 
-    <include layout="@layout/merge_audio_simple" />
-
-    <TextView
-        android:layout_width="wrap_content"
-        android:layout_height="wrap_content"
-        android:lines="1"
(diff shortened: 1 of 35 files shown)


apps/OboeTester/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.7-all.zip
```
flamme, line 6: I did a quick try with cherry-picking the change to my local. The new UI looks okay to me. The only question for me is that after I cherry-pick, the build asked me to upgrade gradle. And after that, `gradle-wrapper.properties` had some local change.

robertwu1: I ran ./gradlew wrapper --gradle-version 8.14.3. Seems like the new version of the gradle tool no longer adds timestamps

apps/OboeTester/app/src/main/res/layout/activity_dynamic_workload.xml
```diff
@@ -134,46 +119,87 @@
...
+                android:layout_height="wrap_content"
+                android:minHeight="100dp"
+                app:backgroundColor="@color/background_warm"
+                app:exampleDimension="24sp"
+                app:exampleDrawable="@android:drawable/ic_menu_add"
+                app:exampleString="Hello, MultiLineChart"
+                app:lineColor="@color/light_blue_600" />
+        </LinearLayout>
+    </ScrollView>
+</LinearLayout>
```
flamme: missing empty line?

robertwu1: Done! Thanks!

PR #2249, opened 2025-07-17
OboeTester: Add noise pulse for Tap to Tone

### LINKED ISSUE #2247: OboeTester: Use noise pulse instead of saw ping in tap to tone test
From initial testing, when playing a noise pulse instead of a sawtooth wave is much more consistent. I seem to never get a failed test with a noise pulse

robertwu1:
Fixes #2247

### DIFF
--- a/apps/OboeTester/app/src/main/cpp/NativeAudioContext.cpp
+++ b/apps/OboeTester/app/src/main/cpp/NativeAudioContext.cpp
@@ -677,11 +677,18 @@ void ActivityTapToTone::configureAfterOpen() {
     mSinkI32 = std::make_shared<SinkI32>(mChannelCount);
 
     std::shared_ptr<oboe::AudioStream> outputStream = getOutputStream();
-    sawPingGenerator.setSampleRate(outputStream->getSampleRate());
-    sawPingGenerator.frequency.setValue(FREQUENCY_SAW_PING);
-    sawPingGenerator.amplitude.setValue(AMPLITUDE_SAW_PING);
 
-    sawPingGenerator.output.connect(&(monoToMulti->input));
+    mNoisePulseGenerator.amplitude.setValue(AMPLITUDE_SAW_PING);
+    mSawPingGenerator.setSampleRate(outputStream->getSampleRate());
+    mSawPingGenerator.frequency.setValue(FREQUENCY_SAW_PING);
+    mSawPingGenerator.amplitude.setValue(AMPLITUDE_SAW_PING);
+
+    if (mUseNoisePulse) {
+        mNoisePulseGenerator.output.connect(&(monoToMulti->input));
+    } else {
+        mSawPingGenerator.output.connect(&(monoToMulti->input));
+    }
+
     monoToMulti->output.connect(&(mSinkFloat.get()->input));
     monoToMulti->output.connect(&(mSinkI16.get()->input));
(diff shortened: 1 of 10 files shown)


apps/OboeTester/app/src/main/cpp/jni-bridge.cpp
```diff
@@ -994,6 +994,12 @@ Java_com_mobileer_oboetester_TestAudioActivity_setDefaultAudioValues(JNIEnv *env
     oboe::DefaultStreamValues::FramesPerBurst = audio_manager_frames_per_burst;
 }
 
+JNIEXPORT void JNICALL
+Java_com_mobileer_oboetester_TapToToneActivity_useNoisePulse(JNIEnv *env, jclass clazz,
+                                                                     jboolean enabled) {
```
flamme: nit, align with previous line.

apps/OboeTester/app/src/main/cpp/SawPingGenerator.cpp
```diff
@@ -50,7 +51,7 @@ int32_t SawPingGenerator::onProcess(int numFrames) {
     // Check level to prevent numeric underflow.
     if (mLevel > 0.000001) {
         for (int i = 0; i < numFrames; i++) {
```
flamme: If noise is desired, isn't it better to use noise generator instead of changing the SawPingGenerator?

apps/OboeTester/app/src/main/cpp/jni-bridge.cpp
robertwu1: Thanks! Done

apps/OboeTester/app/src/main/cpp/SawPingGenerator.cpp
robertwu1: No, we want to send a signal that starts at 1.0 and gets multiplied by 0.999 every iteration. This gives us a clean signal that we don't accidentally double-count. This is still a saw ping and belongs in this class. It's just that we aren't playing at a note anymore.

flamme: As discussed offline, please make a new class for the noise generator and use it when noise is preferred.

robertwu1: Done

apps/OboeTester/app/src/main/cpp/NativeAudioContext.cpp
```diff
@@ -677,11 +677,18 @@ void ActivityTapToTone::configureAfterOpen() {
     mSinkI32 = std::make_shared<SinkI32>(mChannelCount);
 
     std::shared_ptr<oboe::AudioStream> outputStream = getOutputStream();
-    sawPingGenerator.setSampleRate(outputStream->getSampleRate());
-    sawPingGenerator.frequency.setValue(FREQUENCY_SAW_PING);
-    sawPingGenerator.amplitude.setValue(AMPLITUDE_SAW_PING);
 
-    sawPingGenerator.output.connect(&(monoToMulti->input));
+    mNoisePulseGenerator.amplitude.setValue(AMPLITUDE_SAW_PING);
```
flamme: AMPLITUDE_NOISE_PULSE?

flamme (approved):

PR #2253, opened 2025-07-18
OboeTester: Automated tests should test for MMAP and MMAP shared

robertwu1:
Fixes ##2252

### DIFF
--- a/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/BaseAutoGlitchActivity.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/BaseAutoGlitchActivity.java
@@ -413,9 +413,6 @@ public class BaseAutoGlitchActivity extends GlitchActivity {
         StreamConfiguration requestedInConfig = mAudioInputTester.requestedConfiguration;
         StreamConfiguration requestedOutConfig = mAudioOutTester.requestedConfiguration;
 
-        requestedInConfig.setSharingMode(StreamConfiguration.SHARING_MODE_SHARED);
-        requestedOutConfig.setSharingMode(StreamConfiguration.SHARING_MODE_SHARED);
-
         // Legacy NONE
         requestedInConfig.setMMap(false);
         requestedOutConfig.setMMap(false);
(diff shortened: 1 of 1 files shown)


apps/OboeTester/app/src/main/java/com/mobileer/oboetester/BaseAutoGlitchActivity.java
```diff
@@ -413,6 +414,8 @@ void testPerformancePaths() throws InterruptedException {
         StreamConfiguration requestedInConfig = mAudioInputTester.requestedConfiguration;
         StreamConfiguration requestedOutConfig = mAudioOutTester.requestedConfiguration;
 
+        boolean wasMMapEnabled = NativeEngine.isMMapEnabled();
```
flamme: Shouldn't this can be simply cached by the following code?
```
final boolean requestedInConfigwasMmap = requestedInConfig.isMmap();
final boolean requestedInConfigwasMmap = requestedOutConfig.isMmap();
```

We don't need to know what is the actual mmap enabled or not. What we need is to keep the original requested configuration and request it after the test is completed.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/BaseAutoGlitchActivity.java
```diff
@@ -28,6 +28,7 @@
 
 import java.io.File;
 import java.io.IOException;
+import java.lang.annotation.Native;
```
flamme: Seems no needed?

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/BaseAutoGlitchActivity.java
robertwu1: We don't need this because the only time we look at the requested configs are with testCurrentConfigurations(), which is within this function.

NativeEngine.isMMapEnabled() is good to keep the MMAP state unchanged after the test.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/BaseAutoGlitchActivity.java
robertwu1: Thanks! Done!

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/BaseAutoGlitchActivity.java
flamme: That doesn't sound correct. For instance, if this test is first run, orignially, the global mmap enabled can be by default as `true`. The `requestedInConfig` can either have `isMmap` as true or false. It is also no necessary that the `inConfig` and the `outConfig` has the same `isMmap` value.

robertwu1: Done! Thanks!

flamme (approved):

PR #2255, opened 2025-07-23
Limit timeout between stop and close

### LINKED ISSUE #2254: Limit timeout between stop and close
When audio streams are closed, there is a delay between stop and close to prevent buffer corruption.

This timeout is based on the burst duration. However, for offload streams, one burst is way too much as it could be several seconds. We need to cap this to something like 100ms

~~~
void AudioStream::calculateDefaultDelayBeforeCloseMillis() {
    // Calculate delay time before close based on burst duration.
    // Start with a burst duration then add 1 msec as a safety margin.
    mDelayBeforeCloseMillis = std::max(kMinDelayBeforeCloseMillis,
                                       1 + ((mFramesPerBurst * 1000) / getSampleRate()));
    LOGD("calculateDefaultDelayBeforeCloseMillis() default = %d",
         static_cast<int>(mDelayBeforeCloseMillis));
}
~~~

robertwu1:
Fixes #2254

### DIFF
--- a/src/common/AudioStream.cpp
+++ b/src/common/AudioStream.cpp
@@ -223,8 +223,9 @@ ResultWithValue<FrameTimestamp> AudioStream::getTimestamp(clockid_t clockId) {
 void AudioStream::calculateDefaultDelayBeforeCloseMillis() {
     // Calculate delay time before close based on burst duration.
     // Start with a burst duration then add 1 msec as a safety margin.
-    mDelayBeforeCloseMillis = std::max(kMinDelayBeforeCloseMillis,
-                                       1 + ((mFramesPerBurst * 1000) / getSampleRate()));
+    mDelayBeforeCloseMillis = std::min(kMaxDelayBeforeCloseMillis,
+                                       std::max(kMinDelayBeforeCloseMillis,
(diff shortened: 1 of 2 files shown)


src/common/AudioStream.cpp
```diff
@@ -223,8 +223,9 @@ ResultWithValue<FrameTimestamp> AudioStream::getTimestamp(clockid_t clockId) {
 void AudioStream::calculateDefaultDelayBeforeCloseMillis() {
     // Calculate delay time before close based on burst duration.
     // Start with a burst duration then add 1 msec as a safety margin.
-    mDelayBeforeCloseMillis = std::max(kMinDelayBeforeCloseMillis,
-                                       1 + ((mFramesPerBurst * 1000) / getSampleRate()));
+    mDelayBeforeCloseMillis = std::min(kMaxDelayBeforeCloseMillis,
+                                       std::max(kMinDelayBeforeCloseMillis,
```
flamme: May be simplified by using std::clamp

robertwu1: Done! Thanks!

flamme (approved):

PR #2257, opened 2025-07-24
Samples: Update to Android 36

### LINKED ISSUE #2256: Update Samples and Apps to Android 36
This is needed for apps in the play store by the end of august.

There is a change in android 35 and edge-to-edge enforcement that makes all our apps really ugly on Android 35. All our samples and apps need to be changed so this works properly. See https://developer.android.com/about/versions/15/behavior-changes-15#edge-to-edge

robertwu1:
Fixes #2256


samples/LiveEffect/build.gradle
```diff
@@ -2,11 +2,11 @@ apply plugin: 'com.android.application'
 apply plugin: 'kotlin-android'
 
 android {
-    compileSdkVersion 35
+    compileSdkVersion 36
     defaultConfig {
         applicationId 'com.google.oboe.samples.liveeffect'
-        minSdkVersion 21
-        targetSdkVersion 35
+        minSdkVersion 23
```
flamme, line 8: Why the minSdk is changed?

robertwu1: It's because android:windowLightStatusBar support starts in Android 23. If we don't change it, we will need to fork the xml for the 23+ version and the 22- version. Since Android 22 is already not supported in emulators, I prefer just changing this number to 23

flamme (approved):

PR #2266, opened 2025-07-30
OboeTester: Test reverse jni

### LINKED ISSUE #2265: Test reverse jni
We should add an test in oboetester to test reverse jni. If this works, we may be able to extend Oboe for code that's exclusively non-native through a wrapper

robertwu1:
Fixes #2265

### DIFF
--- a/apps/OboeTester/app/src/main/cpp/ReverseJniEngine.cpp
+++ b/apps/OboeTester/app/src/main/cpp/ReverseJniEngine.cpp
@@ -125,6 +125,7 @@ oboe::DataCallbackResult ReverseJniEngine::onAudioReady(oboe::AudioStream *oboeS
     jsize floatsToCopy = numFrames * mChannelCount;
     env->GetFloatArrayRegion(javaBuffer, 0, floatsToCopy, static_cast<jfloat*>(audioData));
 
+    // Sleep to simulate an actual workload.
     if (mSleepDurationUs > 0) {
         usleep(mSleepDurationUs);
     }
--- a/apps/OboeTester/app/src/main/cpp/ReverseJniEngine.h
+++ b/apps/OboeTester/app/src/main/cpp/ReverseJniEngine.h
@@ -47,7 +47,7 @@ private:
     jobject mJavaObject = nullptr;
     jmethodID mOnAudioReadyId = nullptr;
     std::atomic<bool> mIsThreadAttached{false};
-    std::atomic<int> mSleepDurationUs{false};
+    std::atomic<int> mSleepDurationUs{0};
 
     int mChannelCount;
 


apps/OboeTester/app/src/main/cpp/ReverseJniEngine.cpp
```diff
@@ -0,0 +1,126 @@
...
+    if (env->ExceptionCheck()) {
+        env->ExceptionDescribe();
+        env->ExceptionClear();
+        __android_log_print(ANDROID_LOG_ERROR, TAG, "Exception thrown in Java onAudioReady");
+        return oboe::DataCallbackResult::Stop;
+    }
+
+    // Copy the data from the Java float array to the Oboe audio buffer.
+    jfloat *arrayElements = env->GetFloatArrayElements(audioDataArray, nullptr);
+    memcpy(audioData, arrayElements, static_cast<unsigned long>(numFrames * mAudioStream->getChannelCount() * sizeof(float)));
```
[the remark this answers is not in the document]
flamme: Please fix

apps/OboeTester/app/src/main/res/layout/activity_extra_tests.xml
```diff
@@ -136,5 +136,15 @@
...
+
+        <Button
+            android:id="@+id/buttonReverseJniTest"
+            android:layout_width="0dp"
+            android:layout_height="wrap_content"
+            android:layout_columnWeight="1"
+            android:layout_gravity="fill"
+            android:backgroundTint="@color/button_tint"
+            android:onClick="onLaunchReverseJniTest"
+            android:text="Reverse JNI" />
```
flamme, line 148: nit, can use `title_reverse_jni`?

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/ReverseJniEngine.java
```diff
@@ -0,0 +1,95 @@
+package com.mobileer.oboetester;
```
flamme, line 17: missing copyright

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/ReverseJniEngine.java
```diff
@@ -0,0 +1,95 @@
...
+        Log.i(TAG, "XRun count: " + mXRunCount);
+        try {
+            Thread.sleep(mSleepDurationUs / 1000, (mSleepDurationUs % 1000) * 1000);
+        } catch (InterruptedException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    // Native methods that are implemented in jni-bridge.cpp
+    private native long native_createEngine();
```
flamme: These native methods are not named consistently to other native methods for OboeTester

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/ReverseJniEngine.java
```diff
@@ -0,0 +1,95 @@
...
+            audioData[i] = (float) Math.sin(mPhase);
+            mPhase += mPhaseIncrement;
+            if (mPhase > 2 * Math.PI) {
+                mPhase -= 2 * Math.PI;
+            }
+        }
+        mXRunCount = totalXRunCount;
+        Log.i(TAG, "XRun count: " + mXRunCount);
+        try {
+            Thread.sleep(mSleepDurationUs / 1000, (mSleepDurationUs % 1000) * 1000);
```
flamme: This sleep doesn't look correct. The second parameters should be nanos seconds, right?

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/ReverseJniEngine.java
```diff
@@ -0,0 +1,95 @@
...
+        // Simple sine wave generator
+        for (int i = 0; i < numFrames; i++) {
+            audioData[i] = (float) Math.sin(mPhase);
+            mPhase += mPhaseIncrement;
+            if (mPhase > 2 * Math.PI) {
+                mPhase -= 2 * Math.PI;
+            }
+        }
+        mXRunCount = totalXRunCount;
+        Log.i(TAG, "XRun count: " + mXRunCount);
```
flamme: Log spam

apps/OboeTester/app/src/main/cpp/ReverseJniEngine.cpp
```diff
@@ -0,0 +1,126 @@
...
+    if (mAudioStream) {
+        mAudioStream->close();
+    }
+}
+
+oboe::DataCallbackResult ReverseJniEngine::onAudioReady(oboe::AudioStream *oboeStream,
+                                                        void *audioData,
+                                                        int32_t numFrames) {
+    JNIEnv *env;
+    // Attach the current thread to the JVM to make JNI calls.
```
flamme: Why it is needed to attach current thread to JVM to make JNI calls?

apps/OboeTester/app/src/main/cpp/ReverseJniEngine.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+ReverseJniEngine::~ReverseJniEngine() {
+    // It's important to release the global reference when the engine is destroyed.
+    JNIEnv *env;
+    mJavaVM->GetEnv(reinterpret_cast<void **>(&env), JNI_VERSION_1_6);
+    env->DeleteGlobalRef(mJavaObject);
+}
+
+void ReverseJniEngine::start(int bufferSizeInBursts) {
+    __android_log_print(ANDROID_LOG_INFO, TAG, "Starting audio stream.");
```
flamme: nit, there are debugger helper function LOGI, LOGW...

apps/OboeTester/app/src/main/cpp/ReverseJniEngine.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+ReverseJniEngine::ReverseJniEngine(JNIEnv *env, jobject thiz) {
+    // Store the JavaVM and a global reference to the Java object.
+    env->GetJavaVM(&mJavaVM);
+    mJavaObject = env->NewGlobalRef(thiz);
+
+    // Get the method ID for the onAudioReady callback in the Java class.
+    // The signature "([FII)V" means a method that takes a float array and two ints, and returns void.
+    jclass javaClass = env->GetObjectClass(mJavaObject);
+    mOnAudioReadyId = env->GetMethodID(javaClass, "onAudioReady", "([FII)V");
```
flamme: If this method id is not a global reference, will this be a problem that the local reference is gone?

apps/OboeTester/app/src/main/cpp/ReverseJniEngine.cpp
```diff
@@ -0,0 +1,126 @@
...
+    }
+}
+
+void ReverseJniEngine::setupAudioStream() {
+    oboe::AudioStreamBuilder builder;
+    builder.setDirection(oboe::Direction::Output)
+            ->setPerformanceMode(oboe::PerformanceMode::LowLatency)
+            ->setSharingMode(oboe::SharingMode::Exclusive)
+            ->setFormat(oboe::AudioFormat::Float)
+            ->setChannelCount(oboe::ChannelCount::Mono)
```
flamme: Why mono?

apps/OboeTester/app/src/main/cpp/ReverseJniEngine.cpp
```diff
@@ -0,0 +1,126 @@
...
+    jfloatArray audioDataArray = env->NewFloatArray(numFrames * mAudioStream->getChannelCount());
+
+    // Call the Java method via reverse JNI.
+    env->CallVoidMethod(mJavaObject, mOnAudioReadyId, audioDataArray, numFrames, xRunCount);
+
+    if (env->ExceptionCheck()) {
+        env->ExceptionDescribe();
+        env->ExceptionClear();
+        __android_log_print(ANDROID_LOG_ERROR, TAG, "Exception thrown in Java onAudioReady");
+        return oboe::DataCallbackResult::Stop;
```
flamme, line 121: There can be memory leak given the float array is not released.

apps/OboeTester/app/src/main/cpp/ReverseJniEngine.cpp
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    // Copy the data from the Java float array to the Oboe audio buffer.
+    jfloat *arrayElements = env->GetFloatArrayElements(audioDataArray, nullptr);
+    memcpy(audioData, arrayElements, static_cast<unsigned long>(numFrames) * mAudioStream->getChannelCount() * sizeof(float));
```
flamme: Why unsigned long?

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/ReverseJniEngine.java
```diff
@@ -0,0 +1,95 @@
...
+public class ReverseJniEngine {
+
+    private static final String TAG = "ReverseJniEngine";
+    private long mNativeEngineHandle = 0;
+    private double mPhase = 0.0;
+    private double mPhaseIncrement = 0.0;
+    private final int mSampleRate = 48000;
+    private final double mFrequency = 440.0; // A4 tone
+    private int mXRunCount = 0;
+    private int mSleepDurationUs = 0;
```
flamme: Why sleep during data callback? It will cause underrun.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/ReverseJniEngine.java
```diff
@@ -0,0 +1,95 @@
...
+    public int getXRunCount() {
+        return mXRunCount;
+    }
+
+    public void setSleepDurationUs(int sleepDurationUs) {
+        mSleepDurationUs = sleepDurationUs;
+    }
+
+    @SuppressWarnings("unused") // Called from JNI
+    private void onAudioReady(float[] audioData, int numFrames, int totalXRunCount) {
```
flamme: May need to verify the audioData is null or not?

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/ReverseJniActivity.java
```diff
@@ -0,0 +1,137 @@
...
+
+        mUiUpdateHandler = new Handler(Looper.getMainLooper());
+        mUiUpdateRunnable = new Runnable() {
+            @Override
+            public void run() {
+                if (mIsPlaying && mAudioEngine != null) {
+                    int xruns = mAudioEngine.getXRunCount();
+                    // This must be run on the UI thread
+                    mStatusTextView.setText(String.format("XRuns: %d", xruns));
+                    mUiUpdateHandler.postDelayed(this, 50); // Update every 50ms
```
flamme, line 92: Given the data callback will also update the UI, why it is needed to update every 50ms?

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/ReverseJniActivity.java
```diff
@@ -0,0 +1,137 @@
...
+import android.widget.Button;
+import android.widget.CheckBox;
+import android.widget.RadioButton;
+import android.widget.Spinner;
+import android.widget.TextView;
+
+import androidx.appcompat.app.AppCompatActivity;
+
+/**
+ * Test for reverse jni
```
flamme: Please add more comment about what it is testing here.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/ReverseJniActivity.java
```diff
@@ -0,0 +1,137 @@
...
+import static com.mobileer.oboetester.TestAudioActivity.TAG;
+
+import android.content.Context;
+import android.media.AudioManager;
+import android.os.Bundle;
+import android.os.Handler;
+import android.os.Looper;
+import android.util.Log;
+import android.view.View;
+import android.view.WindowManager;
```
flamme: Unnecessary import? Please double check and remove all unnecessary imports.

apps/OboeTester/app/src/main/res/layout/activity_extra_tests.xml
robertwu1: Leaving as is because some other activities have shortened names.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/ReverseJniEngine.java
robertwu1: Thanks!

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/ReverseJniEngine.java
robertwu1: This is correct. I added some comments. The first param is in milliseconds and the second param is the remaining time in nanoseconds

apps/OboeTester/app/src/main/cpp/ReverseJniEngine.cpp
robertwu1: Gemini:

You need to attach a native thread to the JVM because the JVM operates as a managed, secure environment and needs to be aware of any thread that interacts with it. Think of it as needing an official visitor's pass to enter a secure building.

Here's a breakdown of the key reasons why this "pass" is necessary:

1. To Get a JNIEnv* Pointer
The JNIEnv* is your handle to virtually all JNI functions (NewObject, CallVoidMethod, etc.). This pointer is thread-local, meaning each thread needs its own unique JNIEnv*. The JVM will only grant this pointer to a thread that it knows and manages. Threads created in Java automatically have one, but threads created natively do not.

The AttachCurrentThread function is what asks the JVM to create this JNIEnv* for the calling native thread.

2. Garbage Collection (GC) Safety
The JVM's garbage collector automatically frees up Java objects that are no longer in use. If a native thread is using a Java object without the JVM's knowledge, the GC might mistakenly think the object is unused and delete it. This would cause your native code to access invalid memory, leading to a crash. 💣

When a thread is attached, the JVM tracks all the Java object references the thread is using, ensuring the GC won't touch them.

3. Thread State and Synchronization
The JVM needs to manage a thread's state to handle features core to the Java language:

Exceptions: If your native code calls a Java method that throws an exception, the JVM needs to associate that pending exception with the correct thread.

Synchronization: To use Java's synchronized blocks or methods like Object.wait() and Object.notify(), the thread must be registered with the JVM's thread scheduler and monitor system.

In short, AttachCurrentThread formally introduces a native thread to the JVM, temporarily making it behave like a Java thread. This registration is crucial for the stability, safety, and proper functioning of both the native code and the Java runtime environment. When the thread is done, DetachCurrentThread tells the JVM it's "leaving the building," allowing the JVM to clean up its resources.

apps/OboeTester/app/src/main/cpp/ReverseJniEngine.cpp
robertwu1: Gemini:

No, this will not be a problem. It's a common and correct practice to cache a jmethodID.

The key reason is that a jmethodID is not a Java object reference like a jobject or jclass. It's an opaque pointer to an internal JVM data structure that represents the method itself.

Object References vs. Method IDs
This is a crucial distinction in JNI:

Object References (jobject, jclass, etc.)
These are pointers to actual objects living on the Java heap. They are managed by the Garbage Collector (GC). A local reference is only guaranteed to be valid for the duration of the native method call. If you need to hold onto it longer, you must create a NewGlobalRef(), which tells the GC not to delete the object.

Method and Field IDs (jmethodID, jfieldID)
These are simply identifiers for a class's members (its methods and fields). They are not objects on the heap and are not managed by the GC.

Think of it like this:

A jobject is like a specific book you've checked out from the library. Your loan (the local reference) is temporary.

A jmethodID is like the book's permanent call number (e.g., 813.54) in the library's catalog. The call number itself doesn't expire and remains the same as long as the library (the JVM class) exists.

Lifetime Guarantee
The JNI specification guarantees that a jmethodID is valid until the JVM unloads the class to which it belongs. In Android, class unloading is very rare and generally doesn't happen while your app is running.

Therefore, looking up a method ID once in a constructor and caching it for later use is the standard, safe, and most performant approach.

apps/OboeTester/app/src/main/cpp/ReverseJniEngine.cpp
robertwu1: It's easier to create a sine wave this way, especially when passing params back up. This test is a proof of concept so I'm happy leaving this as is.

apps/OboeTester/app/src/main/cpp/ReverseJniEngine.cpp
robertwu1: Thanks! Done

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/ReverseJniEngine.java
robertwu1: This is done on purpose. This is to simulate an actual workload

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/ReverseJniEngine.java
robertwu1: Added an error log if this is null

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/ReverseJniActivity.java
robertwu1: The data callback doesn't update the UI. We don't want callbacks to update the UI as the UI code is delayed.

We could have the engine code hold a reference to the activity to update the UI but I prefer leaving the code as is

apps/OboeTester/app/src/main/cpp/ReverseJniEngine.cpp
flamme: I thought PoC should be more close to real practice.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/ReverseJniEngine.java
```diff
@@ -0,0 +1,115 @@
...
+        if (audioData == null) {
+            Log.e(TAG, "Audio data is null");
```
flamme: If it is null, it should abort instead of let the app crash

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/ReverseJniEngine.java
flamme: Wow, I missed `mSleepDurationUs` was a microsecond. Just curious, why not just sleep milliseconds? https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.html#sleep-long-

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/ReverseJniActivity.java
flamme: Ack

apps/OboeTester/app/src/main/cpp/ReverseJniEngine.cpp
```diff
@@ -0,0 +1,127 @@
...
+        if (mJavaVM->AttachCurrentThread(&env, nullptr) != 0) {
+            LOGE("Failed to attach current thread to JVM");
+            return oboe::DataCallbackResult::Stop;
+        }
+    }
+
+    int xRunCount = mAudioStream->getXRunCount().value();
+
+    // Create a new Java float array to pass the audio data.
+    jfloatArray audioDataArray = env->NewFloatArray(numFrames * mAudioStream->getChannelCount());
```
flamme: Allocating the array can be expensive if it is done per callback. Could it be faster by set the callback size and allocate the data in JAVA? In this case, every callback is asking for a fixed size data and can be always fulfill by JAVA side and read from callback thread. In that case, there is no need to allocate here.

apps/OboeTester/app/src/main/cpp/ReverseJniEngine.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+    // Copy the data from the Java float array to the Oboe audio buffer.
+    jfloat *arrayElements = env->GetFloatArrayElements(audioDataArray, nullptr);
+    memcpy(audioData, arrayElements, static_cast<unsigned long>(numFrames) * mAudioStream->getChannelCount() * sizeof(float));
+    env->ReleaseFloatArrayElements(audioDataArray, arrayElements, JNI_ABORT);
+    env->DeleteLocalRef(audioDataArray);
+
+    // Detach the thread if we attached it.
+    if (getEnvStat == JNI_EDETACHED) {
+        mJavaVM->DetachCurrentThread();
```
flamme: Attaching and detaching the thread every callback is expensive and affect the performance a lot.

apps/OboeTester/app/src/main/cpp/ReverseJniEngine.cpp
```diff
@@ -0,0 +1,127 @@
...
+    if (env->ExceptionCheck()) {
+        env->ExceptionDescribe();
+        env->ExceptionClear();
+        LOGE("Exception thrown in Java onAudioReady");
+        env->DeleteLocalRef(audioDataArray);
+        return oboe::DataCallbackResult::Stop;
+    }
+
+    // Copy the data from the Java float array to the Oboe audio buffer.
+    jfloat *arrayElements = env->GetFloatArrayElements(audioDataArray, nullptr);
```
flamme: This can be done via GetFloatArrayRegion

apps/OboeTester/app/src/main/cpp/ReverseJniEngine.cpp
flamme: My bad, I forgot this was the native thread.

BTW, this can be done by `mJavaVM->AttachCurrentThread(&env, &args)`

apps/OboeTester/app/src/main/cpp/ReverseJniEngine.cpp
robertwu1: Thanks! Done!

apps/OboeTester/app/src/main/cpp/ReverseJniEngine.cpp
robertwu1: no longer needed with GetFloatArrayRegion

apps/OboeTester/app/src/main/cpp/ReverseJniEngine.cpp
robertwu1: Great suggestion! Done!

apps/OboeTester/app/src/main/cpp/ReverseJniEngine.cpp
robertwu1: This was a great idea! By allocating the data in java, I get no xruns

apps/OboeTester/app/src/main/cpp/ReverseJniEngine.cpp
robertwu1: Done!

apps/OboeTester/app/src/main/cpp/ReverseJniEngine.cpp
robertwu1: Yea, I do that without any arguments

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/ReverseJniEngine.java
robertwu1: I prefer to crash for an early prototype as this helps us find bugs

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/ReverseJniEngine.java
```diff
@@ -0,0 +1,144 @@
...
+     * and returns the index of the buffer that is ready to be read.
+     *
+     * @param numFrames The number of frames the native side needs.
+     * @param totalXRunCount The current x-run count from Oboe.
+     * @return The index (0 or 1) of the buffer that was just filled.
+     */
+    @SuppressWarnings("unused") // Called from JNI
+    private int onAudioReady(int numFrames, int totalXRunCount) {
+        int writeBufferIndex = mCurrentBufferIndex;
+        float[] writeBuffer = mAudioBuffers[writeBufferIndex];
```
flamme: Why using two audio buffer? The buffer is only synchronized access by the JAVA and native.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/ReverseJniEngine.java
flamme: Unresolving for question.

robertwu1: Because if we have 2ms buffers, we want more specificity

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/ReverseJniEngine.java
robertwu1: Removed the double buffer. Seems fine without it

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/ReverseJniEngine.java
robertwu1: Changed to sleep in native after discussing how Thread.sleep doesn't sleep at a very precisely

apps/OboeTester/app/src/main/cpp/ReverseJniEngine.h
```diff
@@ -46,6 +47,7 @@ class ReverseJniEngine : public oboe::AudioStreamDataCallback, public oboe::Audi
     jobject mJavaObject = nullptr;
     jmethodID mOnAudioReadyId = nullptr;
     std::atomic<bool> mIsThreadAttached{false};
+    std::atomic<int> mSleepDurationUs{false};
```
flamme: Should not be initialized with false

apps/OboeTester/app/src/main/cpp/ReverseJniEngine.cpp
```diff
@@ -120,6 +125,10 @@ oboe::DataCallbackResult ReverseJniEngine::onAudioReady(oboe::AudioStream *oboeS
     jsize floatsToCopy = numFrames * mChannelCount;
     env->GetFloatArrayRegion(javaBuffer, 0, floatsToCopy, static_cast<jfloat*>(audioData));
 
+    if (mSleepDurationUs > 0) {
+        usleep(mSleepDurationUs);
```
flamme, line 130: nit, add a comment about why sleep as the data callback is not supposed to sleep.

apps/OboeTester/app/src/main/cpp/ReverseJniEngine.h
robertwu1: Thanks! Done!

apps/OboeTester/app/src/main/cpp/ReverseJniEngine.cpp
robertwu1: Thanks! Done!

flamme (approved):

PR #2267, opened 2025-07-30
OboeTester: optimize AudioWorkloadTest.h

### LINKED ISSUE #2258: Race condition in AudioWorkloadTrace.h
Internal bug b/424242729

robertwu1:
This change several changes to make AudioWorkloadTest run better and have less crashes.

1. SynthWorkload is now created with an unique_ptr to reduce the number of crashes as SynthWorkload uses new on an array.
2. IsRunning is set to false when the stream is close/stopped.
3. Added more checks in open and start to make sure the system is in the right place when the functions are called.
4. We added waitForStateChange for stop to make sure that the system indeeds stops.
5. LOGE is used instead of std::cerr to not have blocking logs.

Fixes #2258

### DIFF
--- a/apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTest.h
+++ b/apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTest.h
@@ -79,6 +79,11 @@ public:
      */
     int32_t open() {
         std::lock_guard<std::mutex> lock(mStreamLock);
+        if (mStream) {
+            std::cerr << "Error: Stream already open." << std::endl;
+            return static_cast<int32_t>(oboe::Result::ErrorUnavailable);
+        }
+
         oboe::AudioStreamBuilder builder;
         builder.setDirection(oboe::Direction::Output);
         builder.setPerformanceMode(oboe::PerformanceMode::LowLatency);
@@ -98,6 +103,7 @@ public:
         mPreviousXRunCount = 0;
         mXRunCount = 0;
         mPhaseIncrement = 2.0f * (float) M_PI * 440.0f / mSampleRate; // 440 Hz sine wave
+        mSynthWorkload = std::make_unique<SynthWorkload>((int) 0.2 * mSampleRate, (int) 0.3 * mSampleRate);
 
         return 0;
     }
@@ -172,12 +178,9 @@ public:
         mStream->setPerformanceHintEnabled(adpfEnabled);
         mStream->setBufferSizeInFrames(mNumBursts * mFramesPerBurst);
         mBufferSizeInFrames = mStream->getBufferSizeInFrames();
-        {
-            std::lock_guard<std::mutex> synthWorkloadLock(mSynthWorkloadLock);
-            mSynthWorkload = std::make_unique<SynthWorkload>((int) 0.2 * mSampleRate, (int) 0.3 * mSampleRate);
-        }
         oboe::Result result = mStream->start();
         if (result != oboe::Result::OK) {
+            mRunning = false;
             std::cerr << "Error starting stream: " << oboe::convertToText(result) << std::endl;
             return static_cast<int32_t>(result);
         }
@@ -344,16 +347,13 @@ public:
             if (mPhase >= M_PI) mPhase = mPhase - 2.0f * M_PI;
         }
 
-        {
-            std::lock_guard<std::mutex> synthWorkloadLock(mSynthWorkloadLock);
-            if (mSynthWorkload) {
-                mSynthWorkload->onCallback(currentVoices);
-                if (currentVoices > 0) {
-                    // Render synth workload into the buffer or discard the synth voices.
-                    float *buffer = (audioStream->getChannelCount() == 2 && mHearWorkload)
-                                    ? static_cast<float *>(audioData) : nullptr;
-                    mSynthWorkload->renderStereo(buffer, numFrames);
-                }
+        if (mSynthWorkload) {
+            mSynthWorkload->onCallback(currentVoices);
+            if (currentVoices > 0) {
+                // Render synth workload into the buffer or discard the synth voices.
+                float *buffer = (audioStream->getChannelCount() == 2 && mHearWorkload)
+                                ? static_cast<float *>(audioData) : nullptr;
+                mSynthWorkload->renderStereo(buffer, numFrames);
             }
         }
 
@@ -421,8 +421,6 @@ private:
     std::atomic<float> mPhase{0.0f};           // Current phase of the sine wave oscillator
     std::atomic<float> mPhaseIncrement{0.0f};  // Phase increment for sine wave
 
-    // Lock to protect mSynthWorkload
-    std::mutex mSynthWorkloadLock;
     std::unique_ptr<SynthWorkload> mSynthWorkload; // Pointer to the synthetic workload generator
 };
 
--- a/apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTest.h
+++ b/apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTest.h
@@ -149,6 +149,10 @@ public:
             std::cerr << "Error: Stream not open." << std::endl;
             return static_cast<int32_t>(oboe::Result::ErrorInvalidState);
         }
+        if (mRunning) {
+            std::cerr << "Error: Stream already started." << std::endl;
+            return static_cast<int32_t>(oboe::Result::ErrorUnavailable);
+        }
         mTargetDurationMs = targetDurationMillis;
         mNumBursts = numBursts;
         mNumVoices = numVoices;
@@ -256,6 +260,7 @@ public:
                 return static_cast<int32_t>(result);
             }
         }
+        mRunning = false;
         return 0;
     }
 
@@ -273,6 +278,7 @@ public:
                 return static_cast<int32_t>(result);
             }
         }
+        mRunning = false;
         return 0;
     }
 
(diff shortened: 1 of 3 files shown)


apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTest.h
```diff
@@ -168,7 +168,10 @@ class AudioWorkloadTest : oboe::AudioStreamDataCallback {
         mStream->setPerformanceHintEnabled(adpfEnabled);
         mStream->setBufferSizeInFrames(mNumBursts * mFramesPerBurst);
         mBufferSizeInFrames = mStream->getBufferSizeInFrames();
-        mSynthWorkload = SynthWorkload((int) 0.2 * mSampleRate, (int) 0.3 * mSampleRate);
```
robertwu1, line 171: SynthWorkload was initialized before in this class's constructor and here and this caused bugs as SynthWorkload has a new in it's constructor.

flamme: It looks like the problem is that you are using assignment operation while there is not a correct override for assignment operator in SynthWorkload.

apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTest.h
```diff
@@ -407,7 +415,9 @@ class AudioWorkloadTest : oboe::AudioStreamDataCallback {
     std::atomic<float> mPhase{0.0f};           // Current phase of the sine wave oscillator
     std::atomic<float> mPhaseIncrement{0.0f};  // Phase increment for sine wave
 
-    SynthWorkload mSynthWorkload;              // Instance of the synthetic workload generator
+    // Lock to protect mSynthWorkload
+    std::mutex mSynthWorkloadLock;
```
flamme: The lock is not needed. The onAudioReady will not be called before calling start.

Instead, from the start function, it is better to return earlier when the stream is already started.

apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTest.h
```diff
@@ -407,7 +415,9 @@ class AudioWorkloadTest : oboe::AudioStreamDataCallback {
...
+    std::shared_ptr<SynthWorkload> mSynthWorkload; // Pointer to the synthetic workload generator
```
flamme: This can be unique_ptr as it is not shared by anyone else.

apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTest.h
robertwu1: Great idea with the mRunning checks. I added it.

I still think having a lock on this is good for best practices as this is on another thread

flamme: I agreed that locking is good for multi-thread competing same resource. But this doesn't have the resource competition between the callback thread and start.

robertwu1: It might because the user can manually stop the test vs the test being stopped from the callback thread. Because of this, I prefer leaving the lock

apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTest.h
```diff
@@ -168,7 +172,10 @@ class AudioWorkloadTest : oboe::AudioStreamDataCallback {
...
         mStream->setBufferSizeInFrames(mNumBursts * mFramesPerBurst);
         mBufferSizeInFrames = mStream->getBufferSizeInFrames();
-        mSynthWorkload = SynthWorkload((int) 0.2 * mSampleRate, (int) 0.3 * mSampleRate);
+        {
+            std::lock_guard<std::mutex> synthWorkloadLock(mSynthWorkloadLock);
+            mSynthWorkload = std::make_unique<SynthWorkload>((int) 0.2 * mSampleRate, (int) 0.3 * mSampleRate);
+        }
         oboe::Result result = mStream->start();
         if (result != oboe::Result::OK) {
             std::cerr << "Error starting stream: " << oboe::convertToText(result) << std::endl;
```
flamme: Need to set mRunning to false when there it fails to start

apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTest.h
flamme: That seems not make sense. No matter it is user manually stopping it or returning stopped from the callback thread, it doesn't access and change `mSynthWorkload`. `mSynthWorkload` is only accessed by onAudioReady, which is the callback thread and modified by `start` method.

The one that may caused problem is `mRunning` is accessed and modified by multiple threads. Especially in the callback thread, it is changed without any lock protecting. Even if it is atomic, it causes problem as it is not guaranteed the result.

For `mSynthWorkload`, why it cannot be allocated after the stream is opened given only sample rate is needed.

apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTest.h
robertwu1: Done! Thanks!

apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTest.h
robertwu1: Thanks! I added a check to mStream null check in open and removed the lock

apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTest.h
```diff
@@ -335,12 +347,14 @@ class AudioWorkloadTest : oboe::AudioStreamDataCallback {
             if (mPhase >= M_PI) mPhase = mPhase - 2.0f * M_PI;
         }
 
-        mSynthWorkload.onCallback(currentVoices);
-        if (currentVoices > 0) {
-            // Render synth workload into the buffer or discard the synth voices.
-            float *buffer = (audioStream->getChannelCount() == 2 && mHearWorkload)
-                            ? static_cast<float *>(audioData) : nullptr;
-            mSynthWorkload.renderStereo(buffer, numFrames);
+        if (mSynthWorkload) {
```
flamme, line 360: If mSynthWorkload is null, does it affect the test result? If so, it may be good to abort the test earlier if mSynthWorkload is null or return error immediately when opening and close the stream.

flamme (commented):
`mRunning` value may be a problem especially when running automation test. Imaging the test is on legacy mode and there can be one more callback after calling stop. When it starts, startTimeMs is reset to 0 and mRunning is set to true. When the callback arrives, it may change `mRunning` to false due to timeout.

apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTest.h
robertwu1: I thinking leaving this alone is fine for now

robertwu1:
I added waitForStateChange for the stop() function

apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTest.h
```diff
@@ -247,12 +257,23 @@ class AudioWorkloadTest : oboe::AudioStreamDataCallback {
...
         std::lock_guard<std::mutex> lock(mStreamLock);
         if (mStream) {
-            oboe::Result result = mStream->stop();
+            oboe::Result result = mStream->requestStop();
             if (result != oboe::Result::OK) {
                 std::cerr << "Error stopping stream: " << oboe::convertToText(result) << std::endl;
                 return static_cast<int32_t>(result);
             }
+            oboe::StreamState next;
+            result = mStream->waitForStateChange(oboe::StreamState::Stopping, &next, 0);
```
flamme: timeoutNanos as 0 means it returns immediately.

robertwu1: Added a timeout of 500ms. Thanks!

apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTest.h
```diff
@@ -247,12 +257,23 @@ class AudioWorkloadTest : oboe::AudioStreamDataCallback {
...
-            oboe::Result result = mStream->stop();
+            oboe::Result result = mStream->requestStop();
             if (result != oboe::Result::OK) {
                 std::cerr << "Error stopping stream: " << oboe::convertToText(result) << std::endl;
                 return static_cast<int32_t>(result);
             }
+            oboe::StreamState next;
+            result = mStream->waitForStateChange(oboe::StreamState::Stopping, &next, kTimeoutInNanos);
+            if (result != oboe::Result::OK) {
+                std::cerr << "Error while waiting for stream to stop: " << oboe::convertToText(result) << std::endl;
```
flamme: I have questions but forgot to ask about why use `std::cerr` instead of `LOGE`?

apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTest.h
flamme: ack

apps/OboeTester/app/src/main/cpp/cpu/AudioWorkloadTest.h
robertwu1: Thanks! This is really helpful as LOGE is non-blocking while std::cerr is blocking

flamme (approved):

flamme (commented):
Please don't forget to update the commit message when submitting. This change doesn't add lock any more.

PR #2282, opened 2025-08-11
OboeTester: Add audio focus

### LINKED ISSUE #2277: Add AudioFocus to OboeTester
We should add AudioFocus to OboeTester to mimic real apps

robertwu1:
Fixes #2277

### DIFF
--- a/apps/OboeTester/app/src/main/cpp/NativeAudioContext.h
+++ b/apps/OboeTester/app/src/main/cpp/NativeAudioContext.h
@@ -300,6 +300,8 @@ public:
 
     virtual void setAmplitude(float amplitude) {}
 
+    virtual void setDuck(bool isDucked) {}
+
     virtual int32_t saveWaveFile(const char *filename);
 
     virtual void setMinimumFramesBeforeRead(int32_t numFrames) {}
@@ -470,8 +472,13 @@ public:
     void setAmplitude(float amplitude) override {
         mAmplitude = amplitude;
         if (mVolumeRamp) {
-            mVolumeRamp->setTarget(mAmplitude);
+            mVolumeRamp->setTarget(mAmplitude * mDuckingMultiplier);
         }
+    }
+
+    void setDuck(bool isDucked) override {
+        mDuckingMultiplier = isDucked ? 0.2f : 1.0f;
+        setAmplitude(mAmplitude);
     }
 
     void setupMemoryBuffer(std::unique_ptr<uint8_t[]>& buffer, int length) final;
@@ -493,6 +500,7 @@ protected:
 
     static constexpr int             kRampMSec = 10; // for volume control
     float                            mAmplitude = 1.0f;
+    float                            mDuckingMultiplier = 1.0f;
     std::shared_ptr<RampLinear> mVolumeRamp;
 
     std::unique_ptr<ManyToMultiConverter>   manyToMulti;
@@ -840,6 +848,12 @@ public:
 
     void setDelayTime(double delayTimeMillis) {
         mActivityEcho.setDelayTime(delayTimeMillis);
+    }
+
+    void setDuck(bool isDucked) {
+        if (currentActivity == &mActivityTestOutput) {
+            mActivityTestOutput.setDuck(isDucked);
+        }
     }
 
     ActivityTestOutput           mActivityTestOutput;
--- a/apps/OboeTester/app/src/main/cpp/jni-bridge.cpp
+++ b/apps/OboeTester/app/src/main/cpp/jni-bridge.cpp
@@ -1002,6 +1002,11 @@ Java_com_mobileer_oboetester_TestAudioActivity_setDefaultAudioValues(JNIEnv *env
 }
 
 JNIEXPORT void JNICALL
+Java_com_mobileer_oboetester_TestAudioActivity_setDuck(JNIEnv *env, jobject thiz, jboolean isDucked) {
+    engine.setDuck(isDucked);
+}
+
+JNIEXPORT void JNICALL
 Java_com_mobileer_oboetester_TapToToneActivity_useNoisePulse(JNIEnv *env,
                                                              jclass clazz,
                                                              jboolean enabled) {
--- a/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/StreamConfiguration.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/StreamConfiguration.java
@@ -16,6 +16,8 @@
 
 package com.mobileer.oboetester;
 
+import android.media.AudioAttributes;
+import android.os.Build;
 import android.text.TextUtils;
 
 import androidx.annotation.Nullable;
@@ -483,6 +485,42 @@ public class StreamConfiguration {
         }
     }
 
+    static int convertUsageToAudioAttributeUsage(int usage) {
+        switch(usage) {
+            case UNSPECIFIED:
+                return AudioAttributes.USAGE_UNKNOWN;
+            case USAGE_MEDIA:
+                return AudioAttributes.USAGE_MEDIA;
+            case USAGE_VOICE_COMMUNICATION:
+                return AudioAttributes.USAGE_VOICE_COMMUNICATION;
+            case USAGE_VOICE_COMMUNICATION_SIGNALLING:
+                return AudioAttributes.USAGE_VOICE_COMMUNICATION_SIGNALLING;
+            case USAGE_ALARM:
+                return AudioAttributes.USAGE_ALARM;
+            case USAGE_NOTIFICATION:
+                return AudioAttributes.USAGE_NOTIFICATION;
+            case USAGE_NOTIFICATION_RINGTONE:
+                return AudioAttributes.USAGE_NOTIFICATION_RINGTONE;
+            case USAGE_NOTIFICATION_EVENT:
+                return AudioAttributes.USAGE_NOTIFICATION_EVENT;
+            case USAGE_ASSISTANCE_ACCESSIBILITY:
+                return AudioAttributes.USAGE_ASSISTANCE_ACCESSIBILITY;
+            case USAGE_ASSISTANCE_NAVIGATION_GUIDANCE:
+                return AudioAttributes.USAGE_ASSISTANCE_NAVIGATION_GUIDANCE;
+            case USAGE_ASSISTANCE_SONIFICATION:
+                return AudioAttributes.USAGE_ASSISTANCE_SONIFICATION;
+            case USAGE_GAME:
+                return AudioAttributes.USAGE_GAME;
+            case USAGE_ASSISTANT:
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+                    return AudioAttributes.USAGE_ASSISTANT;
+                }
+                return AudioAttributes.USAGE_UNKNOWN;
+            default:
+                return AudioAttributes.USAGE_UNKNOWN;
+        }
+    }
+
     public static int convertTextToUsage(String text) {
         return mUsageStringToIntegerMap.get(text);
     }
@@ -501,6 +539,23 @@ public class StreamConfiguration {
                 return "Sonification";
             default:
                 return "?=" + contentType;
+        }
+    }
+
+    static int convertContentTypeAudioAttributesContentType(int contentType) {
+        switch(contentType) {
+            case UNSPECIFIED:
+                return AudioAttributes.CONTENT_TYPE_UNKNOWN;
+            case CONTENT_TYPE_SPEECH:
+                return AudioAttributes.CONTENT_TYPE_SPEECH;
+            case CONTENT_TYPE_MUSIC:
+                return AudioAttributes.CONTENT_TYPE_MUSIC;
+            case CONTENT_TYPE_MOVIE:
+                return AudioAttributes.CONTENT_TYPE_MOVIE;
+            case CONTENT_TYPE_SONIFICATION:
+                return AudioAttributes.CONTENT_TYPE_SONIFICATION;
+            default:
+                return AudioAttributes.CONTENT_TYPE_UNKNOWN;
         }
     }
 
--- a/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
@@ -19,7 +19,9 @@ package com.mobileer.oboetester;
 import static com.mobileer.oboetester.AudioForegroundService.ACTION_START;
 import static com.mobileer.oboetester.AudioForegroundService.ACTION_STOP;
 import static com.mobileer.oboetester.IntentBasedTestSupport.KEY_RESTART_STREAM_IF_CLOSED;
+import static com.mobileer.oboetester.StreamConfiguration.convertContentTypeAudioAttributesContentType;
 import static com.mobileer.oboetester.StreamConfiguration.convertErrorToText;
+import static com.mobileer.oboetester.StreamConfiguration.convertUsageToAudioAttributeUsage;
 
 import android.content.Context;
 import android.content.Intent;
@@ -28,6 +30,7 @@ import android.content.pm.PackageManager;
 import android.content.pm.ServiceInfo;
 import android.media.AudioAttributes;
 import android.media.AudioDeviceInfo;
+import android.media.AudioFocusRequest;
 import android.media.AudioManager;
 import android.os.Build;
 import android.os.Bundle;
@@ -56,7 +59,7 @@ import java.util.Locale;
 /**
  * Base class for other Activities.
  */
-abstract class TestAudioActivity extends AppCompatActivity {
+abstract class TestAudioActivity extends AppCompatActivity implements AudioManager.OnAudioFocusChangeListener {
     public static final String TAG = "OboeTester";
 
     protected static final int FADER_PROGRESS_MAX = 1000;
@@ -113,6 +116,8 @@ abstract class TestAudioActivity extends AppCompatActivity {
     private static boolean mBackgroundEnabled;
     private static boolean mForegroundServiceEnabled;
     private static boolean mRestartStreamIfClosed = false;
+    private static boolean mAudioFocusEnabled;
+    private boolean mIsDucked = false;
 
     protected Bundle mBundleFromIntent;
     protected boolean mTestRunningByIntent;
@@ -216,6 +221,14 @@ abstract class TestAudioActivity extends AppCompatActivity {
 
     public static boolean isForegroundServiceEnabled() {
         return mForegroundServiceEnabled;
+    }
+
+    public static void setAudioFocusEnabled(boolean enabled) {
+        mAudioFocusEnabled = enabled;
+    }
+
+    public static boolean isAudioFocusEnabled() {
+        return mAudioFocusEnabled;
     }
 
     public int getServiceType() {
@@ -655,11 +668,15 @@ abstract class TestAudioActivity extends AppCompatActivity {
     public void openAudio() throws IOException {
         closeAudio();
 
-        updateNativeAudioParameters();
-
         if (!isTestConfiguredUsingBundle()) {
             applyConfigurationViewsToModels();
         }
+
+        if (isAudioFocusEnabled()) {
+            requestAudioFocus();
+        }
+
+        updateNativeAudioParameters();
 
         int sampleRate = 0; // Use the OUTPUT sample rate for INPUT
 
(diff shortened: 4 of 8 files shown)


flamme:
Please resolve the conflict

apps/OboeTester/docs/AutomatedTesting.md
```diff
@@ -80,6 +80,7 @@ There are some optional parameter in common for all tests:
                             Stream type for the setStreamVolume() call. Default is "music".
     --ez background         {"true", 1, "false", 0} // if true then Oboetester will continue to run in the background
     --ez foreground_service {"true", 1, "false", 0} // if true then Oboetester will ask for record/play permissions via a foreground service
+    --ez audio_focus        {"true", 1, "false", 0} // if true then Oboetester will request for an exclusive stream via audio focus
```
flamme: Why it is for exclusive stream only?

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
```diff
@@ -988,4 +1100,4 @@ public AudioConfig(int sampleRate, int format, int channelMask, int channelCount
             mChannelCount = channelCount;
         }
     }
-}
+}
```
flamme: missing empty line at the end of file.

apps/OboeTester/app/src/main/res/layout/activity_main.xml
```diff
@@ -226,6 +226,14 @@
             android:checked="false"
             android:text="enable foreground service" />
 
+        <CheckBox
+            android:id="@+id/boxEnableAudioFocus"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_marginTop="6dp"
+            android:checked="false"
```
flamme: Can we by default use audio focus given that is the right behavior as an app? As a test app, it is okay to test misbehavior, but it will be better to by default doing the right thing.

apps/OboeTester/app/src/main/cpp/jni-bridge.cpp
```diff
@@ -1373,4 +1378,4 @@ Java_com_mobileer_oboetester_AudioWorkloadTestRunnerActivity_getXRunCount(JNIEnv
     return sAudioWorkloadRunner.getXRunCount();
 }
 
-} // extern "C"
+} // extern "C"
```
flamme: Missing empty line

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
```diff
@@ -700,6 +717,94 @@ public void openAudio() throws IOException {
...
+    private void requestAudioFocus() {
+        // Get attributes for focus request.
+        AudioAttributes.Builder attributesBuilder = new AudioAttributes.Builder();
+        // Default to media usage.
+        attributesBuilder.setUsage(AudioAttributes.USAGE_MEDIA);
+        attributesBuilder.setContentType(AudioAttributes.CONTENT_TYPE_MUSIC);
+
+        StreamContext firstOutputStreamContext = getFirstOutputStreamContext();
+        if (firstOutputStreamContext == null) {
+            firstOutputStreamContext = getFirstInputStreamContext();
```
flamme: This is so confused to set output stream context to input stream context. It will be better to renamed or explain the reason here.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
```diff
@@ -700,6 +717,94 @@ public void openAudio() throws IOException {
         mStreamSniffer.startStreamSniffer();
     }
 
+    private void requestAudioFocus() {
+        // Get attributes for focus request.
```
flamme: Maybe get rid of the unnecessary comment from AI

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
```diff
@@ -700,6 +717,94 @@ public void openAudio() throws IOException {
...
+        AudioAttributes.Builder attributesBuilder = new AudioAttributes.Builder();
+        // Default to media usage.
+        attributesBuilder.setUsage(AudioAttributes.USAGE_MEDIA);
+        attributesBuilder.setContentType(AudioAttributes.CONTENT_TYPE_MUSIC);
```
flamme: This two lines can be merged to previous line.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
```diff
@@ -700,6 +717,94 @@ public void openAudio() throws IOException {
...
+                    .setOnAudioFocusChangeListener(this)
+                    .build();
+            int result = am.requestAudioFocus(focusRequest);
+            if (result != AudioManager.AUDIOFOCUS_REQUEST_GRANTED) {
+                showErrorToast("Could not get audio focus");
+                return;
+            }
+        } else {
+            // Use deprecated method for older APIs.
+            int result = am.requestAudioFocus(this, AudioManager.STREAM_MUSIC,
```
flamme: Why only use STREAM_MUSIC for older platform?

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
```diff
@@ -700,6 +717,94 @@ public void openAudio() throws IOException {
...
+                    AudioManager.AUDIOFOCUS_GAIN);
+            if (result != AudioManager.AUDIOFOCUS_REQUEST_GRANTED) {
+                showErrorToast("Could not get audio focus");
+                return;
+            }
+        }
+    }
+
+    private void duck() {
+        if (!mIsDucked) {
```
flamme: It looks like caching `mIsDucked` is not necessary.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
```diff
@@ -700,6 +717,94 @@ public void openAudio() throws IOException {
...
+    }
+
+    @Override
+    public void onAudioFocusChange(int focusChange) {
+        runOnUiThread(() -> {
+            Log.i(TAG, "onAudioFocusChange(" + focusChange + ")");
+            switch (focusChange) {
+                case AudioManager.AUDIOFOCUS_GAIN:
+                    unduck();
+                    if (mAudioState == AUDIO_STATE_STOPPED || mAudioState == AUDIO_STATE_PAUSED) {
```
flamme: Accessing `mAudioState` without locking can cause unexpected behavior as this value can be updated from other thread.

apps/OboeTester/docs/AutomatedTesting.md
robertwu1: Changed the messaging. I wanted to say that audio focus is for getting exclusive access to output but it's hard to explain

apps/OboeTester/app/src/main/res/layout/activity_main.xml
robertwu1: Good idea! Thanks!

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
robertwu1: Changed to streamContext

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
robertwu1: Thanks!

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
robertwu1: Done! Thanks!

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
robertwu1: I added a converter. Thanks!

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
robertwu1: Thanks! Good catch!

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
robertwu1: I added a lock. Let me know what you think!

apps/OboeTester/app/src/main/cpp/NativeAudioContext.h
```diff
@@ -478,10 +480,15 @@ class ActivityTestOutput : public ActivityContext {
     void setAmplitude(float amplitude) override {
         mAmplitude = amplitude;
         if (mVolumeRamp) {
-            mVolumeRamp->setTarget(mAmplitude);
+            mVolumeRamp->setTarget(mAmplitude * mDuckingMultiplier);
         }
     }
 
+    void setDuck(bool isDucked) override {
+        mDuckingMultiplier = isDucked ? 0.2f : 1.0f;
```
flamme: Please define variables for magic numbers

apps/OboeTester/docs/AutomatedTesting.md
```diff
@@ -80,6 +80,7 @@ There are some optional parameter in common for all tests:
                             Stream type for the setStreamVolume() call. Default is "music".
     --ez background         {"true", 1, "false", 0} // if true then Oboetester will continue to run in the background
     --ez foreground_service {"true", 1, "false", 0} // if true then Oboetester will ask for record/play permissions via a foreground service
+    --ez audio_focus        {"true", 1, "false", 0} // if true then Oboetester will request for audio focus
```
flamme: Should also mention the default value

apps/OboeTester/app/src/main/cpp/NativeAudioContext.h
```diff
@@ -854,6 +862,12 @@ class NativeAudioContext {
         mActivityEcho.setDelayTime(delayTimeMillis);
     }
 
+    void setDuck(bool isDucked) {
```
flamme: This doesn't look like consistent to other methods.

If we only preferred to setting duck to test output, it should be a JNI function to TestAudioOutputActivity or call getCurrentActivity->setDuck and have different implementation in different activity

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/StreamConfiguration.java
```diff
@@ -483,6 +486,69 @@ static String convertUsageToText(int usage) {
         }
     }
 
+    static int convertUsageToAudioAttributeUsage(int usage) {
+        switch(usage) {
+            case UNSPECIFIED:
```
flamme: How about move line 491 to before `default`?

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/StreamConfiguration.java
```diff
@@ -504,6 +570,23 @@ static String convertContentTypeToText(int contentType) {
         }
     }
 
+    static int convertContentTypeAudioAttributesContentType(int contentType) {
+        switch(contentType) {
+            case UNSPECIFIED:
```
flamme: Move UNSPECIFIED to before default

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
```diff
@@ -409,20 +427,30 @@ public void enableForegroundService(boolean enabled) {
...
-            mPauseButton.setBackgroundColor(mAudioState == AUDIO_STATE_PAUSED ? COLOR_ACTIVE : COLOR_IDLE);
-            mFlushButton.setBackgroundColor(mAudioState == AUDIO_STATE_FLUSHED ? COLOR_ACTIVE : COLOR_IDLE);
-            mStopButton.setBackgroundColor(mAudioState == AUDIO_STATE_STOPPED ? COLOR_ACTIVE : COLOR_IDLE);
-            mReleaseButton.setBackgroundColor(mAudioState == AUDIO_STATE_RELEASED ? COLOR_ACTIVE : COLOR_IDLE);
-            mCloseButton.setBackgroundColor(mAudioState == AUDIO_STATE_CLOSED ? COLOR_ACTIVE : COLOR_IDLE);
-        }
-        setConfigViewsEnabled(mAudioState == AUDIO_STATE_CLOSED);
+        // This method is called from state-changing methods.
+        // It must run on the UI thread because it touches Views.
+        runOnUiThread(() -> {
```
flamme: If this is required, please do in a separated PR instead of in the same PR of adding audio focus.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
```diff
@@ -657,53 +685,123 @@ public int getSampleRate() {
...
 
-        if (!isTestConfiguredUsingBundle()) {
-            applyConfigurationViewsToModels();
-        }
+            if (isAudioFocusEnabled() && (getFirstOutputStreamContext() != null)) {
+                requestAudioFocus();
+            }
 
-        int sampleRate = 0; // Use the OUTPUT sample rate for INPUT
+            updateNativeAudioParameters();
```
flamme: Why updating native audio parameters is moving to  after applying configuration view?

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
```diff
@@ -657,53 +685,123 @@ public int getSampleRate() {
...
-                openStreamContext(streamContext);
             }
         }
         updateEnabledWidgets();
         onStartAllContexts();
         mStreamSniffer.startStreamSniffer();
     }
 
+    private void requestAudioFocus() {
+        AudioAttributes.Builder attributesBuilder = new AudioAttributes.Builder()
```
flamme: This should be initialized in line 750 instead of here given it is only used when the version is higher than O.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
```diff
@@ -657,53 +685,123 @@ public int getSampleRate() {
...
+                        try {
+                            startAudioLocked();
+                        } catch (IOException e) {
+                            showErrorToast("startAudio() caught " + e.getMessage());
+                        }
+                    }
+                    break;
+                case AudioManager.AUDIOFOCUS_LOSS:
+                    if (currentState == AUDIO_STATE_STARTED) {
+                        stopAudioLocked();
```
flamme, line 844: Please explain why stop here and pause below.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
```diff
@@ -791,16 +890,39 @@ private void openStreamContext(StreamContext streamContext) throws IOException {
...
+        int result;
+
+        if (mAudioState != AUDIO_STATE_OPEN
+                && mAudioState != AUDIO_STATE_STOPPED
+                && mAudioState != AUDIO_STATE_PAUSED
+                && mAudioState != AUDIO_STATE_FLUSHED) {
+            Log.w(TAG, "startAudio() called in invalid state " + mAudioState + ", ignoring.");
+            return;
+        }
+        Log.i(TAG, "startAudio() transitioning to STARTED");
```
flamme: This log looks incorrect, it is not always transited to STARTED.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
```diff
@@ -791,16 +890,39 @@ private void openStreamContext(StreamContext streamContext) throws IOException {
...
+        result = startNative();
+        if (result == 0) {
+            mAudioState = AUDIO_STATE_STARTED;
+        }
+
         if (result != 0) {
```
flamme, line 973: if...else. Save one comparison

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
```diff
@@ -820,43 +941,93 @@ protected void toastPauseError(int result) {
...
+
+    @GuardedBy("mAudioStateLock")
+    private void pauseAudioLocked() {
+        int result;
+        if (mAudioState != AUDIO_STATE_STARTED) {
+            Log.w(TAG, "pauseAudio() called in invalid state " + mAudioState + ", ignoring.");
+            return;
+        }
+        result = pauseNative();
+        if (result == 0) {
```
flamme: this is redundant as the following has `if(result != 0)... else`

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
```diff
@@ -820,43 +941,93 @@ protected void toastPauseError(int result) {
...
     public void flushAudio() {
-        int result = flushNative();
+        int result;
+        synchronized (mAudioStateLock) {
+            if (mAudioState != AUDIO_STATE_PAUSED) {
+                Log.w(TAG, "flushAudio() called in invalid state " + mAudioState + ", ignoring.");
+                return;
+            }
+            result = flushNative();
+            if (result == 0) {
```
flamme: ditto

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
```diff
@@ -820,43 +941,93 @@ protected void toastPauseError(int result) {
...
-        int result = stopNative();
+        synchronized (mAudioStateLock) {
+            stopAudioLocked();
+        }
+    }
+
+    @GuardedBy("mAudioStateLock")
+    private void stopAudioLocked() {
+        int result;
+        if (mAudioState != AUDIO_STATE_STARTED && mAudioState != AUDIO_STATE_PAUSED) {
```
flamme: Why stop cannot be called when paused?

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
```diff
@@ -820,43 +941,93 @@ protected void toastPauseError(int result) {
...
             updateEnabledWidgets();
             onStopAllContexts();
         }
     }
 
     public void flushAudio() {
-        int result = flushNative();
+        int result;
+        synchronized (mAudioStateLock) {
+            if (mAudioState != AUDIO_STATE_PAUSED) {
```
flamme: I don't this we should check the state here and this is wrong. See https://cs.android.com/android/platform/superproject/main/+/main:frameworks/av/media/libaaudio/include/aaudio/AAudio.h;l=1874?q=aaudio.h

OboeTester should be able to call flush and let the lower layer return failure.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
```diff
@@ -820,43 +941,93 @@ protected void toastPauseError(int result) {
...
-        int result = pauseNative();
+        synchronized (mAudioStateLock) {
+            pauseAudioLocked();
+        }
+    }
+
+    @GuardedBy("mAudioStateLock")
+    private void pauseAudioLocked() {
+        int result;
+        if (mAudioState != AUDIO_STATE_STARTED) {
```
flamme: Please do not silent the pause call. OboeTester should be able to call pause and let the lower layer decide if it should success or not.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
```diff
@@ -820,43 +941,93 @@ protected void toastPauseError(int result) {
...
             updateEnabledWidgets();
             onStopAllContexts();
         }
     }
 
     public void releaseAudio() {
-        int result = releaseNative();
+        int result;
+        synchronized (mAudioStateLock) {
+            if (mAudioState < AUDIO_STATE_OPEN || mAudioState >= AUDIO_STATE_RELEASED) {
```
flamme: Don't check the state and silence the failure here. Reason as above

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
```diff
@@ -906,8 +1092,12 @@ public synchronized void closeAudio() {
             }
         }
 
+        if (isAudioFocusEnabled()) {
+            AudioManager am = (AudioManager) getSystemService(Context.AUDIO_SERVICE);
```
flamme: A more practical way is to cache the audio manager instead of getting it every time when closing.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/StreamConfiguration.java
flamme: This is not updated.

PR #2293, opened 2025-08-19
Toggle That Disables MMap Mode

madebymozart:
A toggle that will enable or disable the MMap Mode


samples/powerplay/src/main/cpp/PowerPlayMultiPlayer.cpp
```diff
@@ -112,4 +112,22 @@ void PowerPlayMultiPlayer::triggerDown(int32_t index, oboe::PerformanceMode perf
     if (mAudioStream) {
         startStream();
     }
-}
\ No newline at end of file
+}
+
+bool PowerPlayMultiPlayer::setMMapEnabled(bool enabled) {
+    auto result = static_cast<MMapPolicy>(oboe::OboeExtensions::setMMapEnabled(enabled));
+    return result == MMapPolicy::Auto;
```
flamme: I thought the API returned 0 or negative error. It doesn't return the mmap policy. https://github.com/google/oboe/blob/6b03a2c07c090b132febc83fec99aa4912b33c0a/include/oboe/OboeExtensions.h#L53

samples/powerplay/src/main/kotlin/com/google/oboe/samples/powerplay/engine/PowerPlayAudioPlayer.kt
```diff
@@ -80,6 +105,10 @@ class PowerPlayAudioPlayer() : DefaultLifecycleObserver {
     private external fun setLoopingNative(index: Int, looping: Boolean)
     private external fun startPlayingNative(index: Int, mode: OboePerformanceMode)
     private external fun stopPlayingNative(index: Int)
+    private external fun setMMapEnabledNative(enabled: Boolean): Boolean
+    private external fun isMMapEnabledNative(): Boolean
+    private external fun isMMapUsedNative(): Boolean
```
flamme: What is the usage of `isMMapUsed` and `isMMapSupported`? When I saw these methods, I supposed to only see mmap checkbox when `isMMapSupported` is true and another hint/text about if mmap is used. But I seems not see  any usage of these two methods.

samples/powerplay/src/main/kotlin/com/google/oboe/samples/powerplay/MainActivity.kt
```diff
@@ -285,6 +293,47 @@ class MainActivity : ComponentActivity() {
...
+                            .fillMaxWidth()
+                            .padding(horizontal = 32.dp)
+                            .selectable(
+                                selected = !isMMapEnabled.value,
+                                onClick = {
+                                    if (!isPlaying.value) {
+                                        isMMapEnabled.value = !isMMapEnabled.value
+                                        player.setMMapEnabled(isMMapEnabled.value)
+                                        player.teardownAudioStream()
+                                        player.setupAudioStream()
```
robertwu1: Why does the MMAP toggle teardown and setup the audio stream when the performance mode toggle only tears down the audio stream?

samples/powerplay/src/main/kotlin/com/google/oboe/samples/powerplay/MainActivity.kt
```diff
@@ -256,6 +257,13 @@ class MainActivity : ComponentActivity() {
                                             onOptionSelected(text)
                                             offload.intValue = index
                                             player.teardownAudioStream()
+
+                                            // Reset MMAP checkbox if not PCM Offload
```
robertwu1: Can we use the MMAP checkbox for all the power modes? This way, there's less random maintenance work and it's easier for testing

samples/powerplay/src/main/kotlin/com/google/oboe/samples/powerplay/MainActivity.kt
```diff
@@ -297,10 +346,21 @@ class MainActivity : ComponentActivity() {
...
                             when (isPlaying.value) {
                                 true -> player.stopPlaying(playingSongIndex.intValue)
-                                false -> player.startPlaying(
-                                    playingSongIndex.intValue,
-                                    OboePerformanceMode.fromInt(offload.intValue)
-                                )
+                                false -> {
+                                    // Ensure the audio stream is set up before playing,
+                                    // especially if it was torn down due to mode changes.
+                                    // You might need a more sophisticated way to check if setup is needed.
```
robertwu1: AI generated comment. Why do we need this new code? Can we use the same code as before if we don't setup the audio stream when it's toggled? Can we make the MMAP toggle like the performance mode buttons where's it's disabled once a stream is started?

samples/powerplay/src/main/cpp/PowerPlayMultiPlayer.cpp
madebymozart: My thoughts we're to have the API's there just when they will be used. I think this can inform other UI elements

flamme: I agreed that it was needed to notify UI elements. My comment is that `OboeExtensions::setMMapEnabled` returns 0 or negative error. In that case, to check if the call to `OboeExtensions::setMMapEnabled` succeed or not, it needs to be compared with 0 instead of `MMapPolicy::Auto`. `MMapPolicy::Auto` is the enum defined for mmap policy, its under layer value is 2.

madebymozart: Ah got it and updated!

samples/powerplay/src/main/cpp/PowerPlayMultiPlayer.cpp
```diff
@@ -112,4 +112,17 @@ void PowerPlayMultiPlayer::triggerDown(int32_t index, oboe::PerformanceMode perf
...
+    return result == 0;
+}
+
+bool PowerPlayMultiPlayer::isMMapEnabled() {
+    return oboe::OboeExtensions::isMMapEnabled();
+}
+
+bool PowerPlayMultiPlayer::isMMapSupported() {
+    return oboe::OboeExtensions::isMMapSupported();
 }
```
robertwu1: nit: newline at the end of file

samples/powerplay/src/main/kotlin/com/google/oboe/samples/powerplay/MainActivity.kt
```diff
@@ -312,6 +352,7 @@ class MainActivity : ComponentActivity() {
         }
     }
 
+
```
robertwu1: nit: remove extra line

PR #2298, opened 2025-08-20
OboeTester: Don't crash cold start latency for invalid paths

robertwu1:
Fixes b/440110746

For example, when you start cold start latency with input, no MMAP but exclusive set, the stream crashes. We should send a toast instead


apps/OboeTester/app/src/main/cpp/TestColdStartLatency.cpp
```diff
@@ -52,7 +52,9 @@ int32_t TestColdStartLatency::open(bool useInput, bool useLowLatency, bool useMm
     // Revert MMAP back to its previous state
     AAudioExtensions::getInstance().setMMapEnabled(wasMMapEnabled);
 
-    mDeviceId = mStream->getDeviceId();
+    if (mDeviceId) {
```
flamme: Why check mdeviceid instead of opening result?

robertwu1: Done. Thanks!

PR #2302, opened 2025-08-29
Add compile flag option for not including flowgraph

robertwu1:
This PR introduces a new compile flag to not use the flowgraph. Since AAudio has included sample rate conversion in Android V, there is not strong reason we need to use the flowgraph anymore. Adding a compiler flag to turn this off. This saves half the binary size for Oboe

With -DOBOE_DISABLE_CONVERSION=OFF
~~~
robertwu-mac:arm64-v8a robertwu$ ls -l
total 7304
-rw-r--r--   1 robertwu  primarygroup     3522 Aug 29 11:40 cmake_install.cmake
-rw-r--r--   1 robertwu  primarygroup    17017 Aug 29 11:40 CMakeCache.txt
drwxr-xr-x@ 16 robertwu  primarygroup      512 Aug 29 11:40 CMakeFiles
-rwxr-xr-x@  1 robertwu  primarygroup  3630584 Aug 29 11:40 liboboe.so
-rw-r--r--   1 robertwu  primarygroup    78452 Aug 29 11:40 Makefile
drwxr-xr-x   2 robertwu  primarygroup       64 Dec 16  2024 staging
~~~

With -DOBOE_DISABLE_CONVERSION=ON
~~~
robertwu-mac:arm64-v8a robertwu$ ls -l
total 3672
-rw-r--r--   1 robertwu  primarygroup     3522 Aug 29 11:40 cmake_install.cmake
-rw-r--r--   1 robertwu  primarygroup    17016 Aug 29 11:38 CMakeCache.txt
drwxr-xr-x@ 16 robertwu  primarygroup      512 Aug 29 11:40 CMakeFiles
-rwxr-xr-x@  1 robertwu  primarygroup  1815848 Aug 29 11:40 liboboe.so
-rw-r--r--   1 robertwu  primarygroup    35600 Aug 29 11:40 Makefile
drwxr-xr-x   2 robertwu  primarygroup       64 Dec 16  2024 staging
~~~

Note that cleaning the project doesn't remove the flag if you set it.
For me, if I enable it once in OboeTester, I need to manually remove it in /Users/robertwu/Documents/oboe/apps/OboeTester/app/.cxx/Debug/41t402x1/arm64-v8a/CMakeCache.txt or remove the .cxx subdirectory to delete it

### DIFF
--- a/build_all_android.sh
+++ b/build_all_android.sh
@@ -34,7 +34,6 @@ CMAKE_ARGS="-S. \
   -DANDROID_TOOLCHAIN=clang \
   -DANDROID_STL=c++_shared \
   -DCMAKE_TOOLCHAIN_FILE=${ANDROID_NDK}/build/cmake/android.toolchain.cmake \
-  -DOBOE_DISABLE_CONVERSION=OFF \
   -DCMAKE_INSTALL_PREFIX=."
 
 function build_oboe {
--- a/build_all_android.sh
+++ b/build_all_android.sh
@@ -34,6 +34,7 @@ CMAKE_ARGS="-S. \
   -DANDROID_TOOLCHAIN=clang \
   -DANDROID_STL=c++_shared \
   -DCMAKE_TOOLCHAIN_FILE=${ANDROID_NDK}/build/cmake/android.toolchain.cmake \
+  -DOBOE_DISABLE_CONVERSION=OFF \
   -DCMAKE_INSTALL_PREFIX=."
 
 function build_oboe {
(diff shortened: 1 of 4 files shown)


build_all_android.sh
```diff
@@ -34,6 +34,7 @@ CMAKE_ARGS="-S. \
   -DANDROID_TOOLCHAIN=clang \
   -DANDROID_STL=c++_shared \
   -DCMAKE_TOOLCHAIN_FILE=${ANDROID_NDK}/build/cmake/android.toolchain.cmake \
+  -DOBOE_DISABLE_CONVERSION=OFF \
```
flamme: If it is by default off, is it still needed here?

robertwu1: Removed. Thanks for all your help!

PR #2309, opened 2025-09-16
OboeTester: Stop stream sniffer when screen is off

robertwu1:
Fixes b/445260454

### DIFF
--- a/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
@@ -133,6 +133,21 @@ abstract class TestAudioActivity extends AppCompatActivity {
         }
     };
 
+    private BroadcastReceiver mScreenStateReceiver = new BroadcastReceiver() {
+        @Override
+        public void onReceive(Context context, Intent intent) {
+            if (intent.getAction().equals(Intent.ACTION_SCREEN_OFF)) {
+                if (mStreamSniffer != null) {
+                    mStreamSniffer.stopStreamSniffer();
+                }
+            } else if (intent.getAction().equals(Intent.ACTION_SCREEN_ON)) {
+                if (mStreamSniffer != null && mAudioState < AUDIO_STATE_CLOSING) {
+                    mStreamSniffer.startStreamSniffer();
+                }
+            }
+        }
+    };
+
     private final long mActivityId = generateActivityId();
     private static long CURRENT_ACTIVITY_ID = 0;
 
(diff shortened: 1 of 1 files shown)


apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
```diff
@@ -133,6 +133,21 @@ public void onReceive(Context context, Intent intent) {
...
 
+    private BroadcastReceiver mScreenStateReceiver = new BroadcastReceiver() {
+        @Override
+        public void onReceive(Context context, Intent intent) {
+            if (intent.getAction().equals(Intent.ACTION_SCREEN_OFF)) {
+                if (mStreamSniffer != null) {
+                    mStreamSniffer.stopStreamSniffer();
+                }
+            } else if (intent.getAction().equals(Intent.ACTION_SCREEN_ON)) {
+                if (mStreamSniffer != null && mAudioState < AUDIO_STATE_CLOSING) {
```
flamme: Should the audio state compare in this way?

robertwu1: I think it's fine because closeAudio looks as follows:

~~~
    // Make synchronized so we don't close from two streams at the same time.
    public synchronized void closeAudio() {
        if (mAudioState >= AUDIO_STATE_CLOSING) {
            Log.d(TAG, "closeAudio() already closing");
            return;
        }
~~~

flamme: That is also something I want us to fix. There is not any guarantee this will always work.

robertwu1: Thanks! Done!

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestAudioActivity.java
```diff
@@ -133,6 +133,21 @@ public void onReceive(Context context, Intent intent) {
         }
     };
 
+    private BroadcastReceiver mScreenStateReceiver = new BroadcastReceiver() {
+        @Override
+        public void onReceive(Context context, Intent intent) {
+            if (intent.getAction().equals(Intent.ACTION_SCREEN_OFF)) {
```
flamme: I think it is more encouraged to use `Intent.ACTION_SCREEN_OFF.equals(intent.getAction())` to avoid action is null.

robertwu1: Done! Thanks!

flamme (approved):

PR #2310, opened 2025-09-16
fxlab: Add default effects

robertwu1:
Add a set of reasonable default effects:
1. Gain: This lets you increase the volume of the output.
2. Echo: This will create a sense of space and ambience with decaying repeats of the sound.
3. Tremolo: This produced a rapid wavering sound.

### DIFF
--- a/apps/fxlab/app/src/main/kotlin/com/mobileer/androidfxlab/MainActivity.kt
+++ b/apps/fxlab/app/src/main/kotlin/com/mobileer/androidfxlab/MainActivity.kt
@@ -154,7 +154,7 @@ class MainActivity : AppCompatActivity() {
     }
 
     private fun addDefaultEffects() {
-        val defaultEffects = listOf("Overdrive", "White Chorus", "Echo")
+        val defaultEffects = listOf("Gain", "Echo", "Tremolo")
         for (effectName in defaultEffects) {
             NativeInterface.effectDescriptionMap[effectName]?.let {
                 val toAdd = Effect(it)


flamme:
As talked offline, please add Atneya for review as he is more familiar with the code here.

apps/fxlab/app/src/main/kotlin/com/mobileer/androidfxlab/MainActivity.kt
```diff
@@ -154,7 +154,7 @@ class MainActivity : AppCompatActivity() {
     }
 
     private fun addDefaultEffects() {
-        val defaultEffects = listOf("Overdrive", "White Chorus", "Echo")
```
atneya: Sounds more interesting? Curious about the inclusion of gain as a default since it is quite "boring"

robertwu1: I added gain just so users can easily adjust volume as needed. Let me know if you like these better

atneya:
The UI logic could probably be easily replaced with some more modern alternatives (i.e. compose). I think right now, we aren't persisting the effectlist in a viewmodel so I don't think it handles rotation correctly.

robertwu1:
> The UI logic could probably be easily replaced with some more modern alternatives (i.e. compose). I think right now, we aren't persisting the effectlist in a viewmodel so I don't think it handles rotation correctly.

That's a bigger change for the future. I think the rotation isn't persisted correctly but at least it can rotate with a little icon at the bottom

PR #2311, opened 2025-09-17
fxlab: add new icon

robertwu1:
This PR adds a new icon for fxlab. It also changes the name of the app to fxlab from "AndroidFx Lab"

### DIFF
--- a/apps/fxlab/app/src/main/res/values/strings.xml
+++ b/apps/fxlab/app/src/main/res/values/strings.xml
@@ -15,6 +15,6 @@
   -->
 
 <resources>
-    <string name="app_name">fxlab</string>
+    <string name="app_name">Fxlab</string>
     <string name="action_toggle_mute">Toggle volume</string>
 </resources>


apps/fxlab/app/src/main/res/values/strings.xml
```diff
@@ -15,6 +15,6 @@
   -->
 
 <resources>
-    <string name="app_name">AndroidFX Lab</string>
+    <string name="app_name">fxlab</string>
```
flamme: nit, do we prefer to have first letter upper case for app name?

apps/fxlab/app/src/main/res/values/ic_launcher_background.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <color name="ic_launcher_background">#111111</color>
+</resources>
```
flamme: missing new line

apps/fxlab/app/src/main/res/mipmap-anydpi-v26/ic_launcher.xml
```diff
@@ -1,21 +1,5 @@
...
-  ~ See the License for the specific language governing permissions and
-  ~ limitations under the License.
-  -->
-
 <adaptive-icon xmlns:android="http://schemas.android.com/apk/res/android">
-    <background android:drawable="@drawable/ic_launcher_background"/>
-    <foreground android:drawable="@drawable/ic_launcher_foreground"/>
+    <background android:drawable="@color/ic_launcher_background"/>
+    <foreground android:drawable="@mipmap/ic_launcher_foreground"/>
 </adaptive-icon>
```
flamme: missing a new line at the end of file

apps/fxlab/app/src/main/res/mipmap-anydpi-v26/ic_launcher_round.xml
```diff
@@ -1,21 +1,5 @@
...
-  ~ See the License for the specific language governing permissions and
-  ~ limitations under the License.
-  -->
-
 <adaptive-icon xmlns:android="http://schemas.android.com/apk/res/android">
-    <background android:drawable="@drawable/ic_launcher_background"/>
-    <foreground android:drawable="@drawable/ic_launcher_foreground"/>
+    <background android:drawable="@color/ic_launcher_background"/>
+    <foreground android:drawable="@mipmap/ic_launcher_foreground"/>
 </adaptive-icon>
```
flamme: missing a new line at the end of file

apps/fxlab/app/src/main/res/values/strings.xml
robertwu1: not necessarily. There's other apps like soundboard that are lowercased. I prefer this to be lowercase  as it's a for fun app

flamme: There is not requirement from the development point of view as a string, it accepts both lower case and upper case letters. But from more common practice point of view, starting with upper case is more flavored. I'm also not seeing the reason of renaming here.

robertwu1: The current name is so long that it goes beyond the normal viewing area of the app. This is why we should rename this.

When you open the app, the top of the app is already "fxlab". I just added a commit where we changed it to Fxlab

PR #2315, opened 2025-09-30
[PCM Offload] • Implement FlushFromFrame for PCM Offload

madebymozart:
This PR implements `flushFromFrame` when PCM offload is enabled. This allows the switching of songs as well by swiping the records at the top.

https://github.com/user-attachments/assets/b2e50c2c-3b96-4ee7-af8c-ecf93d387e45


samples/iolib/src/main/cpp/player/SampleSource.h
```diff
@@ -46,13 +46,13 @@ class SampleSource: public DataSource {
 
     void setPlayMode(bool fromPause = true) {
         mIsPlaying = true;
-        if (fromPause) {
+        if (fromPause == true) {
```
flamme: Q: Why original `if(fromPause)` doesn't work?

samples/iolib/src/main/cpp/player/SampleSource.h
```diff
@@ -46,13 +46,13 @@ class SampleSource: public DataSource {
...
             mCurSampleIndex = 0;
         }
     }
     void setStopMode(bool isPause = false) {
         mIsPlaying = false;
-        if (!isPause) {
+        if (isPause == false) {
```
flamme: Q: why original `if(!isPause)` doesn't work?

samples/iolib/src/main/cpp/player/SampleSource.h
```diff
@@ -46,13 +46,13 @@ class SampleSource: public DataSource {
 
     void setPlayMode(bool fromPause = true) {
         mIsPlaying = true;
-        if (fromPause) {
```
robertwu1, line 49: Why change this? Was this AI generated?

madebymozart: Nope, might have been my testing as I was trying to figure out the functionality as it didn't make sense initially. will revert back

samples/powerplay/src/main/cpp/PowerPlayMultiPlayer.cpp
```diff
@@ -86,18 +86,18 @@ bool PowerPlayMultiPlayer::openStream(oboe::PerformanceMode performanceMode) {
 
 
 void PowerPlayMultiPlayer::triggerUp(int32_t index) {
-    if (index >= 0 && index < mNumSampleBuffers) {
-        mSampleSources[index]->setStopMode(true);
-    }
     if (mAudioStream) {
         mAudioStream->pause();
     }
+    mSampleSources[index]->setStopMode(true);
```
robertwu1: I'm a bit confused about the setStopMode/setPlayMode changes. Can you explain why each of these changes are made?

samples/powerplay/src/main/cpp/PowerPlayMultiPlayer.cpp
```diff
@@ -109,8 +109,17 @@ void PowerPlayMultiPlayer::triggerDown(int32_t index, oboe::PerformanceMode perf
...
+    mSampleSources[index]->setPlayMode(performanceModeChanged);
+
     if (mAudioStream) {
-        startStream();
+        if (performanceModeChanged &&
+            mAudioStream->getPerformanceMode() == PerformanceMode::PowerSavingOffloaded &&
+            mSampleSources[index]->getCurserIndex() != 0) {
+            mAudioStream->flushFromFrame(FlushFromAccuracy::Accurate, 0);
+        } else {
+            startStream(mLastPerformanceMode);
```
robertwu1: Why is mLastPerformanceMode used now? Can you make the minimum set of changes so this is easier to review?

samples/iolib/src/main/cpp/player/SampleSource.h
madebymozart: Addressed in the latest commit

samples/iolib/src/main/cpp/player/SampleSource.h
madebymozart: addressed in the latest commit

samples/powerplay/src/main/cpp/PowerPlayMultiPlayer.cpp
```diff
@@ -109,8 +109,17 @@ void PowerPlayMultiPlayer::triggerDown(int32_t index, oboe::PerformanceMode perf
...
     }
+
+    mSampleSources[index]->setPlayMode(performanceModeChanged);
+
     if (mAudioStream) {
-        startStream();
+        if (performanceModeChanged &&
+            mAudioStream->getPerformanceMode() == PerformanceMode::PowerSavingOffloaded &&
+            mSampleSources[index]->getCurserIndex() != 0) {
+            mAudioStream->flushFromFrame(FlushFromAccuracy::Accurate, 0);
```
flamme: This will always fail as you cannot accurately flush from the 0 frame as it is already played.

samples/powerplay/src/main/cpp/PowerPlayMultiPlayer.cpp
madebymozart: There was a bug where when you started a stream, it would under the hood change the performance mode back to LowLatency, this causing the Power Savings mode to be switched off when changing streams. 

This Pr wouldn't work without this change

samples/powerplay/src/main/cpp/PowerPlayMultiPlayer.cpp
madebymozart: This is actually a slight miss overall here. Ther code here is from drumthumper where multiple samples are played at once. This doesn't happen in PowerPlay so there is less need for us to go through each sample when only one is being played at a time.

samples/iolib/src/main/cpp/player/SampleSource.h
```diff
@@ -61,6 +61,8 @@ class SampleSource: public DataSource {
 
     bool isPlaying() { return mIsPlaying; }
 
+    int32_t getCurserIndex() { return mCurSampleIndex; }
```
robertwu1: typo

samples/powerplay/src/main/cpp/PowerPlayMultiPlayer.cpp
```diff
@@ -109,8 +109,17 @@ void PowerPlayMultiPlayer::triggerDown(int32_t index, oboe::PerformanceMode perf
             return;
         }
     }
+
+    mSampleSources[index]->setPlayMode(performanceModeChanged);
```
robertwu1: Shouldn't this be reversed? We want to start from the beginning of the stream if the performance mode has changed and continue from where we originally were if we just press the stop button and the start button?

samples/powerplay/src/main/cpp/PowerPlayMultiPlayer.cpp
robertwu1: Can you test that DrumThumper still works after your change?

samples/powerplay/src/main/cpp/PowerPlayMultiPlayer.cpp
```diff
@@ -109,8 +111,17 @@ void PowerPlayMultiPlayer::triggerDown(int32_t index, oboe::PerformanceMode perf
...
+
     if (mAudioStream) {
-        startStream();
+        if (performanceModeChanged &&
```
flamme: Sorry but I don't get the code here. If the performance mode changed, shouldn't the stream be closed and reopened with the new performance mode? I don't think aaudio support dynamically changing the performance mode.

samples/powerplay/src/main/cpp/PowerPlayMultiPlayer.cpp
madebymozart: This shouldn't affect drumthumper at all because I've subclassed the `iolib::SimpleMultiPlayer` which is what drumthumper uses. This class is the subclass [PowerPlayMultiPlayer.cpp](https://github.com/google/oboe/pull/2315/files/f6fc201e7d899c82454004f8fb573cf6eb9bd88b#diff-2e5d5b4fe76b7401609b97c4577e4f1dc9cc62e4907cb5113a234a357d2eea4b)

samples/powerplay/src/main/cpp/PowerPlayMultiPlayer.cpp
madebymozart: Yeah I realized this afterwards: I've fixed this code and the teardown of the stream already happens earlier @ line 105

```
    if (performanceModeChanged) {
        teardownAudioStream();
        if (!openStream(performanceMode)) {
            __android_log_print(ANDROID_LOG_ERROR,
                                TAG,
                                "Failed to reopen stream with new performance mode");
            return;
        }
    }
```

samples/powerplay/src/main/cpp/PowerPlayMultiPlayer.cpp
```diff
@@ -109,8 +111,24 @@ void PowerPlayMultiPlayer::triggerDown(int32_t index, oboe::PerformanceMode perf
...
+            if (result != Result::OK) {
+
+                __android_log_print(ANDROID_LOG_ERROR,
+                                    TAG,
+                                    "Failed to flush from frame. Error: %s",
+                                    convertToText(result.error()));
+            }
+            mAudioStream->start();
+        } else {
+            startStream(mLastPerformanceMode);
```
robertwu1: Why startStream here and mAudioStream->start above?

samples/powerplay/src/main/cpp/PowerPlayMultiPlayer.cpp
madebymozart: Hmmmm I guess this is more of a preference thing for this one. I like the fact the when the performance mode changes, we start playback from the beginning. But that doesn't need to be the case if we don't want it too.

I doing a more thorough code rewrite this this method.

samples/powerplay/src/main/cpp/PowerPlayMultiPlayer.cpp
```diff
@@ -95,22 +94,70 @@ void PowerPlayMultiPlayer::triggerUp(int32_t index) {
...
+
+
+    const auto currentPerformanceMode = mAudioStream->getPerformanceMode();
+    const auto currentlyPlayingIndex = getCurrentlyPlayingIndex();
+
+    // Assure all other loaded samples are stopped and the play head is reset to zero, avoiding the
+    // currently playing index. Only allow the playback head to reset when the song has changed.
+    for (size_t i = 0; i < mSampleSources.size(); ++i) {
+        if (i != index) mSampleSources[i]->setStopMode(false);
+        else mSampleSources[i]->setPlayMode(currentlyPlayingIndex == i);
```
flamme: Shouldn't this always play from the first sample when switching song?

samples/powerplay/src/main/cpp/PowerPlayMultiPlayer.cpp
```diff
@@ -126,3 +173,12 @@ bool PowerPlayMultiPlayer::isMMapEnabled() {
...
 }
+
+int32_t PowerPlayMultiPlayer::getCurrentlyPlayingIndex() {
+    for (auto i = 0; i < mSampleSources.size(); ++i) {
+        if (mSampleSources[i]->isPlaying()) return i;
+    }
+
+    // No source is currently playing.
+    return -1;
+}
```
flamme: Need an empty line at the end of file.

samples/powerplay/src/main/cpp/PowerPlayMultiPlayer.cpp
madebymozart: It does. This is part of the confusing part and using DrumThumper as the base for this, it's slightly less clear whats happening.

the `setPlayMode` method looks like this: 

```
    void setPlayMode(bool fromPause = true) {
        mIsPlaying = true;
        if (fromPause) {
            mCurSampleIndex = 0;
        }
    }
```

This means in order for the playhead to be reset to zero, from pause needs to be true, but that doesn't make much sense, which also lead me to confusion. Now I'd change it to make more sense, but it'll affect DrumThumper as well and I'd like to avoid that.

madebymozart: What we do what though is if the user just changes the performance mode, we don't want them to lost the playhead position, thus why the `currentlyPlayingIndex == i` is there.

madebymozart: Ah nevermind, I see what's happened. This is another symptom of DrumThumper being the base. When the user pauses, all of the samples are put in stop mode and reset

samples/iolib/src/main/cpp/player/SampleSource.h
```diff
@@ -57,10 +57,20 @@ class SampleSource: public DataSource {
         }
     }
 
+    void play() {
+        mIsPlaying = true;
```
flamme: I'd suggest to call setPlayMode(false) here instead of modifying the `mIsPlaying`. Same for below.

I'm okay with the change. But would like to have Robert to comment here as he is the owner for code here.

robertwu1 (approved):

samples/iolib/src/main/cpp/player/SampleSource.h
robertwu1: I agree that we should call setPlayMode. Otherwise, we lose the fromPause parameter

madebymozart: Done

samples/powerplay/src/main/cpp/PowerPlayMultiPlayer.cpp
madebymozart: This looks like a old commit - This has been updated a while ago

samples/powerplay/src/main/cpp/PowerPlayMultiPlayer.cpp
```diff
@@ -84,33 +84,101 @@ bool PowerPlayMultiPlayer::openStream(oboe::PerformanceMode performanceMode) {
...
-    if (mAudioStream) {
-        startStream();
+
+    // Assure previous sample is stopped and the play head is reset to zero, avoiding the
+    // currently playing index. Only allow the playback head to reset when the song has changed.
+    const auto currentlyPlayingIndex = getCurrentlyPlayingIndex();
+    if (currentlyPlayingIndex != -1 && currentlyPlayingIndex != index) {
+        mSampleSources[currentlyPlayingIndex]->setStopMode(false);
+    }
+    mSampleSources[index]->setPlayMode(false);
```
robertwu1, line 155: Just to confirm, what happens when you press pause and play and pause and play again? Does the song start from the beginning or from the position that we just paused at?

madebymozart: It plays from the position that we just paused at

PR #2318, opened 2025-10-03
OboeTester: Use callbacks for showCommDeviceStatus

### LINKED ISSUE #2316: oboetester: AudioManager isSpeakerPhoneOn is flooding logcat
steps : 
Open Oboe 
Test Output
Check logs.
AudioManager's isSpeakerPhoneOn is flooding the logs.

It is coming via 

    protected class CommDeviceSniffer extends NativeSniffer {
        @Override
        public void updateStatusText() {
           showCommDeviceStatus();
        }
    }

triggered from here
    public void run() {
        if (!isComplete()) {
            updateStatusText();
        }

        // When this is no longer enabled, stop calling run.
        if (mEnabled) {
            mHandler.postDelayed(this, SNIFFER_UPDATE_PERIOD_MSEC);
        }
    }

->probably isComplete is not overriden in CommunicationDeviceView to mark sniffing complete.

Can you please check ?

robertwu1:
Fixes #2316

### DIFF
--- a/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/CommunicationDeviceView.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/CommunicationDeviceView.java
@@ -45,14 +45,10 @@ public class CommunicationDeviceView extends LinearLayout {
     private boolean mScoStateReceiverRegistered = false;
     private CommunicationDeviceSpinner mDeviceSpinner;
     private int mScoState;
-    private CommDeviceSniffer mCommDeviceSniffer = new CommDeviceSniffer();;
 
-    protected class CommDeviceSniffer extends NativeSniffer {
-        @Override
-        public void updateStatusText() {
-            showCommDeviceStatus();
-        }
-    }
+    private AudioManager.OnCommunicationDeviceChangedListener mCommDeviceListener;
+    private BroadcastReceiver mSpeakerphoneStateReceiver;
+    private boolean mSpeakerphoneStateReceiverRegistered = false;
 
     public CommunicationDeviceView(Context context) {
         super(context);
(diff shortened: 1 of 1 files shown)


apps/OboeTester/app/src/main/java/com/mobileer/oboetester/CommunicationDeviceView.java
```diff
@@ -45,14 +45,10 @@ public class CommunicationDeviceView extends LinearLayout {
...
     private int mScoState;
-    private CommDeviceSniffer mCommDeviceSniffer = new CommDeviceSniffer();;
 
-    protected class CommDeviceSniffer extends NativeSniffer {
-        @Override
-        public void updateStatusText() {
-            showCommDeviceStatus();
-        }
-    }
+    private AudioManager.OnCommunicationDeviceChangedListener mCommDeviceListener;
```
quicjaidshar, line 49: Isn't mCommDeviceListener sufficient to get device callbacks ? Why do we explicitly need speakerphone? Is there any scenario where we need speaker change callbacks which are not sent as part of mCommunicationDeviceChangedListener ?

robertwu1: Yes, you're right. Explicitly setting the speakerphone on is strongly discouraged. I can clean this up as a follow-up PR

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/CommunicationDeviceView.java
```diff
@@ -107,6 +103,19 @@ public void onReceive(Context context, Intent intent) {
...
 
+        mSpeakerphoneStateReceiver = new BroadcastReceiver() {
+            @Override
+            public void onReceive(Context context, Intent intent) {
+                showCommDeviceStatus();
+            }
+        };
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
+            mCommDeviceListener = (deviceInfo) -> {
```
flamme, line 114: If communication device listener is used, it looks there is no need to get speaker phone state receiver

PR #2324, opened 2025-10-29
Make sure the stream is alive in data callback.

### LINKED ISSUE #2323: Keep the stream alive in data callback
Theoretically, the app is supposed to free the shared pointer of the stream until the stream is fully closed. In this way, there should not be issue from the data callback as the data callback thread is guaranteed to be joined before returning from close. There are some report that there is data callback from older Android after the stream is closed. There can be cases that apps forget to close the stream but free the shared pointer. In that case, it makes sense that Oboe guarantee the stream alive in data callback.

flamme:
Fixes #2323.

### DIFF
--- a/src/aaudio/AudioStreamAAudio.cpp
+++ b/src/aaudio/AudioStreamAAudio.cpp
@@ -137,7 +137,8 @@ static int32_t oboe_aaudio_partial_data_callback_proc(
     if (!isStreamAlive) {
         // Note that the stream is removed from the collection when close is called. However,
         // there can be callback fired until the framework fully close the stream. In that case,
-        // logging a warning here and quick return to stop the stream.
+        // logging a warning here and return negative number for partial callback to stop the
+        // stream.
         LOGW("%s data callback while stream is not longer alive", __func__);
         return -1;
     }


src/aaudio/AudioStreamAAudio.cpp
```diff
@@ -123,6 +132,15 @@ static int32_t oboe_aaudio_partial_data_callback_proc(
...
         int32_t numFrames) {
     AudioStreamAAudio *oboeStream = reinterpret_cast<AudioStreamAAudio*>(userData);
+    auto [isStreamAlive, sharedStream] =
+            AAudioStreamCollection::getInstance().getStream(oboeStream);
+    if (!isStreamAlive) {
+        // Note that the stream is removed from the collection when close is called. However,
+        // there can be callback fired until the framework fully close the stream. In that case,
+        // logging a warning here and quick return to stop the stream.
+        LOGW("%s data callback while stream is not longer alive", __func__);
+        return -1;
```
robertwu1, line 143: Mention why we use -1

robertwu1:
How often does this happen in practice? If it happens often, should we revisit how we let apps hold callbacks when the stream ends?

flamme:
> How often does this happen in practice? If it happens often, should we revisit how we let apps hold callbacks when the stream ends?

It is always suggested apps should call close and wait for it to return until free the shared pointer. It is not related to stream ends.

robertwu1:
> > How often does this happen in practice? If it happens often, should we revisit how we let apps hold callbacks when the stream ends?
> 
> It is always suggested apps should call close and wait for it to return until free the shared pointer. It is not related to stream ends.

What happens when apps are forcefully closed? How do we handle this?

flamme:
> > > How often does this happen in practice? If it happens often, should we revisit how we let apps hold callbacks when the stream ends?
> > 
> > 
> > It is always suggested apps should call close and wait for it to return until free the shared pointer. It is not related to stream ends.
> 
> What happens when apps are forcefully closed? How do we handle this?

This is a different topic... Apps being forced stopping is not supposed to handle in this PR... If it is apps forcing stop, the application should have the right handle on their destroy process that will call close. If there is other reason that crashing the app and the app is not able to call close, it indicates more serious problem that is out of our scope here.

robertwu1:
> > > > How often does this happen in practice? If it happens often, should we revisit how we let apps hold callbacks when the stream ends?
> > > 
> > > 
> > > It is always suggested apps should call close and wait for it to return until free the shared pointer. It is not related to stream ends.
> > 
> > 
> > What happens when apps are forcefully closed? How do we handle this?
> 
> This is a different topic... Apps being forced stopping is not supposed to handle in this PR... If it is apps forcing stop, the application should have the right handle on their destroy process that will call close. If there is other reason that crashing the app and the app is not able to call close, it indicates more serious problem that is out of our scope here.

Yes, I just wanted to bring this up. The change looks good to me.

robertwu1 (approved):

PR #2327, opened 2025-11-04
Improve TapLatencyAnalyzer to detect TAP and TONE peaks separately

Priyadharshini-commits:
Added Logic to improve the latency caluculation for fatcat device.

### DIFF
--- a/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TapLatencyAnalyser.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TapLatencyAnalyser.java
@@ -1,3 +1,4 @@
+
 /*
  * Copyright 2015 The Android Open Source Project
  *
@@ -17,16 +18,24 @@ package com.mobileer.oboetester;
 
 import java.util.ArrayList;
 
+import android.util.Log;
+
 /**
  * Analyze a recording and extract edges for latency analysis.
  */
 public class TapLatencyAnalyser {
     public static final int TYPE_TAP = 0;
+    public static final int TYPE_TONE = 1;
     float[] mHighPassBuffer;
 
     private float mDroop = 0.995f;
     private static final float EDGE_THRESHOLD = 0.01f;
     private static final float LOW_FRACTION = 0.5f;
+
+    private static final float SETTLE_FRACTION = 0.25f; // 75% decay
+    private static final float REARM_FRACTION  = 0.30f;
+    private static final float SLOPE_EPSILON   = 0.002f;
+    private static final int MAX_SETTLE_SAMPLES = 5000;
 
     public static class TapLatencyEvent {
         public int type;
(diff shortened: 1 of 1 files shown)


robertwu1:
Please fill the cla

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TapLatencyAnalyser.java
```diff
@@ -90,25 +91,43 @@ private TapLatencyEvent[] scanForEdges(float[] peakBuffer, int numSamples) {
...
-                // This allows us to detect a second, smaller peak, but not to trigger
-                // on the smaller variations that occur after the initial peak.
-                lowThreshold = fast * LOW_FRACTION;
-            }
-            // Use hysteresis when rearming.
-            if (fast < lowThreshold) {
-                armed = true;
-            }
-            sampleIndex++;
+		slow = slow + (level - slow) * slowCoefficient; // low pass filter
```
robertwu1: fix indentation

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TapLatencyAnalyser.java
```diff
@@ -90,25 +91,43 @@ private TapLatencyEvent[] scanForEdges(float[] peakBuffer, int numSamples) {
         float fast = 0.0f;
         final float slowCoefficient = 0.01f;
         final float fastCoefficient = 0.10f;
-        float lowThreshold = EDGE_THRESHOLD;
+    	final float EDGE_THRESHOLD = 0.01f;     // trigger threshold for tap
+        final float REARM_FRACTION = 0.3f;      // fraction of last peak to rearm
+        final int MIN_TONE_DELAY = 2000;
```
robertwu1: What units is this in?

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TapLatencyAnalyser.java
```diff
@@ -90,25 +91,43 @@ private TapLatencyEvent[] scanForEdges(float[] peakBuffer, int numSamples) {
...
+			Log.i("TTL", "TTL Event of TYPE_TAP detected at " + sampleIndex + " fast=" + fast);
+			events.add(new TapLatencyEvent(TYPE_TAP, sampleIndex));
+			armed = false;
+			tapDetected = true;
+			lastPeak = fast;
+               	 	lastTapIndex = sampleIndex;
+            	}
+
+	    	if (tapDetected && !toneDetected && !armed && (sampleIndex - lastTapIndex) > MIN_TONE_DELAY) {
+                	if ((fast > EDGE_THRESHOLD) && (fast > 1.5f * slow)) {
```
robertwu1: Where is this 1.5f from?

robertwu1:
Also go to extras -> external tap to try this on another device. The idea is that you can use one device to detect edges on other devices

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TapLatencyAnalyser.java
Priyadharshini-commits: Fixed Indentation

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TapLatencyAnalyser.java
Priyadharshini-commits: samples between TAP and TONE (~45 ms) in milliseconds

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TapLatencyAnalyser.java
Priyadharshini-commits: After detecting a TAP, the code waits a bit (MIN_TONE_DELAY) for the tone generated by the speaker to appear.
But when the tone comes, it’s not as sharp as a finger tap — it’s more continuous or sinusoidal.

fast > 1.5×slow → definite transient change (tone or pulse)
It’s a stability guard — making sure the “tone peak” is real and not part of noise or reverberation.

If we have used fast > 2.0 * slow again, the tone might never be strong enough to pass the condition — you’d miss it entirely (which is what was happening earlier in your logs).

So,
TAP → fast > 2.0 * slow (strong transient)
TONE → fast > 1.5 * slow (milder but distinct increase).

This make sure that TONE appears and latency event been added.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TapLatencyAnalyser.java
robertwu1: Many devices have much lower latency. I really don't like hardcoding something like this 2000 sample check. How does round trip latency look for your device? Maybe if that passes, we can just skip the tap to tone for your form factor

Priyadharshini-commits: The Round Trip Latency for our Device is 88-91ms.
Using the same logic of MIN_TONE_DELAY with a 2000-sample check, we observed an average latency of 116 ms over 10 tests on a Google Pixel device, and an average latency of 106 ms over 10 tests on a standard Android phone. We believe that this logic consistent and does not impact the latency calculations on other devices, ensuring reliable comparisons across platforms.

robertwu1:
Rejecting change because a better solution was found with #2360. Please try it. Thanks!

PR #2328, opened 2025-11-05
Refactor: Use View Binding and remove Data Binding

blundell:
Switches `MainActivity.kt` from using Data Binding to using View Binding. This was possible after removing the root `<layout>` tag from `activity_main.xml`, as the file was only using the layout wrapper for view access.

This change:
* Removes the dependency on Data Binding utilities (`DataBindingUtil`).
* Eliminates the need for the `kotlin-kapt` plugin and the (implicit) `databinding-compiler` dependency, simplifying the build process.
* Modern Android prefers KSP and so removing KAPT makes this module compatible with current apps builds.

### DIFF
--- a/apps/fxlab/app/build.gradle
+++ b/apps/fxlab/app/build.gradle
@@ -14,11 +14,10 @@
  * limitations under the License.
  */
 
-apply plugin: 'com.android.application'
-
-apply plugin: 'kotlin-android'
-
-apply plugin: 'kotlin-kapt'
+plugins {
+    id 'com.android.application'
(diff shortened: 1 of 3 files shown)


robertwu1:
Can you verify that turning the app works as is?

blundell:
> Can you verify that turning the app works as is?


https://github.com/user-attachments/assets/1a32c7be-f206-4c43-8773-4474641dd3ba


Can confirm that the app works the same before and after this change 👍

robertwu1:
Thanks!

Issue #2120, opened 2024-11-24
OboeTester needs volume report in Recorder


philburk:
To help with issues such as #2117, we should add a volume report in the Recorder test.
When the recording is stopped, we could scan the signal and measure peak amplitude and RMS.
These could be displayed above the VU meters.

elivity:
Hello, i created a pull request for this 👋 
https://github.com/google/oboe/pull/2337

PR #2337, opened 2026-01-26
OboeTester needs volume report in Recorder (#2120)

elivity:
Hey, 

This commit adds a report with peak amplitude and RMS of the last recording. It will be output above the meters, as suggested in the issue.

Is this good as is, or can I change something?

Best regards, 
Oskar


robertwu1:
LGTM. Please feel free to submit this

PR #2338, opened 2026-01-26
Re-open stream when MMAP is disabled

madebymozart:
This fixes a bug where the stream would not restart if MMAP setting was changed

### DIFF
--- a/samples/powerplay/src/main/cpp/PowerPlayMultiPlayer.cpp
+++ b/samples/powerplay/src/main/cpp/PowerPlayMultiPlayer.cpp
@@ -38,6 +38,7 @@ void PowerPlayMultiPlayer::setupAudioStream(int32_t channelCount,
 bool PowerPlayMultiPlayer::openStream(oboe::PerformanceMode performanceMode) {
     __android_log_print(ANDROID_LOG_INFO, TAG, "openStream()");
     mLastPerformanceMode = performanceMode;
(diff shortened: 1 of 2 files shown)


samples/powerplay/src/main/cpp/PowerPlayMultiPlayer.cpp
```diff
@@ -131,8 +131,10 @@ void PowerPlayMultiPlayer::triggerDown(int32_t index, oboe::PerformanceMode perf
         return;
     }
 
-    // If the performance mode has changed, we need to reopen the stream.
-    if (performanceMode != mLastPerformanceMode) {
+    // If the performance mode has changed, or if MMAP is disabled but currently used,
+    bool isMMapGlobal = isMMapEnabled();
+    bool isMMapCurrentlyUsed = OboeExtensions::isMMapUsed(mAudioStream.get());
+    if (performanceMode != mLastPerformanceMode || (!isMMapGlobal && isMMapCurrentlyUsed)) {
```
flamme: This may not be correct. If the mAudioStream is originally opened with mmap disabled and then the user reenable mmap, the stream should also be reopened.

madebymozart: Ah yes, just added the fix by making it global

PR #2339, opened 2026-01-26
Fix flushOnFrame producing times where 0 is peered out

madebymozart:
This addresses and issue where 0 frame from the flushonframe were being called, this causing audio to drop (sometimes)

10-14 09:05:04.351 12339 12339 E AudioStreamInternalPlay_Client: flushFromFrame_l() do not have enough data, safePosition=5760, frameWritten=0


madebymozart:
So for some reason, this caused and issue where the audio would just "stop" when switching between songs, but it'd only happen in a couple of occasions. 

Since putting this fix in, I haven't heard it, but that could also be that I just haven't encountered it again.

flamme:
> So for some reason, this caused and issue where the audio would just "stop" when switching between songs, but it'd only happen in a couple of occasions.
> 
> Since putting this fix in, I haven't heard it, but that could also be that I just haven't encountered it again.

This doesn't sound right to me. `getFramesWritten` always returns positive number as long as the stream has ever played. If the stream never starts, the stream is only in open state.

It will be better to capture a bugreport and file a bug to me if calling `flushFromFrame` when active playing cause stream stop.

madebymozart:
Then we ca close this, might have been a old ticket for this one

PR #2340, opened 2026-01-28
Bug/Feature: Fix audio silence when display is turned off

madebymozart:
## Summary
This PR addresses an issue where audio streams could become silent or stall when the device display is turned off. 

Screenshot_20260128_163525 Screenshot_20260128_164517 Screenshot_20260128_164556

## Changes
* **Enhanced State Recovery**: Improved the internal logic to detect when the hardware path has been interrupted by power management.
* **Callback Continuity**: Adjusted the buffer handling to ensure that callbacks continue to be serviced or correctly restarted if the system silences the stream during the display-off transition.
* **Offload Stability**: (If applicable) Fixed a specific stall in the PCM Offload path that occurred when the display-off signal triggered a resource re-evaluation.

## Checklist
- [x] Tested on physical device (Pixel 9 Pro)
- [x] No regressions in Oboe PCM Offload playback
- [x] Code follows Oboe coding style guidelines

### DIFF
--- a/samples/powerplay/src/main/kotlin/com/google/oboe/samples/powerplay/engine/AudioForegroundService.kt
+++ b/samples/powerplay/src/main/kotlin/com/google/oboe/samples/powerplay/engine/AudioForegroundService.kt
@@ -34,7 +34,6 @@ import android.media.session.MediaSession
 import android.media.session.PlaybackState
 import android.os.Binder
 import android.os.IBinder
-import android.os.PowerManager
 import android.util.Log
 import androidx.lifecycle.Observer
 import com.google.oboe.samples.powerplay.MainActivity
@@ -53,7 +52,6 @@ class AudioForegroundService : Service() {
     private lateinit var audioManager: AudioManager
     private lateinit var audioFocusRequest: AudioFocusRequest
     private lateinit var mediaSession: MediaSession
-    private lateinit var wakeLock: PowerManager.WakeLock
     private var currentAlbumArt: Bitmap? = null
 
     lateinit var player: PowerPlayAudioPlayer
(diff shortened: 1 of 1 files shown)


samples/powerplay/src/main/kotlin/com/google/oboe/samples/powerplay/engine/AudioForegroundService.kt
```diff
@@ -23,83 +23,239 @@ import android.app.PendingIntent
...
+
+            val result = audioManager.requestAudioFocus(audioFocusRequest)
+            if (result == AudioManager.AUDIOFOCUS_REQUEST_GRANTED) {
+                // Let UI control playback start
+            } else {
+                Log.e(TAG, "Failed to get audio focus, result: $result")
+            }
+
+            if (!wakeLock.isHeld) {
+                wakeLock.acquire(10*60*1000L /*10 minutes*/)
```
flamme: If wake lock is held when doing audio playback, the device won't be able to go into deep suspend, which indicates the offload work may not really save much power.

samples/powerplay/src/main/kotlin/com/google/oboe/samples/powerplay/engine/AudioForegroundService.kt
```diff
@@ -23,83 +23,239 @@ import android.app.PendingIntent
...
+                })
+                setFlags(MediaSession.FLAG_HANDLES_MEDIA_BUTTONS or MediaSession.FLAG_HANDLES_TRANSPORT_CONTROLS)
+                isActive = true
+            }
+
+            player.getCurrentSongIndexLive().observeForever(songIndexObserver)
+
+            val powerManager = getSystemService(POWER_SERVICE) as PowerManager
+            wakeLock = powerManager.newWakeLock(PowerManager.PARTIAL_WAKE_LOCK, "PowerPlay::AudioWakeLock")
+        } catch (e: Throwable) {
```
flamme: Could it be more specific which exception may be thrown from which function call?

PR #2345, opened 2026-02-24
Add A Progress Slider to PowerPlay

madebymozart:
Added a Progress Slider to PowerPlay
Screenshot_20260224_160652

Note: This currently is disabled in PCM Offload Mode

### DIFF
--- a/samples/powerplay/src/main/cpp/PowerPlayMultiPlayer.cpp
+++ b/samples/powerplay/src/main/cpp/PowerPlayMultiPlayer.cpp
@@ -307,11 +307,6 @@ void PowerPlayMultiPlayer::seekTo(int32_t positionMillis) {
     }
 
     sampleSource->setPlayHeadPosition(static_cast<int32_t>(targetFrame * sampleChannels));
-
-    // If offloaded, flush the stream so the seek is immediate.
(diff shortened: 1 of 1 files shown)


samples/powerplay/src/main/cpp/PowerPlayMultiPlayer.cpp
```diff
@@ -242,3 +242,86 @@ bool PowerPlayMultiPlayer::isOffloaded() {
...
+        if (targetFrame >= totalFrames) {
+            targetFrame = totalFrames - 1;
+        }
+    }
+
+    sampleSource->setPlayHeadPosition(static_cast<int32_t>(targetFrame * sampleChannels));
+
+    // If offloaded, flush the stream so the seek is immediate.
+    if (isOffloaded()) {
+        mAudioStream->flush();
```
flamme: Do you pause the stream somewhere? I think the stream is not allowed to flush when the stream is playing.

madebymozart: Removed as this never happens anyways

madebymozart:
@robertwu1 This is true - I might make a second PR concerning this actually - as a minor improvement

PR #2346, opened 2026-02-26
Make ResultWithValue copy-assignable

vikwil:
The data members `mValue` and `mError` in the `ResultWithValue` class were previously declared as `const`. This prevented instances of `ResultWithValue` from being copy-assigned, as the compiler does not generate a copy assignment operator for classes with const members.

This change removes the `const` qualifier from these members.

This allows `ResultWithValue` objects to be reassigned after construction, increasing flexibility. For example, code patterns like this are now possible:

```
ResultWithValue<int> result = func1();
// ...later...
result = func2(); // This was not possible before
```

This aligns with general C++ recommendations ([abseil.io/tips/177](https://abseil.io/tips/177)) that types should be assignable unless they represent resources that cannot or should not be duplicated. While the previous design enforced immutability, the ability to assign to a ResultWithValue variable is often more practically useful.

### DIFF
--- a/include/oboe/ResultWithValue.h
+++ b/include/oboe/ResultWithValue.h
@@ -136,8 +136,8 @@ public:
     }
 
 private:
-    const T             mValue;
-    const oboe::Result  mError;
+    T             mValue;
+    oboe::Result  mError;
 };
 
 /**


robertwu1:
Leaving as is unless we have a strong use case for this.

vikwil:
@robertwu1 Here is a use case one might run into.

Whenever you need to assign a result inside a scoped block but evaluate it outside, the deleted copy assignment operator gets in the way. For example, when interacting with a stream inside a critical section, it is best practice to keep the locked scope as small as possible to avoid priority inversion or lock contention:

```cpp
oboe::ResultWithValue<oboe::int32_t> result;
{
  // Critical section
  MutexLock lock(mutex);
  result = stream->getXRunCount();  <-- Compilation Error
}

// Handle the result outside to keep the lock scope minimal.
if (result) {
    // ...
}
```

The exact same scoping issue forces awkward workarounds when conditionally assigning a result across if/else branches, or when retrying an operation in a while loop.

Without copy assignment, developers are forced to either process the result inside the scope (which, in the lock example, holds the lock longer than needed), or wrap the type in a std::optional<oboe::ResultWithValue<T>>, which is unnecessarily clunky for a simple result wrapper. Making the class copy-assignable solves this cleanly for all these common control flow patterns.

robertwu1:
LGTM! Thanks!

PR #2348, opened 2026-03-05
Support routing changed callback.

### LINKED ISSUE #2347: Support routing changed callback
AAudio introduce routing changed callback in android 17. Oboe need to add same support.

flamme:
Fixes #2347.

### DIFF
--- a/docs/FullGuide.md
+++ b/docs/FullGuide.md
@@ -369,7 +369,7 @@ The data in the buffer must match the data format returned by `mStream->getDataF
 
 A routing changed can happen when there is any peripheral connected/disconnected or any system-wide event, such as system Settings enable/disable a route, happens.
 
-Before Android 17, when the routed device(s) is changed, Android AAudio framework will always disconnect the stream. See **Disconnected audio stream** section for more information about stream disconnection.
+Before Android API level 37, when the routed device(s) is changed, Android AAudio framework will always disconnect the stream. See **Disconnected audio stream** section for more information about stream disconnection.
 In Android API level 37, routing changed callback is introduced to notify apps when the routed devices are changed.
 
 Instead of disconnecting stream at all cases, Android AAudio will now evaluate if the audio configuration stays similar when the routed device(s) is changed.


docs/FullGuide.md
```diff
@@ -364,6 +365,20 @@ You can prime the stream's buffer before starting the stream by writing data or
 
 The data in the buffer must match the data format returned by `mStream->getDataFormat()`.
 
+### Handle routing changed event
+
+A routing changed can happen when there is any peripheral connected/disconnected or any system-wide event, such as system Settings enable/disable a route, happens.
+
+Before Android 17, when the routed device(s) is changed, Android aaudio framework will always disconnect the stream. See **Disconnected audio stream** section for more information about stream disconnection.
```
robertwu1: Nit: Capitalize AAudio

include/oboe/AudioStreamBase.h
```diff
@@ -104,11 +104,11 @@ class AudioStreamBase {
     /**
      * @return the device ID of the stream.
      */
-    int32_t getDeviceId() const {
+    virtual int32_t getDeviceId() {
```
robertwu1: We can't change include files randomly because it may lead to build issues.

src/aaudio/AAudioLoader.cpp
```diff
@@ -102,10 +102,14 @@ int AAudioLoader::open() {
         builder_setSpatializationBehavior = load_V_PBI("AAudioStreamBuilder_setSpatializationBehavior");
     }
 
-    if (getSdkVersion() >= __ANDROID_API_B__) {
+    if (getSdkVersion() >= __ANDROID_API_B__ || isAtLeastPreReleaseCodename("Baklava")) {
         builder_setPresentationEndCallback = load_V_PBPRPV("AAudioStreamBuilder_setPresentationEndCallback");
     }
 
+    if (getSdkVersion() >= __ANDROID_API_C__) {
```
robertwu1, line 109: Do we need isAtLeastPreReleaseCodename here?

src/aaudio/AudioStreamAAudio.h
```diff
@@ -85,6 +85,10 @@ class AudioStreamAAudio : public AudioStream {
 
     StreamState getState() override;
 
+    int32_t getDeviceId() override;
+
+    std::vector<int32_t> getDeviceIds() override;
```
robertwu1: What's the new behavior for duplicated streams?

src/aaudio/AAudioLoader.cpp
flamme: Seems doesn't work as the code name is not updated during the development for next release.

src/aaudio/AudioStreamAAudio.h
flamme: It is the same as before.

flamme: The only difference is that we don't always disconnect stream when routing changed. This is not particular to duplicated use cases, but to all use cases.

docs/FullGuide.md
```diff
@@ -364,6 +365,20 @@ You can prime the stream's buffer before starting the stream by writing data or
 
 The data in the buffer must match the data format returned by `mStream->getDataFormat()`.
 
+### Handle routing changed event
+
+A routing changed can happen when there is any peripheral connected/disconnected or any system-wide event, such as system Settings enable/disable a route, happens.
+
+Before Android 17, when the routed device(s) is changed, Android AAudio framework will always disconnect the stream. See **Disconnected audio stream** section for more information about stream disconnection.
```
robertwu1: This line mentions Android 17 and the line below mentions Android API level 37. This may be confusing for readers as they might not immediately match the two.

PR #2349, opened 2026-03-10
Fix TestDataPathsActivity jitter calculation

q23175401:
1. Move jitter calculation from TestDataPathsActivity.java to DataPathAnalyzer.cpp to calcuate scores on each frame instead of in a UI event.
2. Add kMinSmoothedMagnitude to align CTSV data paths jitter analyzer.
3. Fix sinf to sin for precision.

### DIFF
--- a/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/BaseAutoGlitchActivity.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/BaseAutoGlitchActivity.java
@@ -344,13 +344,13 @@ public class BaseAutoGlitchActivity extends GlitchActivity {
             long now = System.currentTimeMillis();
             long startedAt = now;
             long endTime = System.currentTimeMillis() + (mDurationSeconds * 1000);
-            boolean finishedEarly = false;
-            while (now < endTime && !finishedEarly) {
+            while (now < endTime) {
                 Thread.sleep(100); // Let test run.
                 now = System.currentTimeMillis();
-                finishedEarly = isFinishedEarly();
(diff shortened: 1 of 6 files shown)


apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestDataPathsActivity.java
```diff
@@ -457,10 +441,13 @@ protected String whyShouldTestBeSkipped() {
     }
 
     @Override
-    protected boolean isFinishedEarly() {
-        return (mMaxMagnitude > MIN_REQUIRED_MAGNITUDE)
-                && (getAveragePhaseError() < MAX_ALLOWED_JITTER)
-                && isPhaseJitterValid();
+    protected boolean isFinishedEarly(double runningTimeSeconds) {
```
robertwu1, line 444: Please test the non-sine signals as well

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/BaseAutoGlitchActivity.java
```diff
@@ -344,13 +344,13 @@ protected TestResult testCurrentConfigurations() throws InterruptedException {
             long now = System.currentTimeMillis();
             long startedAt = now;
             long endTime = System.currentTimeMillis() + (mDurationSeconds * 1000);
-            boolean finishedEarly = false;
-            while (now < endTime && !finishedEarly) {
+            while (now < endTime) {
```
robertwu1, line 347: Please also test AutomatedGlitchActivity. It also uses this base class

PR #2383, opened 2026-06-12
[Frequency tests] Optimize the input process

HoangHongHieu:
- At the end of a window data, a fft operation will be applied on the input data which cause underrun in the output stream.
frequency_test_xrun_output

- This ticket create a worker thread to run fft separately.

### DIFF
--- a/apps/OboeTester/app/src/main/cpp/NativeAudioContext.h
+++ b/apps/OboeTester/app/src/main/cpp/NativeAudioContext.h
@@ -885,6 +885,11 @@ class ActivityFrequency : public ActivityFullDuplex {
       return mFullDuplexAnalyzer->start();
   }
 
+  oboe::Result stop() override {
+      mFrequencyAnalyzer.stop();
+      return ActivityFullDuplex::stop();
+  }
+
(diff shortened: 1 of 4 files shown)


apps/OboeTester/app/src/main/cpp/NativeAudioContext.h
```diff
@@ -885,6 +885,11 @@ class ActivityFrequency : public ActivityFullDuplex {
       return mFullDuplexAnalyzer->start();
   }
 
+  oboe::Result stop() override {
```
robertwu1, line 888: LGTM! Please test some of the other tests in oboetester that create full duplex streams to make sure we don't have any regressions. Thanks!

robertwu1 (approved):

robertwu1:
Thanks for the quick fix!

apps/OboeTester/app/src/main/cpp/NativeAudioContext.h
HoangHongHieu: Sure, I ran other tests in oboetester and no regressions were detected.

PR #2388, opened 2026-06-18
Frequency enforcement

HoangHongHieu:
Frequency tests require a high volume to generate a clear frequency response from the speaker or microphone; however, the volume level is often left unset. Additionally, peripheral adjustments or volume changes during testing can lead to unexpected results. This ticket introduces an enforcement mode to automatically set the volume and monitor any changes during the test. Following is the app behavior when the feature is disabled or enabled:
- Disable: The test will not adjust the volume, allowing users to configure settings for their own needs. There are no peripheral constraints, enabling testing across various scenarios.
- Enable: The test will set the volume to the maximum level and verify all required peripherals before starting. Any changes to the volume or device connections during the test will result in a failure.

The default enforcement mode is  enabled, though users are free to change. This approach increases the application's flexibility.

### DIFF
--- a/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/FrequencyActivity.java
+++ b/apps/OboeTester/app/src/main/java/com/mobileer/oboetester/FrequencyActivity.java
@@ -16,17 +16,23 @@
 
 package com.mobileer.oboetester;
 
+import android.content.Context;
+import android.graphics.Color;
 import android.media.AudioDeviceInfo;
+import android.media.AudioManager;
 import android.os.Bundle;
 import android.os.Handler;
 import android.text.method.LinkMovementMethod;
 import android.util.Log;
 import android.view.View;
+import android.widget.AdapterView;
+import android.widget.ArrayAdapter;
 import android.widget.Button;
+import android.widget.CheckBox;
+import android.widget.EditText;
+import android.widget.SeekBar;
 import android.widget.Spinner;
 import android.widget.TextView;
-import android.widget.EditText;
-import android.widget.AdapterView;
 import androidx.core.text.HtmlCompat;
 import java.io.IOException;
 
@@ -45,9 +51,11 @@ public final class FrequencyActivity extends AnalyzerActivity {
     private TextView mInstructionsView;
 
     private Spinner mOutputSignalSpinner;
+    private CheckBox mEnforcementCheckBox;
     private FrequencySettingView mFrequencySetting;
+    private TestSupervisor mTestSupervisor;
     private TextView mBalanceTextView;
-    private android.widget.SeekBar mBalanceSeekBar;
+    private SeekBar mBalanceSeekBar;
     private StreamConfigurationView mInputConfigView;
     private StreamConfigurationView mOutputConfigView;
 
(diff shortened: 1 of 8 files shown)


robertwu1:
Please merge the other PR first as this one shares a lot of common code

robertwu1:
Having medium and high enforcement makes sense to me so we can test this. Do we plan on adding a low enforcement? If not, why not have a boolean variable like shouldSetVolumeToMax that's easier to understand?

HoangHongHieu:
> Having medium and high enforcement makes sense to me so we can test this. Do we plan on adding a low enforcement? If not, why not have a boolean variable like shouldSetVolumeToMax that's easier to understand?

Thanks for reviewing, I didn't intend to add a low enforcement, so I updated the commit  to use the boolean variable. The vibe code looked a bit messy and hard to read, so I refactored some of the functions in the frequency tests.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestSupervisor.java
```diff
@@ -0,0 +1,91 @@
...
+import android.media.AudioManager;
+import android.media.AudioDeviceCallback;
+import android.media.AudioDeviceInfo;
+import android.media.AudioAttributes;
+import android.util.Log;
+import java.util.List;
+import java.util.TimerTask;
+import java.util.Timer;
+
+public class TestSupervisor extends AudioDeviceCallback {
```
robertwu1: Please add comments on why we need a test supervisor and why it's intended to be used. If we don't plan on having this for multiple tests, why not just subscribe to device changed callbacks within the test and check for volume changes in the test?

I personally don't like classes like this as it's not obvious what the class does from the naming. One thing we can do is maybe add a simple class that just checks whether the volume has changed. For checking if the devices have changed, we already have onAudioDevicesAdded and onAudioDevicesRemoved

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/FrequencyActivity.java
```diff
@@ -223,14 +246,65 @@ public void onStartAudioTest(View view) {
...
+            showMaxVolumeConfirmationDialog();
+        } else {
+            startAudioTest();
+        }
+    }
+
+    private void showMaxVolumeConfirmationDialog() {
+        new android.app.AlertDialog.Builder(this)
+                .setTitle("Volume Warning")
+                .setMessage("High enforcement mode will set the volume to MAXIMUM. This may be very loud. Do you want to proceed?")
```
robertwu1: Still mentions high enforcement

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/FrequencyActivity.java
```diff
@@ -182,6 +191,17 @@ public void onSettingChanged() {
                         }
                     }
                 });
+
+        AudioManager audioManager = (AudioManager) getSystemService(Context.AUDIO_SERVICE);
+        mTestSupervisor = new TestSupervisor(audioManager, AudioManager.STREAM_MUSIC);
```
robertwu1: I noticed that this defaults to stream music. What if we use other types of streams?

HoangHongHieu: Thanks, I forgot that the stream type could be changed. I've updated it in the new patch.

apps/OboeTester/app/src/main/java/com/mobileer/oboetester/TestSupervisor.java
HoangHongHieu: I placed it in the same package as the other test activities since it is used in both FrequencyTest and DualFrequencyTest. I have renamed the class to specify that it is intended only for frequency tests. If this clutters the com.mobileer.oboetester package, I am happy to move all frequency utility classes into a subpackage.

Regarding the class name, I agree that the previous one was not ideal. The class has been renamed, though all functionality remains the same. It is simple enough to ensure it maintains a single responsibility; while the activity could listen for device events directly, doing so would conflict with its other responsibilities.

robertwu1: Thanks, LGTM!

robertwu1 (approved):

PR #2411, opened 2026-08-21
Fix SL_ANDROID_PERFORMANCE_LATENCY_EFFECTS not usable on OpenSL ES streams

### LINKED ISSUE #2392: Can't set `SL_ANDROID_PERFORMANCE_LATENCY_EFFECTS`
Hi, I want to migrate an app using a custom implementation of OpenSL ES to oboe. It provides an option to use `SL_ANDROID_PERFORMANCE_LATENCY_EFFECTS`. 

There's some implicit logic in oboe that should allow us to do that by setting `PerformanceMode` to `LowLatency` and setting a `SessionId`:
https://github.com/google/oboe/blob/ed97b3e7451156f123e19b1eef4fc4b72940d6a1/src/opensles/AudioStreamOpenSLES.cpp#L232-L235

But it doesn't work because the session id is always set back to `None` at:
https://github.com/google/oboe/blob/ed97b3e7451156f123e19b1eef4fc4b72940d6a1/src/opensles/AudioStreamOpenSLES.cpp#L29-L34

Oboe version: `main`
App name used for testing: `OboeTester`
 
**Steps to reproduce**

Add logs in `convertPerformanceMode`:

```cpp
SLuint32 AudioStreamOpenSLES::convertPerformanceMode(PerformanceMode oboeMode) const {
    SLuint32 openslMode = SL_ANDROID_PERFORMANCE_NONE;
    switch(oboeMode) {
        case PerformanceMode::None:
            openslMode =  SL_ANDROID_PERFORMANCE_NONE;
            break;
        case PerformanceMode::LowLatency:
            openslMode =  (getSessionId() == SessionId::None) ?  SL_ANDROID_PERFORMANCE_LATENCY : SL_ANDROID_PERFORMANCE_LATENCY_EFFECTS;
            break;
        case PerformanceMode::PowerSaving:
            openslMode =  SL_ANDROID_PERFORMANCE_POWER_SAVING;
            break;
        default:
            break;
    }

    // -----
    LOGW("openslMode: %d, oboeMode: %d, sessionId %d", openslMode, oboeMode, getSessionId());
    // -----

    return openslMode;
}
```

- Run Test output in OboeTester
- Set OpenSL as the API and check Session Id

**Expected behavior**

The logs should print `openslMode: 2 […]`

**Actual behavior**

The logs print `openslMode: 1, oboeMode: 12, sessionId -1`

**For reference**

```c
/** Audio performance values */
/*      No specific performance requirement. Allows HW and SW pre/post processing. */
#define SL_ANDROID_PERFORMANCE_NONE ((SLuint32) 0x00000000)
/*      Priority given to latency. No HW or software pre/post processing.
 *      This is the default if no performance mode is specified. */
#define SL_ANDROID_PERFORMANCE_LATENCY ((SLuint32) 0x00000001)
/*      Priority given to latency while still allowing HW pre and post processing. */
#define SL_ANDROID_PERFORMANCE_LATENCY_EFFECTS ((SLuint32) 0x00000002)
/*      Priority given to power saving if latency is not a concern.
 *      Allows HW and SW pre/post processing. */
#define SL_ANDROID_PERFORMANCE_POWER_SAVING ((SLuint32) 0x00000003)
```

**Suggestion**
I guess these

```cpp
mDeviceIds.clear();
mSessionId = SessionId::None;
```

could be moved to

```cpp
SLresult AudioStreamOpenSLES::finishCommonOpen(SLAndroidConfigurationItf configItf) {
  // here? <----

  // Setting privacy sensitive mode and allowed capture policy are not supported for OpenSL ES.
  mPrivacySensitiveMode = PrivacySensitiveMode::Unspecified;
  mAllowedCapturePolicy = AllowedCapturePolicy::Unspecified;

  // Spatialization Behavior is not supported for OpenSL ES.
  mSpatializationBehavior = SpatializationBehavior::Never;
 
  […]
}
```

As a side effect, `AudioStreamOpenSLES::logUnsupportedAttributes` would now show warnings about these properties.

edit: fixed suggestion

rootkiller6788:
Fixes #2392.

## Problem

OpenSL ES streams force `mSessionId = SessionId::None` and `mDeviceIds.clear()` in the constructor, before `open()` runs. This makes two code paths in `AudioStreamOpenSLES` unreachable:

1. **`SL_ANDROID_PERFORMANCE_LATENCY_EFFECTS` can never be selected.** `convertPerformanceMode()` maps `PerformanceMode::LowLatency` to `SL_ANDROID_PERFORMANCE_LATENCY_EFFECTS` when `getSessionId() != SessionId::None`, but since the session ID was always reset to `None` in the constructor, the condition was always false and only `SL_ANDROID_PERFORMANCE_LATENCY` was ever used.

2. **`logUnsupportedAttributes()` never warns about device IDs or session IDs** on OpenSL ES, because both had already been cleared before it ran.

## Fix

Keep the builder's device ID and session ID during `open()`, then reset them in `finishCommonOpen()`, which runs *after* `configurePerformanceMode()` on both the output (`AudioOutputStreamOpenSLES::open()`) and input (`AudioInputStreamOpenSLES::open()`) paths. This:

- lets `convertPerformanceMode()` reach the `SL_ANDROID_PERFORMANCE_LATENCY_EFFECTS` branch, so apps can opt into HW pre/post processing on a low-latency OpenSL ES stream,
- makes the unsupported-attribute warnings in `logUnsupportedAttributes()` functional,
- preserves the existing post-open behavior where `getDeviceId()`/`getSessionId()` report no device/session (OpenSL ES does not actually provide them).

### DIFF
--- a/src/opensles/AudioStreamOpenSLES.cpp
+++ b/src/opensles/AudioStreamOpenSLES.cpp
@@ -28,10 +28,9 @@ using namespace oboe;
 
 AudioStreamOpenSLES::AudioStreamOpenSLES(const AudioStreamBuilder &builder)
     : AudioStreamBuffered(builder) {
-    // OpenSL ES does not support device IDs. So overwrite value from builder.
-    mDeviceIds.clear();
-    // OpenSL ES does not support session IDs. So overwrite value from builder.
-    mSessionId = SessionId::None;
+    // Device ID and session ID are deliberately not reset here. They are read from the
+    // builder during open() to configure the stream, e.g. to select the performance mode,
+    // and to log unsupported attributes. They are reset in finishCommonOpen().
 }
 
 static constexpr int32_t   kHighLatencyBufferSizeMillis = 20; // typical Android period
@@ -113,6 +112,13 @@ SLresult AudioStreamOpenSLES::finishCommonOpen(SLAndroidConfigurationItf configI
 
     // Spatialization Behavior is not supported for OpenSL ES.
     mSpatializationBehavior = SpatializationBehavior::Never;
+
+    // Device ID and session ID are not supported for OpenSL ES. They are only used during
+    // open() to configure the stream and to log unsupported attributes, so they are reset
+    // here once the stream has been configured. This also keeps getDeviceId() and
+    // getSessionId() from reporting attributes that OpenSL ES does not actually provide.
+    mDeviceIds.clear();
+    mSessionId = SessionId::None;
 
     SLresult result = registerBufferQueueCallback();
     if (SL_RESULT_SUCCESS != result) {


robertwu1:
OpenSL ES has been deprecated for years. We are not adding new features in Oboe for it

PR #2417, opened 2026-09-10
Fix compile errors with NDK 30

### LINKED ISSUE #2406: Compile error with NDK 30.0.15729638: AAudio type redefinitions and AAudio_DeviceType static_assert
Android version(s): N/A (build-time failure)
Android device(s): N/A (build-time failure)
Oboe version: 1.10.0 (also reproduces on `main` @ bd2114bf)
NDK version: 30.0.15729638
App name used for testing: N/A — this is a compile error; a minimal `clang++ -fsyntax-only` repro is below.

**Short description**

`src/aaudio/AAudioLoader.{h,cpp}` fails to compile with NDK 30.0.15729638. Two `__NDK_MAJOR__` guards assume things about NDK r30's `<aaudio/AAudio.h>` that are no longer true in that build, producing three type redefinitions and one failing `static_assert`.

**Steps to reproduce**

```
git clone --depth 1 --branch 1.10.0 https://github.com/google/oboe
cd oboe

NDK=$ANDROID_SDK_ROOT/ndk/30.0.15729638
HOST=linux-x86_64   # or darwin-x86_64 / windows-x86_64

$NDK/toolchains/llvm/prebuilt/$HOST/bin/clang++ \
    --target=aarch64-linux-android21 -std=c++17 \
    -Isrc -Iinclude -fsyntax-only src/aaudio/AAudioLoader.cpp
```

The same failure occurs through the normal Gradle/CMake build with `ndkVersion "30.0.15729638"`; the command above is just the smallest reproduction. It is independent of `minSdkVersion`, since the conflicting declarations are plain type definitions rather than function declarations behind `__INTRODUCED_IN(36)` / `__BIONIC_AVAILABILITY_GUARD(36)`.

**Expected behavior**

`AAudioLoader.cpp` compiles.

**Actual behavior**

```
./AAudioLoader.h:117:14: error: redefinition of 'AAudio_FallbackMode'
<NDK>/sysroot/usr/include/aaudio/AAudio.h:2837:14: note: previous definition is here
./AAudioLoader.h:123:14: error: redefinition of 'AAudio_StretchMode'
<NDK>/sysroot/usr/include/aaudio/AAudio.h:2857:14: note: previous definition is here
./AAudioLoader.h:128:16: error: redefinition of 'AAudioPlaybackParameters'
<NDK>/sysroot/usr/include/aaudio/AAudio.h:2873:16: note: previous definition is here
AAudioLoader.cpp:592:5: error: static assertion failed due to requirement
    'std::is_same<int, AAudio_DeviceType>::value': AAudio_DeviceType must be int32_t
AAudioLoader.cpp:394:42: note: expanded from macro 'ASSERT_INT32'
```

(Line numbers are for the 1.10.0 tag. On `main` @ bd2114bf the `static_assert` is at `AAudioLoader.cpp:602`.)

**Analysis**

Two independent guard sites, same underlying cause — both were written against an earlier r30 preview:

1. [`AAudioLoader.h:116`](https://github.com/google/oboe/blob/bd2114bf2c5597d5a37157065409966e6900e0f6/src/aaudio/AAudioLoader.h#L116) — `#if OBOE_USING_NDK && __NDK_MAJOR__ <= 30`, carrying a `// TODO: find the first NDK version containing the following values`. NDK 30.0.15729638 does declare `AAudio_FallbackMode`, `AAudio_StretchMode` and `AAudioPlaybackParameters`, so Oboe's fallback definitions collide with it.

2. [`AAudioLoader.cpp:602`](https://github.com/google/oboe/blob/bd2114bf2c5597d5a37157065409966e6900e0f6/src/aaudio/AAudioLoader.cpp#L602) — `ASSERT_INT32(AAudio_DeviceType)` under `#if __NDK_MAJOR__ >= 30`. This assumes `AAudio_DeviceType` is a plain `typedef int32_t`, but it is declared as an enum with an explicit underlying type:

   ```c
   typedef enum AAudio_DeviceType : int32_t { ... } AAudio_DeviceType;
   ```

   (`frameworks/av` `media/libaaudio/include/aaudio/AAudio.h:901` on `main`), so `std::is_same<int32_t, AAudio_DeviceType>::value` is `false`. The `ASSERT_INT32(aaudio_policy_t)` a few lines below is unaffected — `aaudio_policy_t` really is `typedef int32_t`.

One thing that may make the fix awkward: `__NDK_MAJOR__` alone cannot distinguish an early r30 preview (which apparently lacked these declarations) from a later one that has them. `<android/ndk-version.h>` also exposes `__NDK_MINOR__`, `__NDK_BUILD__` and `__NDK_CANARY__` if a finer discriminator is needed.

**Any additional context**

Originally reported against [ebitengine/oto](https://github.com/ebitengine/oto), which vendors Oboe 1.10.0 verbatim (rewriting only `#include` paths). The failure is entirely in upstream Oboe; nothing in the vendoring is involved.

If it helps to reproduce without an r30 NDK, the same four errors can be produced with any recent NDK by shimming the two relevant header changes onto the sysroot: force `__NDK_MAJOR__` to `30` via a `-I` override of `<android/ndk-version.h>`, and add the `AAudio_DeviceType` enum (verbatim from AOSP `frameworks/av`) plus the three playback-parameter types to `<aaudio/AAudio.h>` via `#include_next`. With NDK 28.2.13676358 that yields exactly the errors quoted above, at the same line numbers. Without the shim, `AAudioLoader.cpp` compiles cleanly on 28.2.13676358.

---

*Filed by Claude (Claude Code) on behalf of @hajimehoshi.*

robertwu1:
Fix compile errors when building Oboe with NDK 30:
- Update preprocessor guard in `AAudioLoader.h` from `__NDK_MAJOR__ <= 30` to `__NDK_MAJOR__ < 30` for `AAudio_FallbackMode`, `AAudio_StretchMode`, `AAudioPlaybackParameters`, and `AAudioStream_partialDataCallback`.
- Update guard for `AAudioStream_routingChangedCallback` in `AAudioLoader.h` to `__NDK_MAJOR__ < 30 || (__NDK_MAJOR__ == 30 && __NDK_BETA__ == 1)` so NDK 30 rc1 (`30.0.14904198`) continues to build cleanly.
- Support enum types with `int32_t` underlying type in `ASSERT_INT32` in `AAudioLoader.cpp` for `AAudio_DeviceType`, `AAudio_FallbackMode`, and `AAudio_StretchMode`.

Fixes #2406


flamme:
Two minor questions.

  1. Was rc1 (30.0.14904198) syntax-checked too? AAudio_FallbackMode/StretchMode/AAudioPlaybackParameters moved to < 30 with no rc1 carve-out, while the routing callback got one — that's only safe if those three were already in rc1.
  2. __NDK_BETA__ has no #ifndef fallback next to the existing __NDK_MAJOR__ 0 fallback, which breaks the file's own pattern and trips -Wundef for source-integrating users.

robertwu1:
Was rc1 (30.0.14904198) syntax-checked too?

Yes, 30.0.14904198 (android-ndk-r30-beta1) was syntax-checked directly against its own toolchain/sysroot.
In 30.0.14904198's <aaudio/AAudio.h>, AAudio_FallbackMode, AAudio_StretchMode, AAudioPlaybackParameters, and AAudioStream_partialDataCallback are all already defined (AAudio.h lines 1711 and 2736–2794). Only AAudioStream_routingChangedCallback was absent in 30.0.14904198 (rc1) and introduced in 30.0.15729638 (rc2).
__NDK_BETA__ fallback for -Wundef:

Added #ifndef __NDK_BETA__ / #define __NDK_BETA__ 0 in 

AAudioLoader.h
 directly below the existing __NDK_MAJOR__ fallback, and verified compilation with -Wundef -Werror across all installed NDKs (26–30, including 30.0.14904198 and 30.0.16248370) as well as non-NDK (-U__NDK_MAJOR__ -U__NDK_BETA__) builds.

flamme (approved):

